PR #3, opened 2019-11-15
Feature/115

adamkudrna:


src/lib/components/ui/Alert/Alert.jsx
```diff
@@ -59,8 +54,9 @@ Alert.propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node,
   ]).isRequired,
+  icon: PropTypes.node,
```
bedrich-schindler: There should be `PropTypes.element`. `PropTypes.node` is generic type that can contains everything (like Number, String, Bool, Element, Array)

src/lib/components/ui/Button/Button.jsx
```diff
@@ -69,14 +72,14 @@ const Button = (props) => {
...
     >
-      {(props.icon || props.loading) && (
+      {props.icon && (
         <span
           className={`
             ${styles.icon}
             ${props.loading ? styles.iconLoading : ''}
           `.trim()}
         >
-          <Icon icon={props.loading ? 'sync' : props.icon} size={props.size} />
```
bedrich-schindler: This is going to break all our projects. When `loading` is set to `true`, we need to have button with loading icon in it. After your change, `loading` is not used.

src/lib/components/ui/Button/Button.jsx
```diff
@@ -108,8 +111,8 @@ Button.propTypes = {
   block: PropTypes.bool,
   clickHandler: PropTypes.func,
   disabled: PropTypes.bool,
-  icon: PropTypes.string,
-  iconPosition: PropTypes.oneOf(['before', 'after']),
+  icon: PropTypes.node,
```
bedrich-schindler: There should be PropTypes.element. PropTypes.node is generic type that can contains everything (like Number, String, Bool, Element, Array)

src/lib/components/ui/Alert/Alert.jsx
mbohal: Perhaps we could make it even stricter with `PropTypes.instanceOf(Icon)`? We can always relax this constraint to `PropTypes.element` without introducing breaking change.

src/lib/components/ui/Button/Button.jsx
mbohal: I think we need another `prop` for `loadingIcon`. it is still a breaking change though…

src/lib/components/ui/Button/Button.jsx
mbohal: Perhaps we could make it even stricter with `PropTypes.instanceOf(Icon)`? We can always relax this constraint to `PropTypes.element` without introducing breaking change.

src/lib/components/ui/Alert/Alert.jsx
adamkudrna: I'm going for `PropTypes.element` because `Icon` can be a pure function (and default RUI icons _are_ pure functions), `instanceOf` is applicable only for classes.

mbohal: :+1:

src/lib/components/ui/Button/Button.jsx
```diff
@@ -85,37 +79,56 @@ const Button = (props) => {
...
   block: PropTypes.bool,
   clickHandler: PropTypes.func,
   disabled: PropTypes.bool,
-  icon: PropTypes.string,
-  iconPosition: PropTypes.oneOf(['before', 'after']),
+  endCorner: PropTypes.element,
   id: PropTypes.string,
   label: PropTypes.string.isRequired,
   labelVisibility: PropTypes.oneOf(['all', 'desktop', 'none']),
   loading: PropTypes.bool,
```
mbohal: I think we should indicate loading status by passing in `loadingIcon` element. This way we would be able to completely separate indication of loading from the other icons. I would prefer to avoid the situation where I have to determine what `afterLabel`/`beforeLabel` icon I have to pass in depending on the loading status.

So I think that `loading: PropTypes.bool,` shoul change to `loadingIcon: PropTypes.element,`.

This way we loose the ability to indicate loading status without displaying icon, but I think that is OK.

src/lib/components/ui/Button/Button.jsx
```diff
@@ -85,37 +79,56 @@ const Button = (props) => {
...
   loading: false,
   priority: 'default',
   size: 'medium',
+  startCorner: null,
   type: 'button',
   variant: 'primary',
 };
 
 Button.propTypes = {
+  afterLabel: PropTypes.element,
```
mbohal: Maybe we could rename it to `afterLabelIcon`? This goes for `beforeLabel` as well.

src/lib/components/ui/Button/Button.jsx
```diff
@@ -85,37 +79,56 @@ const Button = (props) => {
...
-  icon: PropTypes.string,
-  iconPosition: PropTypes.oneOf(['before', 'after']),
+  endCorner: PropTypes.element,
   id: PropTypes.string,
   label: PropTypes.string.isRequired,
   labelVisibility: PropTypes.oneOf(['all', 'desktop', 'none']),
   loading: PropTypes.bool,
   priority: PropTypes.oneOf(['default', 'outline', 'flat']),
   size: PropTypes.oneOf(['small', 'medium', 'large']),
+  startCorner: PropTypes.element,
```
mbohal: Maybe we could rename it to `startCornerBadge`? This goes for `endCorner` as well.

src/lib/components/ui/Button/Button.jsx
adamkudrna: This is intentional, it doesn't have to be icon – it can be badge or whatever.

src/lib/components/ui/Button/Button.jsx
adamkudrna: This is intentionally loose, see above.

src/lib/components/ui/Button/Button.jsx
mbohal: What about `afterLabelElement` then?

`afterLabel` just seems very cryptic…

src/lib/components/ui/Button/Button.jsx
adamkudrna: Makes sense, I second your suggestion.

BTW, I'd like to change to loading status a bit: now the loading icon appears when a button is in loading state which makes the button suddenly grow. This is not a smooth behaviour so I'd like the buttons to retain their original width and reveal the (passed in) loading icon _in place_ of the original label. But I haven't found any time for this and I wanted to finally finish this already lengthy task…

src/lib/components/ui/Button/Button.jsx
adamkudrna: >This is going to break all our projects

You're right. I marked the PR with the _BC_ label.

src/lib/components/ui/Button/Button.jsx
adamkudrna: I was also considering that. I found inspiration in MUI which calls it—also unprecisely—`startIcon` and `endIcon`. Since I didn't want to (or maybe rather couldn't) restrict these elements to be just icons, I decided not to specify that at all. I can add the `Element` part to it though if you think it's better.

PR #22, opened 2020-02-07
Create Toolbar layout

adamkudrna:
image


src/lib/components/layout/Toolbar/__tests__/ToolbarSpacer.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import ToolbarSpacer from '../ToolbarSpacer';
+
+describe('rendering', () => {
+  it('renders correctly with a single child', () => {
```
mbohal: I think the description should read: `renders correctly`
There are no children involved.

Issue #24, opened 2020-02-25
Components that wrap native elements should forward refs


bedrich-schindler:
Components that wraps native elements  should forward refs, see https://reactjs.org/docs/forwarding-refs.html.

Following component should use this principe:

- Button
- Checkbox
- MultipleSelectField
- SelectField
- TextArea
- TextField
- Toogle

The only component that cannot use this princip is Radio, because it is rendered as list of radio inputs, so ref forwarding cannot be easily accomplished.

PR #25, opened 2020-02-25
Components that wrap native elements forwards refs (#24)

bedrich-schindler:


src/lib/components/ui/withForwardedRef.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default (WrappedComponent) => {
+  const withForwardedRef = (props, ref) => (
+    <WrappedComponent {...props} forwardedRef={ref} />
+  );
+
+  withForwardedRef.displayName = `withForwardedRef(${WrappedComponent.name || WrappedComponent.name})`;
```
bedrich-schindler: @mbohal Reason is to display readable component name in DevTools and snapshots: https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

src/lib/components/ui/withForwardedRef.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default (WrappedComponent) => {
```
adamkudrna: I'm just wondering where to place this helper (wrapper) because it actually isn't a UI component. Maybe `src/lib/components/` would be fine?

adamkudrna: Never mind, I overlooked it's in the `ui` directory and not in `ui/withForwardedRef/` which disturbed me. It's fine the way it is.

PR #26, opened 2020-02-25
Feature/card layout

adamkudrna:
Card actions now can be aligned in grid layout even when Cards' height differs:

image

⚠️ This is a breaking change, using `CardBody` is now mandatory for all Cards, even when they don't use `CardFooter`.


bedrich-schindler:
> If I understand it correctly, then `<Card>` can only receive as its children `<CardBody>` and/or `<CardFooter>`. That is what makes this is a breaking change.
> 
> The `propTypes` validation of `<Card>` however accepts any node as its child. I think the check should be stricter.

So `children` of `Card` should use `instanceOf(CardBody)` and `instanceOf(CardFooter)` prop types checks? Am I correct?

mbohal:
> So children of Card should use instanceOf(CardBody) and instanceOf(CardFooter) prop types checks? Am I correct?

I think so.

adamkudrna:
@mbohal @bedrich-schindler We already discussed the same thing earlier, cannot recall where – `instanceOf` is only possible for React components, not for pure functions…

The change is breaking mostly visually – if you upgrade and leave out the new `CardBody` element, there will be no spacing around the `Card` content. Which I can imagine may be useful in some cases, but not for those you already have, therefore it's a BC.

bedrich-schindler:
> @mbohal @bedrich-schindler We already discussed the same thing earlier, cannot recall where – `instanceOf` is only possible for React components, not for pure functions…
> 
> The change is breaking mostly visually – if you upgrade and leave out the new `CardBody` element, there will be no spacing around the `Card` content. Which I can imagine may be useful in some cases, but not for those you already have, therefore it's a BC.

Then is it OK. We have forgotten that `instanceOf` applies to classes only.

Issue #29, opened 2020-02-26
Deal with Webpack warnings in console


adamkudrna:
Webpack and npm report warnings:

- production/development mode has not been set
- recommended size limit 244 kB has been exceeded (1.12 MB)
- refactor deprecated npm's `prepublish` task

![image](https://user-images.githubusercontent.com/5614085/75377270-bb2e3c00-58d1-11ea-95d5-4889e5b60fba.png)

![image](https://user-images.githubusercontent.com/5614085/75377612-7c4cb600-58d2-11ea-840d-77d324633cd3.png)

![image](https://user-images.githubusercontent.com/5614085/75392257-ec1b6a80-58eb-11ea-8c64-8892eda2addb.png)

adamkudrna:
@mbohal @bedrich-schindler What shall we do with the size limit? After I have set the mode to production the size dropped to 745 kB which is still 0.5 MB more than recommended. Is it possible to configure the size threshold?

mbohal:
Why is it so big? Icons? I would prefer to try to make it smaller over increasing the threshold?

adamkudrna:
#metoo but don't know how. Does @bedrich-schindler know how to investigate it?

bedrich-schindler:
There are several ways to optimize bundle size. Ex. optimize imports for three shaking, use ESM, use @babel/runtime and so on. 

As I optimized another application (RipEx) that use this library, I told @mbohal that next step is to optimize this library. 

I think that I am able to optimize the size a little bit. But the biggest part of whole bundle are icons. I am in to removing icons from this library because they are not used directly anymore. So we can delete them at all or create react-ui-icons that can be used only if needed.

mbohal:
I think we should remove the icons from the library itself. We will still need some icons in the repo for demo purposes in the documentation. These however should not be part of the package that is installed by the consumers of react-ui library.

In other words, providing the icons should be the responsibility of the consumer of react-ui and the demo application in this repo is just another consumer.

PR #30, opened 2020-02-26
Set production mode in Webpack config to prevent warnings in console

adamkudrna:
As recommended by Webpack warning in console, see #29.


mbohal:
I think we should revert these changes and set production mode in `package.json` via the `-p` flag in the `scripts` section.

That is IMHO more flexible approach.

adamkudrna:
The problem is better addressed in #31, closing the PR.

Issue #16, opened 2019-12-09
Remove unnecessary builds from repository


bedrich-schindler:
As we publish library on NPM registry, we do not need to have builds of library and demo in repository. We can gitignore them and publish only lib.js on NPM registry.

PR #31, opened 2020-02-29
Remove Icon component from library, update dependencies and improve b…

bedrich-schindler:
…uild to decrease bundle sizes (#29, #16)


src/demo/components/Icon/Icon.scss
```diff
@@ -0,0 +1,40 @@
...
+
+@mixin icon($icon) {
+  width: $svg-size;
+  height: $svg-size;
+  background-image: url('#{$svg-data},#{$icon}');
+  background-size: $svg-size $svg-size;
+}
+
+.iconDown {
+  $icon: 'PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAyNCAyNCIgd2lkdGg9IjI0Ij48cGF0aCBkPSJNMCAwaDI0djI0SDBWMHoiIGZpbGw9Im5vbmUiLz48cGF0aCBkPSJNMjAgMTJsLTEuNDEtMS40MUwxMyAxNi4xN1Y0aC0ydjEyLjE3bC01LjU4LTUuNTlMNCAxMmw4IDggOC04eiIvPjwvc3ZnPg==';
```
adamkudrna: For maintenance reasons, I'd prefer to keep here also legible source of the SVG. Now I'd like to make the icons inherit their color from text (leaving the `fill` unset or set to `currentColor`) which is hard without knowing the source image.

My suggestion is to use something like [this](https://gist.github.com/stowball/ff5803cb6cc466ac77ccc21c1ed19842) or [this](https://codepen.io/kevinweber/pen/dXWoRw).

src/lib/index.js
```diff
@@ -36,7 +36,6 @@ export { Card } from './components/ui/Card';
 export { CardBody } from './components/ui/Card';
 export { CardFooter } from './components/ui/Card';
 export { default as CheckboxField } from './components/ui/CheckboxField';
-export { default as Icon } from './components/ui/Icon';
```
adamkudrna: Please also remove Icon configuration from `src/lib/theme.scss`, lines 141–149:

image

src/demo/components/Icon/Icon.scss
```diff
@@ -0,0 +1,40 @@
+// stylelint-disable max-line-length
+$svg-data: 'data:image/svg+xml;base64';
+$svg-size: 16px;
```
adamkudrna:
```suggestion
$svg-size: 1rem;
```

bedrich-schindler:
> Nice cleanup! The new library size (192 kB) is really cool.
> 
> I'd only really wish to have the work committed in a more structured way:
> 
> * Remove icons from library (#29)
> * Update dependencies
> * Decrease bundle build size (#29)
> * Remove unnecessary builds from repository (#16)
> 
> It will be ~three~ four items in the change log anyway. Do you think you could do it even now…?

Done.

src/lib/index.js
bedrich-schindler: Done

src/demo/components/Icon/Icon.scss
bedrich-schindler: Done, thanks

src/demo/components/Icon/Icon.scss
bedrich-schindler: I tried it but without success. Can you try to do it instead of me? Icons are taken from Material Icons:

* sync => loading
* arrow_downward => down
* arrow_upward => up
* stars => star
* check_circle => success

mbohal:
The `README.md` needs to be opdated. We need to:
* Remove obsolete information.
* Inform users that icons are not part of the UI library. It might be helpful to provide a brief explanation why it is so.

bedrich-schindler:
> The `README.md` needs to be opdated. We need to:
> 
> * Remove obsolete information.
> * Inform users that icons are not part of the UI library. It might be helpful to provide a brief explanation why it is so.

Done

src/demo/components/Icon/Icon.scss
adamkudrna: I succeeded but anyway — you can only have SVG inlined in order to inherit current text color (`fill: currentColor`). I added it as a fixup to your commit.

README.md
```diff
@@ -79,6 +82,9 @@ With output size in mind, we recommend to choose from these theming options:
 To keep your app UI consistent, use the theme properties from `src/lib/theme.scss` also in your
 custom styles.
 
+Icons have been removed from library to reduce its size. Components that require icons allow to pass icon
```
adamkudrna: I would simply mention that no icons are included in the library and some components accept them via props. `README` is primarily for new-comers and it's fine when the message about removing icons (and why) remains only in change log (release notes).

bedrich-schindler: @mbohal wrote that he want to have this information in readme, you two need to decide whether to have it here or not.

adamkudrna: @mbohal? I'd just say in the README there are no icons included and describe the reason of their removal in change log. What do you think?

mbohal: After phone call with @adamkudrna we clarified that the problem is not the information itself, but the past tense. Users need to know why the icons are not there, they do not need to know why it is so.

I would rephrase it to something like:
` Icons are not included in the library to allow more flexibility and to reduce its size. Components that require icons allow to them to be passed in via props.`

bedrich-schindler: Ok, got it.

README.md
```diff
@@ -79,6 +82,9 @@ With output size in mind, we recommend to choose from these theming options:
 To keep your app UI consistent, use the theme properties from `src/lib/theme.scss` also in your
 custom styles.
 
+Icons are not included in the library to allow more flexibility and to reduce its size. Components
+that require icons allow to them to be passed in via props.
```
adamkudrna: ❤️ !

PR #33, opened 2020-03-06
Exclude documentation components from library

adamkudrna:
10 kB smaller!


bedrich-schindler:
I changed css-loader settings and decreased bundle size by 32 kB, it means 42 kB in total with @adamkudrna`s changes

webpack.config.demo.js
```diff
@@ -31,7 +31,9 @@ module.exports = [{
             loader: 'css-loader',
             options: {
               modules: {
-                localIdentName: '[name]__[local]__[hash:base64:5]',
+                localIdentName: process.argv.includes('production')
```
mbohal: This seems a bit brittle.
I would prefer to detect the mode explicitly. Something like this:
```
...
module.exports = (env, options) => [{`
...
localIdentName: options.mode === 'production'
                  ? '[hash:base64:8]'
                  : '[name]__[local]__[hash:base64:8]',
...
```

Same in `webpack.config.lib.js`

adamkudrna:
> demo.js and lib.js shoud not be pushed to git, there are ignored in .gitgnore:
> 
> ```
> /demo/generated
> /dist
> ```
> 
> EDIT: I fixed it

I'm just wondering how they got there, maybe with some rebased branch…? Anyway, thanks for the fix.

PR #36, opened 2020-04-09
Introduce Toolbar Groups and alignment option

adamkudrna:
image
image


bedrich-schindler:
It looks almost great except this:

78865653-46a4fc00-7a3e-11ea-91d7-c00e79216670

I think that there shoudn't be a space.

adamkudrna:
I'm able to get rid of the undesired space when I use `justify-content: space-between`. With this approach, `ToolbarSpacer` is no longer needed. But this would make this kind of justification mandatory, not optional like with the `ToolbarSpacer`. Is that OK? Or do we want to support different types of justification like `start`, `center`, `end` and `space-between`?

toolbar-spacing

adamkudrna:
toolbar-justification

PR #38, opened 2020-04-10
Create CTA layout component

### LINKED ISSUE #37: Layout for centered CTA and secondary actions
We need a very specific layout for a centered CTA button and optional secondary actions on sides. This is required specifically for _Load more_ button but it may be useful in other situations.

adamkudrna:
**Breaking changes:**
- renamed `RowLeft` to `RowStart` and `RowRight` to `RowEnd`
- changed default `Toolbar` alignment to `top`

Closes #37.

image

On small and medium screens, the layout folds automatically in a similar way to `Toolbar` (start cell is always full-width here):

image

image


adamkudrna:
Plus added `nowrap` option for `Toolbar`:

image

Issue #53, opened 2020-04-17
Create ScrollView component with automatic scrolling shadows


adamkudrna:
The component should:

- support horizontal or vertical scroll
- optionally hide scrollbar
- optionally show scrolling shadows where necessary

![image](https://user-images.githubusercontent.com/5614085/79574864-1b1ab500-80c1-11ea-8e92-f4bcfbb8b56a.png)
Image source: https://stackoverflow.com/questions/50654217/adding-horizontal-scrolling-shadows-effect/50655240

PR #58, opened 2020-04-21
ScrollView

### LINKED ISSUE #50: Apply ScrollView in Modal and responsive Table
Blocked by #53.

~~See https://codepen.io/evank/pen/wWbRNO. Blocked by #49.~~

adamkudrna:
image

image

image

Closes #50 and #53.

### DIFF
--- a/src/lib/components/ui/Modal/Modal.jsx
+++ b/src/lib/components/ui/Modal/Modal.jsx
@@ -1,27 +1,23 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { createPortal } from 'react-dom';
-import Button from '../Button';
+import transferProps from '../../../utils/transferProps';
+import { withTranslationContext } from '../../../translation';
 import {
   Toolbar,
   ToolbarItem,
 } from '../../layout/Toolbar';
-import transferProps from '../../../utils/transferProps';
-import { withTranslationContext } from '../../../translation';
+import Button from '../Button';
+import ScrollView from '../ScrollView';
 import styles from './Modal.scss';
 
 class Modal extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = {
-      isContentOverflowing: false,
-    };
-
     this.childrenWrapperRef = React.createRef();
     this.submitButtonRef = React.createRef();
 
-    this.setGradient = this.setGradient.bind(this);
     this.keyPressHandler = this.keyPressHandler.bind(this);
   }
 
@@ -55,17 +51,14 @@ class Modal extends React.Component {
     window.document.removeEventListener('keydown', this.keyPressHandler, false);
   }
 
-  setGradient() {
-    if (!this.state.isContentOverflowing) {
-      this.setState({ isContentOverflowing: true });
-    }
-  }
-
   keyPressHandler(e) {
-    const { actions } = this.props;
+    const {
+      actions,
+      closeHandler,
+    } = this.props;
 
-    if (e.keyCode === 27 && this.props.closeHandler) {
-      this.props.closeHandler();
+    if (e.keyCode === 27 && closeHandler) {
+      closeHandler();
     }
 
     if (e.keyCode === 13 && e.target.nodeName !== 'BUTTON') {
@@ -78,7 +71,21 @@ class Modal extends React.Component {
   }
 
   preRender() {
-    const sizeClass = (size) => {
+    const {
+      actions,
+      children,
+      closeHandler,
+      id,
+      scrollMode,
+      scrollViewEndShadowStyle,
+      scrollViewShadowSize,
+      scrollViewStartShadowStyle,
+      size,
+      title,
+      translations,
+    } = this.props;
+
+    const sizeClass = () => {
       if (size === 'small') {
         return styles.isRootSmall;
       }
@@ -94,23 +101,58 @@ class Modal extends React.Component {
       return styles.isRootAuto;
     };
 
+    const modalBody = () => {
+      const content = (
+        <div
+          ref={this.childrenWrapperRef}
+          className={styles.content}
+          {...(id && { id: `${id}__content` })}
+        >
+          {children}
+        </div>
+      );
+
+      if (scrollMode === 'body') {
+        return (
+          <div
+            className={[
+              styles.body,
+              styles.isBodyScrollable,
+            ].join(' ')}
+          >
+            <ScrollView
+              customStartShadowStyle={scrollViewStartShadowStyle}
+              customEndShadowStyle={scrollViewEndShadowStyle}
+              shadowSize={scrollViewShadowSize}
+            >
+              {content}
+            </ScrollView>
+          </div>
+        );
+      }
+
+      return (
+        <div className={styles.body}>
+          {content}
+        </div>
+      );
+    };
+
     return (
       <div
         className={styles.overlay}
-        id={this.props.id}
+        id={id}
         onClick={(e) => {
-          if (this.props.closeHandler) {
-            this.props.closeHandler(e);
+          if (closeHandler) {
+            closeHandler(e);
           }
         }}
-        onScroll={this.setGradient}
         role="presentation"
       >
         <div
           className={`
             ${styles.root}
-            ${sizeClass(this.props.size)}
-            ${this.state.isContentOverflowing ? styles.isContentOverflowing : ''}
+            ${sizeClass()}
           `.trim()}
           onClick={(e) => {
             e.stopPropagation();
@@ -120,32 +162,26 @@ class Modal extends React.Component {
           <div className={styles.head}>
             <h3
               className={styles.headTitle}
-              {...(this.props.id && { id: `${this.props.id}__title` })}
+              {...(id && { id: `${id}__title` })}
             >
-              {this.props.title}
+              {title}
             </h3>
-            {this.props.closeHandler && (
+            {closeHandler && (
               <button
                 type="button"
                 className={styles.close}
-                onClick={this.props.closeHandler}
-                title={this.props.translations.close}
-                {...(this.props.id && { id: `${this.props.id}__closeModalHeaderButton` })}
+                onClick={closeHandler}
+                title={translations.close}
+                {...(id && { id: `${id}__closeModalHeaderButton` })}
               >
                 ×
               </button>
             )}
           </div>
-          <div
-            className={styles.body}
-            ref={this.childrenWrapperRef}
-            {...(this.props.id && { id: `${this.props.id}__content` })}
-          >
-            {this.props.children}
-          </div>
+          {modalBody()}
           <div className={styles.footer}>
             <Toolbar justify="center" dense>
-              {this.props.actions.map((action) => (
+              {actions.map((action) => (
                 <ToolbarItem key={action.label}>
                   <Button
                     {...transferProps(action)}
@@ -160,13 +196,13 @@ class Modal extends React.Component {
                   />
                 </ToolbarItem>
               ))}
-              {this.props.closeHandler && (
+              {closeHandler && (
                 <ToolbarItem>
                   <Button
-                    clickHandler={this.props.closeHandler}
-                    label={this.props.translations.close}
+                    clickHandler={closeHandler}
+                    label={translations.close}
                     priority="flat"
-                    {...(this.props.id && { id: `${this.props.id}__closeModalFooterButton` })}
+                    {...(id && { id: `${id}__closeModalFooterButton` })}
                   />
                 </ToolbarItem>
               )}
@@ -194,6 +230,14 @@ Modal.defaultProps = {
   closeHandler: null,
   id: undefined,
   portalId: null,
+  scrollMode: 'body',
+  scrollViewEndShadowStyle: {
+    background: 'radial-gradient(farthest-side at center bottom, rgba(0, 0, 0, 0.16) 0%, rgba(0, 0, 0, 0.06) 40%, rgba(0, 0, 0, 0.02) 85%, rgba(0, 0, 0, 0) 100%)',
+  },
+  scrollViewShadowSize: '16px',
+  scrollViewStartShadowStyle: {
+    background: 'radial-gradient(farthest-side at center top, rgba(0, 0, 0, 0.15) 0%, rgba(0, 0, 0, 0.05) 60%, rgba(0, 0, 0, 0.02) 85%, rgba(0, 0, 0, 0) 100%)',
+  },
   size: 'medium',
 };
 
@@ -211,6 +255,16 @@ Modal.propTypes = {
   closeHandler: PropTypes.func,
   id: PropTypes.string,
   portalId: PropTypes.string,
+  scrollMode: PropTypes.oneOf(['body', 'modal']),
+  scrollViewEndShadowStyle: PropTypes.shape({
+    background: PropTypes.string,
+    boxShadow: PropTypes.string,
+  }),
+  scrollViewShadowSize: PropTypes.string,
+  scrollViewStartShadowStyle: PropTypes.shape({
+    background: PropTypes.string,
+    boxShadow: PropTypes.string,
+  }),
   size: PropTypes.oneOf(['small', 'medium', 'large', 'auto']),
   title: PropTypes.string.isRequired,
   translations: PropTypes.shape({
--- a/src/lib/components/ui/ScrollView/ScrollView.jsx
+++ b/src/lib/components/ui/ScrollView/ScrollView.jsx
@@ -1,0 +1,247 @@
+import PropTypes from 'prop-types';
+import React, {
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { useLoadResize } from '../../../hooks/useLoadResizeHook';
+import { useScrollPosition } from '../../../hooks/useScrollPositionHook';
+import { getElementsPositionDifference } from '../../../services/elementPositionService';
+import { withTranslationContext } from '../../../translation';
+import styles from './ScrollView.scss';
+
+export const ScrollView = (props) => {
+  const {
+    arrows,
+    arrowsColor,
+    arrowsScrollStep,
+    children,
+    customEndShadowStyle,
+    customNextArrow,
+    customPrevArrow,
+    customStartShadowStyle,
+    id,
+    debounce,
+    direction,
+    scrollbar,
+    shadowColor,
+    shadowSize,
+    translations,
+  } = props;
+
+  const [isScrolledAtStart, setIsScrolledAtStart] = useState(false);
+  const [isScrolledAtEnd, setIsScrolledAtEnd] = useState(false);
+
+  const scrollPositionStart = direction === 'horizontal' ? 'left' : 'top';
+  const scrollPositionEnd = direction === 'horizontal' ? 'right' : 'bottom';
+  const scrollViewContentEl = useRef(null);
+  const scrollViewViewportEl = useRef(null);
+
+  const handleScrollViewState = (currentPosition) => {
+    const isScrolledAtStartActive = currentPosition[scrollPositionStart] < 0;
+    const isScrolledAtEndActive = currentPosition[scrollPositionEnd] > 0;
+
+    if (isScrolledAtStartActive !== isScrolledAtStart) {
+      setIsScrolledAtStart(isScrolledAtStartActive);
+    }
+
+    if (isScrolledAtEndActive !== isScrolledAtEnd) {
+      setIsScrolledAtEnd(isScrolledAtEndActive);
+    }
+  };
+
+  useEffect(
+    () => {
+      handleScrollViewState(
+        getElementsPositionDifference(scrollViewContentEl, scrollViewViewportEl),
+      );
+    },
+    [], // eslint-disable-line react-hooks/exhaustive-deps
+  );
+
+  useLoadResize(
+    (currentPosition) => (handleScrollViewState(currentPosition)),
+    [isScrolledAtStart, isScrolledAtEnd],
+    scrollViewContentEl,
+    scrollViewViewportEl,
+    debounce,
+  );
+
+  useScrollPosition(
+    (currentPosition) => (handleScrollViewState(currentPosition)),
+    [isScrolledAtStart, isScrolledAtEnd],
+    scrollViewContentEl,
+    scrollViewViewportEl,
+    debounce,
+  );
+
+  const setAlpha = (rgba, alpha) => ({
+    ...rgba,
+    alpha,
+  });
+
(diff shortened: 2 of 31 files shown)


src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -0,0 +1,130 @@
...
+  },
+  shadowSize: '2em',
+};
+
+ScrollView.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+  horizontal: PropTypes.bool,
```
mbohal, line 120: I think, I would prefer to have here it expressed as prop `direction` that would take values `horizontal` and `vertical`. 

This makes it seem like horizontal is special case of vertical.

davidurban:
Any chance of including radial shadows? Especially for horizontal and modal windows uses it would be way more elegant. 
I would also recommend reducing spread of the shadows (especially in horizontal case) to make it more elegant and increase visible area.

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -0,0 +1,130 @@
...
+  shadowColor: {
+    alpha: 1,
+    blue: 255,
+    green: 255,
+    red: 255,
+  },
+  shadowSize: '2em',
+};
+
+ScrollView.propTypes = {
```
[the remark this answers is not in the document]
adamkudrna: Added `debounce` prop with default 50 ms.

src/lib/components/ui/ScrollView/ScrollView.jsx
adamkudrna: OK, changed to `direction` prop.

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -0,0 +1,156 @@
...
+    </div>
+  );
+};
+
+ScrollView.defaultProps = {
+  customEndShadow: undefined,
+  customStartShadow: undefined,
+  debounce: 50,
+  direction: 'vertical',
+  shadowColor: {
```
mbohal, line 123: the default shadow is not visible enough

adamkudrna:
One more thing: I didn't figure out how to initialize `ScrollView` in a `Modal` when it's mounted. The problem is that `ScrollView` gets initialized on window `load` and `resize` events but the `ScrollView` instances in `Modal`s don't exist until it's open. So when you open a `Modal`, the `ScrollView` is not yet initialized and you have to start scrolling in order for shadows to appear.

Does anybody know how we should deal with it? @bedrich-schindler?

bedrich-schindler:
> One more thing: I didn't figure out how to initialize `ScrollView` in a `Modal` when it's mounted. The problem is that `ScrollView` gets initialized on window `load` and `resize` events but the `ScrollView` instances in `Modal`s don't exist until it's open. So when you open a `Modal`, the `ScrollView` is not yet initialized and you have to start scrolling in order for shadows to appear.
> 
> Does anybody know how we should deal with it? @bedrich-schindler?

Sorry, I`ve missed notification about your comment. 

If  I am right, neither `useScrollPositionHook` nor `ScrollView`, use `window` object. Both use reference to root `div` of `ScrollView`. Problem with `useLoadResizeHook` is that there are no events that are triggered when specific element is loaded/ready therefore we cannot map it to the element instead of `window` object. But ...

Component is mounted or (ready) when it is rendered.  It means that we can leave `useLoadResize` as it is and add call it directly in useEffect on mount like this:

```
// ScrollView.jsx

useEffect(() => {
    handleScrollViewState(...);
}, [])
```

Important is to have `[]` to trigger effect only once. Parameters of `> One more thing: I didn't figure out how to initialize `ScrollView` in a `Modal` when it's mounted. The problem is that `ScrollView` gets initialized on window `load` and `resize` events but the `ScrollView` instances in `Modal`s don't exist until it's open. So when you open a `Modal`, the `ScrollView` is not yet initialized and you have to start scrolling in order for shadows to appear.
> 
> Does anybody know how we should deal with it? @bedrich-schindler?

Sorry, I`ve missed notification about your comment. 

If  I am right, neither `useScrollPositionHook` nor `ScrollView`, use `window` object. Both use reference to root `div` of `ScrollView`. Problem with `useLoadResizeHook` is that there are no events that are triggered when specific element is loaded/ready. But ...

Component is mounted or (ready) when it is rendered.  It means that we can leave `useLoadResize` as it is and add call `handleScrollViewState` directly in `useEffect` on mount like this:

```
// ScrollView.jsx

useEffect(() => {
    handleScrollViewState(...);
}, [])
```

Important is to pass `[]` to the effect that will tell React to trigger effect on first render only. So ignore eslint that will probably want to pass another deps to this array.

Hope it will work

adamkudrna:
⚠️ **ATTENTION:** BRANCH REBASED onto `master` ⚠️

adamkudrna:
⚠️ Rebased again.

src/lib/components/ui/ScrollView/ScrollView.jsx
adamkudrna: Are we talking about this?

image

I think it's a good-enough default… We cannot anticipate the nature of content nor the size of its viewport therefore I don't want the scrolling shadows to be too large by default.

At the end of the day, it's always a design decision of particular project how to design the shadows. Compare eg. Apple Music (above, very thin shadows) to TIDAL (below, massive shadows):

image

image

adamkudrna:
@bedrich-schindler Magician! As of `ScrollView` in `Modal`:

```js
  useEffect(
    () => {
      handleScrollViewState(
        getScrollPositionRelativeToViewport(scrollViewContentEl, scrollViewViewportEl),
      );
    },
    [], // eslint-disable-line react-hooks/exhaustive-deps
  );
```

👍

adamkudrna:
Arrows are there! Default design probably needs some adjustments.

image

What's left to do:

1. Enable injecting custom start/end arrow element. Providing all necessary design options through API is just way too complicated and yet not 100% flexible.
2. Make the arrows work 👍.

adamkudrna:
Arrows are there and working.

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -0,0 +1,256 @@
...
+
+  useScrollPosition(
+    (currentPosition) => (handleScrollViewState(currentPosition)),
+    [isScrolledAtStart, isScrolledAtEnd],
+    scrollViewContentEl,
+    scrollViewViewportEl,
+    debounce,
+  );
+
+  const setAlpha = (rgba, alpha) => {
```
bedrich-schindler, line 77: What about this?

```
const setAlpha = (rgba, alpha) => ({ ...rgba, alpha });
```

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -0,0 +1,256 @@
...
+      '--rui-local-end-shadow-background':
+        endShadowStyle.background
+        || `linear-gradient(${direction === 'horizontal' ? 'to left' : 'to top'}, ${shadowStartColor}, ${shadowEndColor})`,
+      '--rui-local-end-shadow-box-shadow': endShadowStyle.boxShadow || 'none',
+    };
+    /* eslint-enable sort-keys */
+  };
+
+  const arrowHandler = (contentEl, viewportEl, scrollViewDirection, shiftDirection, step) => {
+    const differenceX = () => {
```
bedrich-schindler, line 116: It is more efficient not to wrap calculations like this into functions. `differenceX` and `differenceY` are used only here, so it is better to write somethink like:

```
let differenceX = 0;
let differenceY = 0;

if (scrollViewDirection === 'horizontal') {
         differenceX = shiftDirection === 'next'
           ? step
           : -1 * step;
}

// ...

```

src/lib/components/ui/ScrollView/__tests__/ScrollView.test.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
```
bedrich-schindler, line 1: I would like to see functionality test that will check if `scrollBy` is called.

src/lib/hooks/useLoadResizeHook.js
```diff
@@ -0,0 +1,34 @@
...
+  const callBack = () => {
+    effect(getElementsPositionDifference(contentEl, viewportEl));
+    throttleTimeout = null;
+  };
+
+  useLayoutEffect(() => {
+    const handleLoadResize = () => {
+      if (wait) {
+        if (throttleTimeout === null) {
+          // eslint-disable-next-line react-hooks/exhaustive-deps
```
bedrich-schindler, line 16: I don't think that `// eslint-disable-next-line react-hooks/exhaustive-deps` goes here

src/lib/hooks/useLoadResizeHook.js
```diff
@@ -0,0 +1,34 @@
...
+      } else {
+        callBack();
+      }
+    };
+
+    window.addEventListener('load', handleLoadResize);
+    window.addEventListener('resize', handleLoadResize);
+
+    return () => {
+      window.removeEventListener('load', handleLoadResize);
```
bedrich-schindler, line 28: `clearTimeout` need to be added here in case we unmount component earlier than setTimeout callback is called

src/lib/hooks/useScrollPositionHook.js
```diff
@@ -0,0 +1,31 @@
...
+  const callBack = () => {
+    effect(getElementsPositionDifference(contentEl, viewportEl));
+    throttleTimeout = null;
+  };
+
+  useLayoutEffect(() => {
+    const handleScroll = () => {
+      if (wait) {
+        if (throttleTimeout === null) {
+          // eslint-disable-next-line react-hooks/exhaustive-deps
```
bedrich-schindler, line 16: I don't think that //`eslint-disable-next-line react-hooks/exhaustive-deps `goes here

src/lib/hooks/useScrollPositionHook.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+      } else {
+        callBack();
+      }
+    };
+
+    viewportEl.current.addEventListener('scroll', handleScroll);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    return () => viewportEl.current.removeEventListener('scroll', handleScroll);
```
bedrich-schindler, line 27: `clearTimeout ` need to be added here in case we unmount component earlier than setTimeout callback is called

src/lib/hooks/useLoadResizeHook.js
```diff
@@ -0,0 +1,34 @@
+import { useLayoutEffect } from 'react';
+import { getElementsPositionDifference } from '../services/elementPositionService';
+
+export const useLoadResize = (effect, dependencies, contentEl, viewportEl, wait) => {
+  let throttleTimeout = null;
+
+  const callBack = () => {
+    effect(getElementsPositionDifference(contentEl, viewportEl));
+    throttleTimeout = null;
```
bedrich-schindler, line 9: I am wondering, what's the reason to have `throttleTimeout` here? When we do not use `wait` than `throttleTimeout = null` is not necessary here, isn't it?

src/lib/hooks/useScrollPositionHook.js
```diff
@@ -0,0 +1,31 @@
+import { useLayoutEffect } from 'react';
+import { getElementsPositionDifference } from '../services/elementPositionService';
+
+export const useScrollPosition = (effect, dependencies, contentEl, viewportEl, wait) => {
+  let throttleTimeout = null;
+
+  const callBack = () => {
+    effect(getElementsPositionDifference(contentEl, viewportEl));
+    throttleTimeout = null;
```
bedrich-schindler, line 9: I am wondering, what's the reason to have throttleTimeout here? When we do not use wait than throttleTimeout = null is not necessary here, isn't it?

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -59,6 +49,42 @@ class Modal extends React.Component {
...
+      if (!disableBodyScrolling) {
+        return (
+          <div
+            className={[
+              styles.body,
+              styles.isBodyScrollable,
+            ].join(' ')}
+          >
+            <ScrollView
+              customStartShadowStyle={this.props.scrollViewStartShadowStyle}
```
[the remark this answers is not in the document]
mbohal: Actually the rule is explicitly ignored: https://github.com/react-ui-org/react-ui/blob/02352043365c571c748f2e8f3f7d2a24caf4282d/.eslintrc#L16

I created #113

src/lib/services/elementPositionService.js
```diff
@@ -0,0 +1,24 @@
+const isBrowser = typeof window !== 'undefined';
+
+export const getElementsPositionDifference = (contentEl, viewportEl) => {
+  if (!isBrowser) {
```
mbohal, line 4: I assume this is here for tests. I think a cleaner solution would be to just mock this service.

src/lib/components/ui/ScrollView/ScrollView.jsx
mbohal: Also the code is the same for `differenceX` and `differenceY`. i think it could be abstracted to a function.

mbohal:
I think documentation should add warning that horizontal scrollView should always use arrows and/or scroll bar.

It might not be obvious to Mac/touchpad users.

adamkudrna:
> I would like to add vertical scroll view with arrows into demo. I tried to do it manually but no arrows appeared. So please add it to the demo to see it working

image

src/lib/components/ui/Modal/Modal.jsx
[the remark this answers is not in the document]
adamkudrna: @bedrich-schindler So would I but such refactoring is outside this issue's scope (which is already huge). I use this approach in all new components.

src/lib/components/ui/ScrollView/ScrollView.jsx
adamkudrna: I like it this way, 10 lines shorter now. Is that OK?

```js
  const arrowHandler = (contentEl, viewportEl, scrollViewDirection, shiftDirection, step) => {
    let differenceX = 0;
    let differenceY = 0;

    if (scrollViewDirection === 'horizontal') {
      differenceX = shiftDirection === 'next' ? step : -1 * step;
    } else {
      differenceY = shiftDirection === 'next' ? step : -1 * step;
    }

    viewportEl.current.scrollBy(differenceX, differenceY);
  };
```

src/lib/hooks/useScrollPositionHook.js
adamkudrna: image

src/lib/hooks/useLoadResizeHook.js
adamkudrna: If we don't wait (line 19), then no, `throttleTimeout` is set to `null` when it _already was_ `null`. But if we do wait (line 14), then `throttleTimeout` is reset back to `null` once it's successfully called.

Do you think it's better to distinct these cases in the callback function?

```js
throttleTimeout = setTimeout(callBack, wait, isDelayed);
```

```js
  const callBack = () => {
    effect(getElementsPositionDifference(contentEl, viewportEl));

    if (isDelayed) {
      throttleTimeout = null;
    }
  };
```

adamkudrna:
@mbohal says:

> It will be no problem in horizontal scroll, because the arrows will be present, but for vertical scroll it might be easy to miss. While applications are expected to reset it, I think we should provide reasonable defaults.
>
> Its up to you though, feel free to resolve this if you disagree.

What's wrong with white gradient as a default value? Also, arrows are an opt-in feature. This is why scrollbars are on by default.

src/lib/components/ui/ScrollView/__tests__/ScrollView.test.jsx
adamkudrna: Well, #meetoo. I spent over an hour exploring this and as far as I've learned I don't see a way how to test an internal, non-exported module function.

If we export `arrowHandler()` which `scrollBy()` is part of, it could look like this:

```js
describe('functionality', () => {
  it('calls arrowHandler() on arrow buttons click', () => {
    const spy = jest.spyOn(ScrollView, 'arrowHandler');
    const component = mount((
      <ScrollView arrows>
        <span>content</span>
      </ScrollView>
    ));

    component.find('button').at(0).simulate('click');
    component.find('button').at(1).simulate('click');
    expect(spy).toHaveBeenCalledTimes(2);
  });
});
```

But I would like to avoid exports or even conditional exports (using `NODE_ENV` and alike) just for the sake of tests.

I'll be grateful if you show the way out of this tunnel if there is one 🙂.

mbohal: Could we test it by taking a look if `viewportEl.scrollBy` function was called?

src/lib/components/ui/Modal/Modal.jsx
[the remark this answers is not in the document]
mbohal: @adamkudrna you can still use destructuring, the `.eslintrc` line only means it is not enforced. The more we use it the less work we will have to do in #113.

I tehrefore do not think it is out of scope of this issue.

src/lib/components/ui/ScrollView/ScrollView.jsx
mbohal: How about:
```javascript
const arrowHandler = (contentEl, viewportEl, scrollViewDirection, shiftDirection, step) => {
    const offset = shiftDirection === 'next' ? step : -1 * step;
    const differenceX = scrollViewDirection === 'horizontal' ? offset : 0;
    const differenceY = scrollViewDirection !== 'horizontal' ? offset : 0;

    viewportEl.current.scrollBy(differenceX, differenceY);
  };
```

src/lib/hooks/useLoadResizeHook.js
adamkudrna: It's there.

src/lib/hooks/useScrollPositionHook.js
adamkudrna: Resolved, see above.

src/lib/hooks/useScrollPositionHook.js
adamkudrna: Fixed, see above.

src/lib/components/ui/ScrollView/ScrollView.jsx
adamkudrna: From 22 lines to 6, not bad!

mbohal: We could shorten it to something like:

```javascript
   viewportEl.current.scrollBy(['horizontal', 'vertical'].map((dir) => scrollViewDirection === dir ? (shiftDirection === 'next' ? step : -1 * step) : 0);
```
but IMHO it ios not as reaadable

src/lib/hooks/useLoadResizeHook.js
adamkudrna: It does, see below.

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -30,20 +26,30 @@ class Modal extends React.Component {
...
-      this.setState({ isContentOverflowing: true });
-    }
-  }
-
   pressEscapeHandler(e) {
-    if (e.keyCode === 27 && this.props.closeHandler) {
+    const { closeHandler } = this.props;
+
+    if (e.keyCode === 27 && closeHandler) {
       this.props.closeHandler();
```
bedrich-schindler, line 33: If you use object destructiong, replace `this.props.closeHandler();` with `closeHandler()` only

src/lib/components/ui/ScrollView/ScrollView.scss
```diff
@@ -0,0 +1,193 @@
...
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  overflow: hidden; // 2.
+
+  // 1.
+  &::before,
+  &::after {
+    @include transition((visibility, opacity, transform));
```
bedrich-schindler, line 27: Is there any reason to have it wrapped with two brackets? It is same for followings lines.

src/lib/components/ui/ScrollView/ScrollView.scss
```diff
@@ -0,0 +1,193 @@
...
+    background: var(--rui-local-end-shadow-background);
+    box-shadow: var(--rui-local-end-shadow-box-shadow);
+  }
+}
+
+.viewport {
+  position: relative; // 3.
+  width: 100%;
+  scroll-behavior: smooth;
+  -webkit-overflow-scrolling: touch;
```
bedrich-schindler, line 54: https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

> Non Standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

Is is OK for us?

src/lib/components/ui/ScrollView/__tests__/ScrollView.test.jsx
bedrich-schindler: @adamkudrna I do not see that test. Do we want to leave it without it?

src/lib/hooks/useScrollPositionHook.js
bedrich-schindler: This is still not resolved. It is necessary to use `useRef` as I wrote above. It is same for `useLoadResizeHook.js`

src/lib/components/ui/Modal/Modal.jsx
adamkudrna: Someone must have forgotten it, damn it.

src/lib/components/ui/ScrollView/ScrollView.scss
adamkudrna: Remove this line, open the demo on your iPhone and you'll see 🕶️ .

src/lib/hooks/useScrollPositionHook.js
adamkudrna: I just checked the ESlint rule, not the error itself. Fixed.

src/lib/components/ui/ScrollView/ScrollView.scss
adamkudrna: It's a list, see https://github.com/react-ui-org/react-ui/blob/master/src/lib/styles/tools/_transitions.scss.

Issue #14, opened 2019-12-09
Consider passing custom attributes to components


adamkudrna:
Currently, most components cannot have custom attributes like `className`, `aria-label` or similar. It might be useful if they could — for example, adding CSS utility classes for offsets would be easier. Optional attributes could be of any type, they would not be defined in component's props.

## Passing CSS class
Current state:
```jsx
<div className="mb-3">
  <Button label="Button with classname" />
</div>
```

~~Proposal:~~
⚠️ This is not a desired solution, it could encourage people to hack the component's inner styles.
```jsx
<Button className="mb-3" label="Button with classname" />
```

👉 Proposal: special utility props, see #18.

## Passing eg. ARIA attribute for accessibility
Proposal (currently supported only in `ButtonGroup` and `CardList`):
```jsx
<ButtonGoup aria-label="group">
  …
</ButtonGroup>
```

bedrich-schindler:
After personal discussion we agreed on passing native attributes without proptype checking. So we need to remove unnecessary prop types (like autocapitalize, autocomplete) and decide way how to pass native attributes to component.

There are two possible ways to pass native attributes to components:

1. Pass all attributes directly as props using spread operator (Material UI use follow this way)
2. Pass all native attributes via special prop type (e.x. `nativeAttributes`):

```
<TextField
  changeHandler={() => {}}
  label="Username"
  nativeAttributes={{
    autoComplete: true,
    autoCapitalize: false,
  }}
/>
```

@mbohal @adamkudrna What do you think?

mbohal:
I would prefer option 2. I think I would prefer calling it `htmlAttibutes` though. Im affraid `nativeAttributes` could wrongly make it seem related to React Native or such.

adamkudrna:
I'm for 2 named as `htmlAttributes`.

How about custom class names (the first point of the issue)? Do we want to allow that?

bedrich-schindler:
It depends on component, for example top-level element of `Button` component is `button`. Top-level element of `TextField` component is `div`, not `input`. 

Is is hard to decide but I my opinion, I aim to be able to change attributes of native element, it means that we want to pass attributes like this:

- `Button` -> `button`
- `ButtonGroup` -> `button` (passed to each button)
- `Checkbox` -> `input`
- `MultipleSelectField` -> `select`
- `Radio` -> `input` 
- `SelectField` -> `select`
- `TextField` -> `input`
- `TextArea` -> `textarea`
[...]

PR #59, opened 2020-04-22
Implement htmlElementAttribute to all UI components that internally i…

bedrich-schindler:
…mplement native form element to be able to custom props to it, remove surplus props from TextField component (#14)


mbohal:
There are still a lot of native attributes declared in the `propTypes`. For example:
* `disabled`
* `id`
* `type` (button)
* `value`
* `required`
* `checked`
* `options` (different format)
* `placeholder`
* `rows` (textarea)

The original design decision was to remove these and pass them in via `htmlElementAttribute`. Now that I'm looking at it though, I'm not so sure any more…
FYI @adamkudrna

bedrich-schindler:
> There are still a lot of native attributes declared in the `propTypes`. For example:
> 
> * `disabled`
> * `id`
> * `type` (button)
> * `value`
> * `required`
> * `checked`
> * `options` (different format)
> * `placeholder`
> * `rows` (textarea)
> 
> The original design decision was to remove these and pass them in via `htmlElementAttribute`. Now that I'm looking at it though, I'm not so sure any more…
> FYI @adamkudrna

I agree. I was not sure after Slack discussion from friday, so I've just removed obviously surplus props and left rest of them for discussion - My plan is to talk about this issue during our meeting at 15:00

mbohal:
We will take a look at other Component Libraries to see how its done. We see no good solution…

adamkudrna:
@bedrich-schindler To be closed?

mbohal:
I looked at several other libraries and they just use spread (`…`) to pass to the native element whatever they got as props.

adamkudrna:
OK guys, so let's close this PR, update #14 according to this outcome and kindly ask @bedrich-schindler to implement prop spread at least in the components where we currently need it.

Do you agree?

mbohal:
I reluctantly agree

bedrich-schindler:
Closing this pull request, assignment of issue #14 will undergo change

Issue #65, opened 2020-05-12
Generate development and production version of lib.js


bedrich-schindler:
Now we generate only production `lib.js`. We also need to generate development version of `lib.js` (probably `lib.development.js`) to be able to develop other applications with friendly named variables.

PR #66, opened 2020-05-12
Generate lib.development.js file that is not optimized and should be …

bedrich-schindler:
…used for development purposes (#65)


src/lib/components/ui/Button/Button.jsx
```diff
@@ -132,7 +132,10 @@ Button.propTypes = {
   id: PropTypes.string,
   label: PropTypes.string.isRequired,
   labelVisibility: PropTypes.oneOf(['all', 'desktop', 'none']),
-  loadingIcon: PropTypes.element,
+  loadingIcon: PropTypes.oneOfType([
+    PropTypes.bool,
```
adamkudrna, line 136: I'm not sure if I'm OK with this but if it's a common practice… To be constructive: different (yet more verbose) syntax when _using the component_ would not be the right way? Like:

```jsx
loadingIcon={isRequestPending ? <Icon icon="puppy" /> : null}
```

Sorry if we tackled this today, hope it's worth mention here, too.

webpack.config.lib.js
```diff
@@ -36,18 +39,26 @@ module.exports = (env, argv) => ({
...
-    filename: '[name].js',
+    filename: argv.mode === 'production'
+      ? '[name].js'
+      : '[name].development.js',
     libraryTarget: 'umd',
     path: path.join(__dirname, 'dist'),
     publicPath: '/dist/',
   },
+  performance: {
+    maxAssetSize: argv.mode === 'production' ? 175000 : 1250000,
```
adamkudrna, line 50: Maybe we could hold these values in constants at the beginning of the file? I think it might be more handy when needed — we suppose to update the values from time to time.

src/lib/components/ui/Button/Button.jsx
adamkudrna: … but yes, writing ternary with `children` instead of current `condition && (<JSX>)` is awkward, especially when the JSX spans multiple lines.

webpack.config.lib.js
bedrich-schindler: Yes, sure! I'm going to change it also in `webpack.config.demo.js` which is the source of this change.

src/lib/components/ui/Button/Button.jsx
bedrich-schindler: There might be better solution that I did not tested:

```
loadingIcon: PropTypes.oneOfType([
  PropTypes.component,
  PropTypes.oneOf[false],
]),
```

This should be way better because it allow component or `false`.

What do you think @mbohal? If example above works, I would like to implement it for all similar prop types that allow to accept component. I think it is only button, but i can check all of them

adamkudrna: > I think it is only button, but i can check all of them

It will be everywhere we have `children`…

mbohal: I think we should be compatible with React: https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

So the current solution seems good to me. If i read it correctly, it should handle:
* `true`
* `false`
* `null`
* `undefined`

bedrich-schindler:
Closing pull request without solution, purpose of issue #14 will undergo change.

bedrich-schindler:
> Closing pull request without solution, purpose of issue #14 will undergo change.

bedrich-schindler:
> Closing pull request without solution, purpose of issue #14 will undergo change.

Accidently closed wrong PR

bedrich-schindler:
**⚠️ ⚠️ ⚠️  Be aware that I have rebased whole branch ⚠️ ⚠️ ⚠️**

PR #69, opened 2020-05-12
Configure GitHub test workflow

adamkudrna:
Run build and tests against every push and PR.


adamkudrna:
Sorry for the force push, I forgot to include issue number in the commit message. I'm addressing this in #70.

build.sh
```diff
@@ -2,6 +2,6 @@
 
 set -e
 
-npm install
+npm ci
```
adamkudrna, line 5: See https://docs.npmjs.com/cli/ci.html

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
+name: Build and run tests
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
bedrich-schindler, line 10: I don't know this workflow and possibilities of it, but `ubuntu` latest is quite big image that we do not need. Can you check if there is only node environment like on Docker? If it is, it is definitely smaller.

bedrich-schindler:
I am not sure if we want to remove docker configuration. I would like to maintain Docker configuration for local development if necessary.

.github/workflows/test.yml
adamkudrna: It's the only Linux that [GitHub offers](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources).

adamkudrna:
> I am not sure if we want to remove docker configuration. I would like to maintain Docker configuration for local development if necessary.

I wasn't 100% sure either but I think Docker is not established/necessary in the JS world, everybody has Node on their machine. Or is it too bold to say? @mbohal?

bedrich-schindler:
@adamkudrna (reacting from mobile, unable to quote)

Reason for maintaining Docker configuration is that sometimes you can have different version installed on your computer. And best practice is to have common environment for testing that is source of true. Because sometimes, npm update can change test outputs. I had happened in the past

adamkudrna:
> Reason for maintaining Docker configuration is that sometimes you can have different version installed on your computer. And best practice is to have common environment for testing that is source of true. Because sometimes, npm update can change test outputs. It had happened in the past

Correct. But the problem is we would have _two_ sources of truth:

1. GitHub actions runner — Ubuntu
2. official Node Docker image (Debian or Alpine)

Which makes in total at least three different environments if we count also the developers' machines.

We may exactly specify the Node versions in both Docker and GH environments but it's still two different places we test against.

Or, in theory, we could change one or the other:

1. change GH actions runner to self-hosted runner (the Docker one) elsewhere — too complicated IMO
2. change our Docker image to mimic the GH actions runner — implies using Ubuntu OS

Finally, if Node versions are the problem, we can be even more specific than now in this PR: `12` in GH actions runner and `>=12.16.1` in `package.json`. I found out that the build doesn't work on Node 14 anyway. Node `12.16.3` is the current LTS so we could use this one. But it might be too restrictive for developers, I'm afraid.

.github/workflows/test.yml
mbohal: Could we please freeze the version to `ubuntu-18.04`?

Otherwise they could add newer image and break everything at any time.

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Cache node modules
+      uses: actions/cache@v1
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-v${{ matrix.node }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/test.yml') }}
```
mbohal, line 29: Is it necessary to add hashes of both `package.json` and `package-lock.json`?

mbohal, line 29: Maybe it would be more explicit to use in the key `npm` instead of `node`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
...
+        restore-keys: |
+          ${{ runner.OS }}-node-v${{ matrix.node }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/test.yml') }}
+          ${{ runner.OS }}-node-v${{ matrix.node }}-
+
+    - name: Install, build and run tests
+      run: sh ./build.sh
```
mbohal, line 35: I would prefer to list the three actions explicitly, but I guess no one else will agree with me :(

.github/workflows/test.yml
adamkudrna: (Copied from Bootstrap.) I guess that the hash of `package-lock.json` should be enough as `npm ci` installs from lockfile, not looking at `package.json` at all.

.github/workflows/test.yml
adamkudrna: OK, there you have it.

.github/workflows/test.yml
adamkudrna: I don't think so, we only have the node version at our disposal and that's the value from the matrix above.

Issue #72, opened 2020-05-13
Add `type` attribute to `actions` prop of Modal component


bedrich-schindler:

bedrich-schindler:
This problem will be fixed here: #77 ... It is better solution for future extension.

PR #73, opened 2020-05-13
Add `type` attribute to `actions` prop of Modal component (#72)

bedrich-schindler:


src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -142,6 +143,7 @@ Modal.propTypes = {
     id: PropTypes.string,
     label: PropTypes.string.isRequired,
     loadingIcon: PropTypes.element,
+    type: PropTypes.string,
```
mbohal, line 146: I think we should provide whitelist like we do here: https://github.com/react-ui-org/react-ui/blob/f949df4fef3f6aa611190e7fd431670098fc5428/src/lib/components/ui/Button/Button.jsx#L139

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -111,6 +111,7 @@ class Modal extends React.Component {
                 label={action.label}
                 loadingIcon={action.loadingIcon}
                 variant={action.variant}
+                type={action.type || 'button'}
```
mbohal, line 114: I think the default value should be handled in `defaultProps`

bedrich-schindler: It is not supported for nested data.

mbohal: True, I did not realize that it is nested.
How about:

`{...action.type && { type: action.type } }`

Although functionally same, I like it a little bit more because it does not provide default value at an unexpected place. Do as you see fit though…

bedrich-schindler:
This problem will be fixed here: https://github.com/react-ui-org/react-ui/pull/77 ... It is better solution for future extension.

PR #77, opened 2020-05-15
Transfer props to native HTML element of UI components using spread o…

bedrich-schindler:
…perator (#14)

### DIFF
--- a/src/lib/components/ui/Button/Button.jsx
+++ b/src/lib/components/ui/Button/Button.jsx
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import transferProps from '../../../utils/transferProps';
 import withForwardedRef from '../withForwardedRef';
 import styles from './Button.scss';
 
@@ -48,8 +49,15 @@ export const Button = (props) => {
     labelVisibilityClass = styles.withLabelHidden;
   }
 
+  const propsToTransfer = transferProps(
+    props,
+    ['afterLabel', 'beforeLabel', 'block', 'clickHandler', 'disabled', 'endCorner', 'grouped', 'id',
+      'label', 'labelVisibility', 'loadingIcon', 'priority', 'size', 'startCorner', 'type', 'variant'],
+  );
+
   return (
     <button
+      {...propsToTransfer}
       className={(`
         ${styles.root}
         ${priorityClass}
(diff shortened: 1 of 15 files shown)


src/lib/utils/transferProps.js
```diff
@@ -0,0 +1,11 @@
+const ALWAYS_BLACKLISTED_PROPS = ['className', 'forwardedRef'];
+
+export default (props, blacklistedProps) => {
```
bedrich-schindler, line 3: For unknown reason, `(props, blacklistedProps = [])` does not work here. That's the reason I wrote `(blacklistedProps || [])` bellow. (No time stuck on this)

src/lib/components/ui/Button/Button.jsx
```diff
@@ -48,8 +49,15 @@ export const Button = (props) => {
     labelVisibilityClass = styles.withLabelHidden;
   }
 
+  const propsToTransfer = transferProps(
+    props,
+    ['afterLabel', 'beforeLabel', 'block', 'clickHandler', 'disabled', 'endCorner', 'grouped', 'id',
```
adamkudrna, line 54: Cannot the second argument simply be `Object.keys(propTypes)`? If I understand it correctly, we always want to transfer all props except `class` and `forwardedRef`, maybe `children`. I'm just pointing this out because it seems to me we should be able to get to the same outcome in a more DRY way.

src/lib/utils/transferProps.js
```diff
@@ -0,0 +1,11 @@
+const ALWAYS_BLACKLISTED_PROPS = ['className', 'forwardedRef'];
```
adamkudrna, line 1: Maybe `children` as well?

src/lib/utils/transferProps.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default (props, blacklistedProps) => {
+  const propsToTransfer = { ...props };
+
+  [...ALWAYS_BLACKLISTED_PROPS, ...(blacklistedProps || [])].forEach((propName) => {
+    delete propsToTransfer[propName];
```
adamkudrna, line 7: Wow, thanks for learning something new.

src/lib/utils/transferProps.js
bedrich-schindler: Yes, I will add it as well

src/lib/components/ui/Button/Button.jsx
bedrich-schindler: We have discused it over mobile. We will leave it now as it is and rework it earlier. We will need to found out way how to pass props into it

PR #78, opened 2020-05-15
Use `master` as the main development branch

adamkudrna:


adamkudrna:
No, we don't want to _rebase and merge_… Just _merge_ will do 🤦.

PR #93, opened 2020-06-02
Create `FormLayout` component

### LINKED ISSUE #44: Better form layout
We need the option to align form fields **to the input field** from the left side, even when they have auto-sized labels in horizontal layout. It must also work with vertical field layout (applied also for horizontal layout on small screens).

adamkudrna:
- Create `FormLayout` component and make form fields ready for it
- Add support for React Fragments to `CardList` and `FormLayout`

Closes #44.

## Vertical Form Layout

image

**With FF grid inspector on:**

image

### Firefox – Subgrid Solution (auto-sized column with labels)

https://caniuse.com/#feat=css-subgrid

image

**With FF grid inspector on:**

image

### Webkit – Fallback Solution (fixed-width column with labels)

image

## Edge Cases

Label width is limited to 50 % of available horizontal space.

⚠️ Maximum label width and text wrapping is not available for small input sizes. It seems to be technically impossible to have both fluid label and wrapping helper text without unwanted impact on design.

Snímek obrazovky 2020-06-02 v 12 20 43

Snímek obrazovky 2020-06-02 v 12 21 12

Snímek obrazovky 2020-06-02 v 12 20 20

### DIFF
--- a/src/demo/pages/DemoContainer.jsx
+++ b/src/demo/pages/DemoContainer.jsx
@@ -21,6 +21,7 @@ import {
   CTAStart,
   CheckboxField,
   ForgotPassword,
+  FormLayout,
   LayoutCenter,
   List,
   ListItem,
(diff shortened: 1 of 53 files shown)


src/lib/components/layout/CardList/CardList.jsx
```diff
@@ -19,7 +20,7 @@ const CardList = (props) => {
       className={styles.root}
       {...other}
     >
-      {React.Children.map(children, (child) => {
+      {flattenChildren(children).map((child) => {
```
adamkudrna, line 23: Fragments are not traversed with `React.children.map()` [1][2] so I'm using the `react-keyed-flatten-children` bundle to flatten the children to a one-dimensional array. This is necessary to be able to pass props to children of these layouts.

[1] https://reactjs.org/docs/react-api.html#reactchildrenmap
[2] https://github.com/reactjs/rfcs/pull/61#issuecomment-584402735

src/demo/pages/DemoContainer.jsx
```diff
@@ -344,6 +345,190 @@ class DemoContainer extends React.Component {
...
+            <h3 id="layout-components-form-layout" className="typography-size-4 mb-6">Form Layout</h3>
+            <p>
+              Vertical Form Layout works similar to List except that List Items are not needed.
+              Vertical form field layout is forced.
+            </p>
+            <Documentation
+              name="Vertical Form Layout"
+              component={(
+                <FormLayout>
+                  <>
```
adamkudrna, line 357: This is an example of `React.Fragment` that we need to be able to traverse for the `Form Layout` to work even with form fields nested in fragments.

adamkudrna:
Nobody noticed the tests were missing? Shame on you! 🙂

adamkudrna:
## Updates:

- I added options to define custom **label width per instance**.
- **Custom auto-width fallback** is now also possible **per instance**.
- I added **fixed label width option** and made it the default value — now when we support other options than `auto` it seems to be better to offer a default that works in all browsers.

Snímek obrazovky 2020-06-04 v 23 15 54

Issue #100, opened 2020-06-08
Perform primary action and close `Modal` on enter


adamkudrna:
`Modal`s should perform primary action and close on enter, just like they close on escape.

PR #107, opened 2020-06-11
Improve Modal component to auto focus first available form field elem…

bedrich-schindler:
…ent or submit button if there no form fields present and allow Modal to be submitted by Enter key (#100)

### DIFF
--- a/src/lib/components/ui/Modal/Modal.jsx
+++ b/src/lib/components/ui/Modal/Modal.jsx
@@ -18,16 +18,41 @@ class Modal extends React.Component {
       isContentOverflowing: false,
     };
 
+    this.childrenWrapperRef = React.createRef();
+    this.submitButtonRef = React.createRef();
+
     this.setGradient = this.setGradient.bind(this);
-    this.pressEscapeHandler = this.pressEscapeHandler.bind(this);
+    this.keyPressHandler = this.keyPressHandler.bind(this);
   }
 
   componentDidMount() {
-    window.document.addEventListener('keydown', this.pressEscapeHandler, false);
+    const { autoFocus } = this.props;
+
+    window.document.addEventListener('keydown', this.keyPressHandler, false);
+
+    // If `autoFocus` is set to `true`, following code finds first form field element
+    // (input, textarea or select) or submit button and auto focuses it. This is necessary
+    // to have focus on one of those elements to be able to submit form by pressing Enter key.
+    if (autoFocus && this.childrenWrapperRef && this.childrenWrapperRef.current) {
+      const childrenWrapperElement = this.childrenWrapperRef.current;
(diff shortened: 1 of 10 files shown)


src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -152,6 +190,7 @@ class Modal extends React.Component {
 
 Modal.defaultProps = {
   actions: [],
+  autoFocusDisabled: false,
```
adamkudrna, line 193: I'd prefer to turn it around: `autoFocus: false,`

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -37,9 +62,19 @@ class Modal extends React.Component {
   }
 
   pressEscapeHandler(e) {
+    const { actions } = this.props;
+
     if (e.keyCode === 27 && this.props.closeHandler) {
       this.props.closeHandler();
     }
+
+    if (e.keyCode === 13 && e.target.nodeName !== 'BUTTON') {
```
adamkudrna, line 71: Does it also handle the numpad enter? I don't have such keyboard so I cannot test it.

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -37,9 +62,19 @@ class Modal extends React.Component {
   }
 
   pressEscapeHandler(e) {
```
mbohal, line 64: this method handles more then escape, should be more generic

src/lib/components/ui/Modal/Modal.jsx
bedrich-schindler: I am somewhere between true and false as default. But I would say we want to have it enabled by default, because there is no reason to let it off. I will always behave better by default. And if someone want to control it manually that it can be disabled.

src/lib/components/ui/Modal/Modal.jsx
bedrich-schindler: It does, there is no difference.

bedrich-schindler:
@adamkudrna @mbohal It should be ready for approval now.

src/lib/components/ui/Modal/Modal.jsx
adamkudrna: @bedrich-schindler I was rather preferring `false` but your last sentence convinced me.

mbohal: I wish someone would do the thinking for me…

adamkudrna: @mbohal In general? Or just here? 🙂

mbohal: @adamkudrna in general, i find it hard and exhausting

PR #108, opened 2020-06-12
Improve form field sizing

### LINKED ISSUE #101: Handle custom input size with CSS to fix different behaviour across browsers
## Problem

It seems that implementation of `input` `size` attribute differs across browsers: `input`s are narrower in Chrome than in Firefox and Safari:

image

↑ Input with `size` of `16`, Chrome.

image

↑ The same field in Firefox.

This is [probably caused by vague specification of HTML](https://stackoverflow.com/a/27315464).

## Suggested Solution

There is probably no pure HTML + CSS solution, CSS cannot apply the value of `size` attribute anywhere else than in `content`.

However, we can take the value of `inputSize` and pass it to CSS through a custom property. Setting a width should then be as easy as this:

```css
width: calc(1ch * var(--rui-local-input-size));
```

Or probably more purely:

```css
width: var(--rui-local-input-size);
```

### LINKED ISSUE #102: Limit width of long form fields with manual `size`
Inputs should shrink automatically when their `size` is too big to fit their container.

image

adamkudrna:
- Limit width of form fields to prevent overflowing their container (closes #102)
- Handle custom input size with CSS to fix different behaviour across browsers (closes #101)

### DIFF
--- a/src/lib/components/ui/TextField/TextField.jsx
+++ b/src/lib/components/ui/TextField/TextField.jsx
@@ -65,6 +65,16 @@ export const TextField = (props) => {
     rootVariantClass = styles.rootVariantOutline;
   }
 
+  const inlineStyle = (inputSize) => {
+    const style = {};
+
+    if (inputSize) {
+      style['--rui-custom-input-size'] = inputSize;
+    }
+
+    return style;
+  };
+
   const propsToTransfer = transferProps(
     props,
     ['changeHandler', 'disabled', 'fullWidth', 'helperText', 'id', 'inFormLayout', 'inputSize', 'isLabelVisible',
@@ -86,6 +96,7 @@ export const TextField = (props) => {
       `).trim()}
       htmlFor={props.id}
       id={`${props.id}__label`}
+      style={inlineStyle(props.inputSize)}
     >
       <div
         className={(`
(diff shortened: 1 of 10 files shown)


bedrich-schindler:
I generally agree with this PR but I have few thing in my mind.

1. You changed `TextField` to accept input size. But there are other input fields so what about them?
2. There is no example in demo part of repository. Don't we want to show it it demo?

src/lib/components/ui/TextField/TextField.jsx
```diff
@@ -65,6 +65,16 @@ export const TextField = (props) => {
     rootVariantClass = styles.rootVariantOutline;
   }
 
+  const inlineStyle = (inputSize) => {
```
bedrich-schindler, line 68: Test is missing for this feature

adamkudrna:
> 1. You changed `TextField` to accept input size. But there are other input fields so what about them?

Only `TextField` accepts input size, the HTML `size` attribute is not available for any other form elements.

> 2. There is no example in demo part of repository. Don't we want to show it it demo?

Yes it is:
image
image

src/lib/components/ui/TextField/TextField.jsx
adamkudrna: It's been already included in _all props_ test suite. Is it necessary to test it separately?

bedrich-schindler: Then is ok.

PR #117, opened 2020-07-02
Create `link` variant of `Button`

### LINKED ISSUE #116: Create `link` variant of `Button`
It proves necessary there should be a component which has semantics of `<button>` but appears as link. Creating `link` variant of `Button` seems to be the best way. Styles should be shared with link styles.

adamkudrna:
Closes #116 

image
image
image


src/lib/components/ui/Button/Button.jsx
```diff
@@ -5,48 +5,79 @@ import withForwardedRef from '../withForwardedRef';
 import styles from './Button.scss';
 
 export const Button = (props) => {
```
mbohal, line 7: I would write it as:
```
export const Button = ({
    afterLabel,
    beforeLabel,
    block,
    clickHandler,
    disabled,
    endCorner,
    forwardedRef,
    grouped,
    id,
    label,
    labelVisibility,
    loadingIcon,
    priority,
    size,
    startCorner,
    type,
    variant,
  }) => {
```

adamkudrna: I see, but… really? That's yet another approach, not yet used anywhere in the library…

mbohal: I think the suggested approach is more concise. I don't think it matters, that it it not used anywhere else - its not a big thing, just syntactic sugar.

mbohal: It is not possible due to `transferProps` functionality.

Issue #60, opened 2020-04-23
Create AutoScroll component


bedrich-schindler:
This issue is blocked by issue #53 (https://github.com/react-ui-org/react-ui/pull/58)

We need to create AutoScroll component that will be able to automatically scroll in selected direction when its content its updated. This functionality will be applied only if user will be scrolled at the edge of viewport. Component should detect change of content, it must not be detected by x or y coordinate because it will not work in case we will remove and add data at the same time.

We can create new AutoScroll component or implement it into ScrollView.

PR #118, opened 2020-07-06
Implement auto scroll into `ScrollView` (#60)

bedrich-schindler:


src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -75,6 +77,22 @@ export const ScrollView = (props) => {
     debounce,
   );
 
+  const autoScrollChildrenHash = autoScroll && children && React.Children
+    .map(children, (child) => child.key)
+    .reduce((reducedHash, childHash) => reducedHash + childHash, '');
```
mbohal, line 82: I think the names `reducedHash` and `childHash`  are misleading. They are no hashes but the `key` attributes.

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -75,6 +77,22 @@ export const ScrollView = (props) => {
...
+  const autoScrollChildrenLength = autoScroll && children && children.length;
+
+  useLayoutEffect(() => {
```
mbohal, line 85: This is not working for me reliably. I tested it using the documentation demo and after a few clicks the component did not scroll to the end. I was expecting to always see the last added text and that is not happening.

Attached is a screenshot of state that I repeatably achieved in both Chrome and Firefox (on Linux) by clicking the `Add text into scroll view` only.

Screenshot from 2020-07-07 15-45-29

src/demo/pages/DemoContainer.jsx
```diff
@@ -69,9 +69,24 @@ const loggerClick = () => console.log('click'); // eslint-disable-line no-consol
 const submitter = () => console.log('submitted'); // eslint-disable-line no-console
 
 class DemoContainer extends React.Component {
+  static generateRandomString() {
+    const texts = [
+      'Lorem ipsum dolor sit amet, consectetuer adipiscing elit.',
```
mbohal, line 74: Lets please use longer texts. It takes a lot of clicks to fill up the `ScrollView` area for vertical scroll on bigger screens.

src/lib/components/ui/ScrollView/ScrollView.jsx
adamkudrna: I confirm this behaviour. The vertical version is fine but the horizontal one is having problems a few clicks (Chrome/macOS).

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -75,6 +77,22 @@ export const ScrollView = (props) => {
...
+    if (!autoScroll || isScrolledAtEnd) {
+      return;
+    }
+
+    const differenceX = direction === 'horizontal' ? scrollViewViewportEl.current.scrollWidth : 0;
+    const differenceY = direction !== 'horizontal' ? scrollViewViewportEl.current.scrollWidth : 0;
+
+    scrollViewViewportEl.current.scrollBy(differenceX, differenceY);
```
adamkudrna, line 93: I'd like to see `scrollBy()` tested. (Just trolling.)

src/lib/components/ui/ScrollView/ScrollView.jsx
mbohal: I noticed the same problems for both vertical and horizontal.

bedrich-schindler: This was partly bug in existing `getElementsPositionDifference` function.  Internally, it use `getBoundingClientRect()` that sometimes returns floating point value that result in incorrect detection of the end of scrollable area that results in not-working autoscroll.

Next problem is that if animation is not completed and new content is added than end of scrollable area is not detected so it do not autoscroll. I will fix this.

bedrich-schindler:
> I know we currently need only this behaviour but at least I'd consider opt-in autoscroll to the end when content is added _and_ the viewport is scrolled somewhere in the middle, not at start, nor at the end. What do you think?

I agree.

bedrich-schindler:
@adamkudrna @mbohal It was re-worked and rebased due to amount of changes in master. A tested it in all modes (off, always, detectEnd) in Chrome, Firefox and Safari on MacOS and it worked in all cases. Hope it is finally done, I spent a lot of time on that.

src/demo/pages/DemoContainer.jsx
```diff
@@ -2182,6 +2211,91 @@ class DemoContainer extends React.Component {
...
+                      label="Add text into scroll view"
+                      clickHandler={
+                        () => {
+                          this.setState((prevState) => ({
+                            scrollView1: `${prevState.scrollView1}  ${this.constructor.generateRandomString()}`,
+                          }));
+                        }
+                      }
+                    />
+                    <CheckboxField
```
adamkudrna, line 2229: `Radio` would be better here — now I'm really unsure what effect the checkbox turns on 🙂.

image

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -75,6 +101,47 @@ export const ScrollView = (props) => {
...
+      return () => {};
+    }
+
+    const scrollViewContentElement = scrollViewContentEl.current;
+    const scrollViewViewportElement = scrollViewViewportEl.current;
+
+    const differenceX = direction === 'horizontal' ? scrollViewContentElement.offsetWidth : 0;
+    const differenceY = direction !== 'horizontal' ? scrollViewContentElement.offsetHeight : 0;
+
+    if (autoScroll === 'always') {
```
adamkudrna, line 126: I think for the `always` value an update should be also triggered after the viewport has been resized.

adamkudrna (commented):
Beautiful! I tested it in Chrome, FF, and on my iPad and it works.

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -75,6 +101,47 @@ export const ScrollView = (props) => {
...
+  const autoScrollChildrenLength = autoScroll !== 'off' && children && children.length;
+
+  /**
+   * If autoScroll is enabled, it automatically scrolls viewport element to the end of the
+   * the viewport when content is changed. It is performed only when viewport element is
+   * scrolled to the end of the viewport or when viewport element is in any position but
+   * autoScroll triggered by previous change is still in progress.
+   */
+  useLayoutEffect(() => {
+    if (autoScroll === 'off') {
```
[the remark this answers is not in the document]
bedrich-schindler: No. Effects must be executed always at the same time and order, otherwise it throws error. ESlint also checks if effect is not in condition. All conditions whether to execute hook or not are passed as deps array into effect - if teey changes, effect is called.

adamkudrna (approved):

src/lib/components/ui/ScrollView/ScrollView.jsx
```diff
@@ -214,6 +290,13 @@ ScrollView.propTypes = {
   arrows: PropTypes.bool,
   arrowsColor: PropTypes.string,
   arrowsScrollStep: PropTypes.number,
+  /**
+   * Auto scroll mechanism requires to have `key` prop set for every child present in `children`
+   * prop because it detects changes of these keys. Otherwise it will not work.
+   *
+   * See https://reactjs.org/docs/lists-and-keys.html#keys
+   */
+  autoScroll: PropTypes.oneOf(['always', 'detectEnd', 'off']),
```
mbohal, line 299: What is the difference between `always` and `detectEnd`? Please include better description.

mbohal (approved):

src/lib/components/ui/ScrollView/ScrollView.jsx
bedrich-schindler: Will update documentation of it.

always = It always autoscroll to the end of the viewport
detectEnd = It autoscroll to the end of the viewport only when end of the viewport is detected when content is changed (you need to be scrolled at the bottom/right [depend on direction] to allow autoscroll functionality)

PR #119, opened 2020-07-07
Lint markdown files to enforce consistent writing and reading experience

adamkudrna:
Let's make the “coding style” of markdown files consistent. Let's try the default settings first and see. There might be a later adjustment of line wrapping rules which have been subject to discussion recently.

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,0 +1,8 @@
+{
+  "comment": "All rules, see https://github.com/DavidAnson/markdownlint#rules--aliases",
+
+  "default": true,
+  "MD013": {
+    "line_length": 100
+  }
+}
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,3 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-[*.md]
-trim_trailing_whitespace = false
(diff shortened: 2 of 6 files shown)


.markdownlintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "comment": "All rules, see https://github.com/DavidAnson/markdownlint#rules--aliases",
+
+  "default": true
```
mbohal, line 4: I'm not sure about the default line length: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md013

IMHO 80 chars is really short. I would prefer to set it to 120 characters.

adamkudrna: IMO: 80 is the best for reading, 120 makes reading difficult, and 100 doesn't hurt neither while editing, nor while reading. Let's go with 100, do you agree?

adamkudrna: By the way, the recommended line length to achieve best possible readability is somewhere around 45–65 characters (sources may vary). 80 may be still acceptable for non-proportional type faces, 100 is on the edge. Compare `LICENSE.md` with `README.md` to see what reads easier. I'd like to get to 80 characters limit one day. But I believe it's good to get used to there is a limit first.

mbohal: I dont really care…

PR #120, opened 2020-07-09
Document how to use React UI in dev mode in application

mbohal:


README.md
```diff
@@ -74,6 +74,32 @@ To use React UI in your app:
...
+      '@react-ui-org/react-ui$': argv.mode === 'production'
+        ? '@react-ui-org/react-ui/dist/lib.js'
+        : '@react-ui-org/react-ui/dist/lib.development.js',
+    }
+  }
+  /* … */
+});
+```
+
+Then run webpack with the `--mode=production` or `--mode=development`flag.
```
adamkudrna, line 101: Please improve this either to Webpack (capital W, it's a name) or `webpack` (as a command).

README.md
```diff
@@ -74,6 +74,32 @@ To use React UI in your app:
    );
    ```
 
+### Development mode
```
adamkudrna, line 77: I think this section better fits the existing _Contributing_ section. So it would end up like this:

- …
- Contributing
  - Package Linking
  - **Development Mode**

README.md
mbohal: Please see: https://webpack.js.org/concepts/, the first paragraph. I was not sure how to write it, so I went to take a look there.

README.md
mbohal: This has nothing to do with development of React UI, and everything to do with development with React UI. It is therefore relevant to the consumers and not to the contributors.

Thus I think it is placed correctly.

adamkudrna: Yes, I realised it a while after I posted my comment. Agree.

Issue #15, opened 2019-12-09
New documentation


adamkudrna:
Current documentation is very basic. It consists of:

1. Foundation showcase (colors, typography, icons, spacing, breakpoints)
2. Layout components showcase
3. UI components showcase
4. Screens showcase
5. Utility classes showcase

## Problems:

* Developers find the docs hard to use because it doesn't contain important information like code examples or description of components' API.
* It is very difficult to maintain since it's held in a single file.
* Designers (and actually no one except contributors) cannot browse the docs because it's not served anywhere in public.

## Proposal:

* Find a way to split the docs into multiple files and screens to improve maintainability and navigation in browser.
* Provide components with code examples to help developers use the components.
* Provide components with API description to help developers use the components. This could be generated from `propTypes` comments.
* Consider using external tools like Storybook as it solves many of the problems.
* Serve the docs as a website to encourage designers use the UI library and to make it accessible to public.

Individual issues are expected to be raised as discussion continues within this issue.

adamkudrna:
https://whoisryosuke.com/blog/2018/automating-react-documentation-using-react-docgen/

adamkudrna:
> Serve the docs as a website to encourage designers use the UI library and to make it accessible to public.

Moved to #67.

PR #152, opened 2020-09-16
Adjust the size of number-type `TextField` according to the `max` attribute

### LINKED ISSUE #151: `TextField` width should reflect `min` and `max` attributes
It should work the same way as the `inputSize` prop — the input should be as wide as the longest expected value (the `max` attribute).

adamkudrna:
Number input types now adjust their width according to the `max` attribute.

image

Closes #151.


src/lib/components/ui/TextField/TextField.jsx
```diff
@@ -6,7 +6,31 @@ import styles from './TextField.scss';
 
 export const TextField = (props) => {
   const SMALL_INPUT_SIZE = 10;
-  const HAS_SMALL_INPUT = props.inputSize && props.inputSize <= SMALL_INPUT_SIZE;
+
+  const setCustomInputSizeByType = (inputType, inputSize, max) => {
```
bedrich-schindler, line 10: I do no personally prefer to have this kind of functions if they are used only once. In this case, you can set value of `customInputeSize` directly without function. If you still want to do that (not only here, but globally), I would prefer to move it out of component function (e.g. before function or into helper file). Advantage of this is that you can use [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) to not to calculate those values on every time component rerenders

src/lib/components/ui/TextField/TextField.jsx
```diff
@@ -6,7 +6,31 @@ import styles from './TextField.scss';
...
+    if (inputType === 'number' && max) {
+      return max.toString().length;
+    }
+
+    if (inputSize) {
```
bedrich-schindler, line 15: I still think about this. I would expect to have this condition as the first one. Reason is that you want to specify `max` attribute that can be entered into the field but at the same time, you could want to display only part of those numbers.

E.g. Phone number for Czech Republic can have 9 digits, so you set inputSize=9, but you allow to have max=9999999999999, because it is maximum value of phone number specified by IETF standards (i hope its IETF)

src/lib/components/ui/TextField/TextField.jsx
```diff
@@ -174,6 +190,7 @@ TextField.propTypes = {
   isLabelVisible: PropTypes.bool,
   label: PropTypes.string.isRequired,
   layout: PropTypes.oneOf(['horizontal', 'vertical']),
+  max: PropTypes.number,
```
bedrich-schindler, line 193: I don't want you to do it now, but we should start to write custom prop checks to throw error if max is set together with another than number

bedrich-schindler (changes requested):

src/lib/components/ui/TextField/TextField.jsx
adamkudrna: I was also thinking about it for a while but eventually I could not figure out any use case. For the phone number you are mentioning I would use the `tel` input type along with the `pattern` attribute. Similarly, the card number is also rather a text string that happens to consist of numbers but it actually isn't a number (`Nan`!). I'm covering this in the docs (#15).

src/lib/components/ui/TextField/TextField.jsx
adamkudrna: I agree. I'd also like to validate mutually incompatible props in the first stage and bad design decisions in the next stage.

src/lib/components/ui/TextField/TextField.jsx
adamkudrna: I favour this kind of functions if they improve readability of the code. Without this function, the constant `customInputSize` would look like this:

```js
const customInputSize = (inputType === 'number' && max)
  ? max.toString().length
  : (inputSize || null);
```

(Not so bad, surprisingly…!)

Or with `let` which I generally try to avoid:

```js
let customInputSize = null;

if (inputType === 'number' && max) {
  customInputSize = max.toString().length;
} else if (inputSize) {
  customInputSize = inputSize;
}
```

Or did you mean just to make the function anonymous?

bedrich-schindler: You can leave those functions as they now, just move them out of `export const TextField = (props) => {}` function. I think that we could/should to this everywhere. Somewhere I read that it is performing better when those helpers are not part of component function (e.g.  ` const TextField = (props) => {}`) because they are created everytime `TextField(props)` is called. Plus you can use `useMemo` (as mentioned above) not to call those helpers if props does not change.

adamkudrna: For the record: on Slack we decided to leave it as it is. We'll get back to the topic of helper functions at the next meeting.

Issue #135, opened 2020-08-10
Refactor fallback of auto-width `FormLayout` to `display: contents`


adamkudrna:
With new markup of form fields that changed recently, it should be possible to use `display: contents` fallback to get similar results as with `subgrid`. This way we should be able to get rid of manual `labelWidthFallback` values.

PR #165, opened 2020-10-06
Make `auto` and `limited` label widths of horizontal `FormLayout` work in all browsers (#135)

adamkudrna:
Refactor styles to make use of `display: contents` which just works. Fallback values are no longer necessary. 🎉

### DIFF
--- a/src/lib/components/layout/FormLayout/FormLayout.jsx
+++ b/src/lib/components/layout/FormLayout/FormLayout.jsx
@@ -11,19 +11,13 @@ export const FormLayout = (props) => {
     fieldLayout,
     id,
     labelWidth,
-    labelWidthFallback,
   } = props;
 
   if (!children) {
     return null;
   }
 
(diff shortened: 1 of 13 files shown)


src/lib/components/layout/FormLayout/FormLayout.jsx
```diff
@@ -70,9 +65,8 @@ export const FormLayout = (props) => {
         styles.root,
         fieldLayoutClass(fieldLayout),
         fieldLayout === 'horizontal' ? labelWidthClass(labelWidth) : '',
-        labelWidthFallback ? styles.hasRootCustomLabelWidthFallback : '',
       ].join(' ')}
-      style={inlineStyle(customLabelWidth, labelWidthFallback)}
+      style={getInlineStyle(customLabelWidth)}
```
mbohal, line 69: I would try to get rid of the `customLabelWidth` var and the `getInlineStyle` function. Like (not tested):

```suggestion
    {...hasCustomLabelWidth ? { style: { '--rui-custom-label-width': labelWidth }} : {}}
```

Issue #175, opened 2020-10-14
Add fullscreen size support to Modal


bedrich-schindler:

mbohal:
I think this is already implemented: https://github.com/react-ui-org/react-ui/blob/6213b0ac712df5c7ad2114f520747cbad71a5e1f/src/lib/components/Modal/Modal.jsx#L206

PR #176, opened 2020-10-14
Add fullscreen size option to Modal (#175)

bedrich-schindler:


src/lib/theme.scss
```diff
@@ -553,6 +553,8 @@
   --rui-modal-size-small-width: 20rem;
   --rui-modal-size-medium-width: 40rem;
   --rui-modal-size-large-width: 60rem;
+  --rui-modal-size-fullscreen-height: 100%;
```
bedrich-schindler, line 556: I was thinking about "hardcoding" 100% values into _varible.scss but I decided to make it this way - consumer of RUI can override his/her appearance of fullscreen modal.

bedrich-schindler:
@adamkudrna - Hustom, we have a problem ...

In fullscreen mode, I want all modal body elements to have 100% heights. In scrollMode=modal it works ok, but not for scrollMode=body. I need to set ScrollView inner content height to 100% to have 100% height for all content elements, but it is not possible now. Are you able to fix it? I know not-clean/dirty solution but there might be better one.

Example with dirty-fixed height using inspector (red content is div with height and width set to 100%):

image

mbohal (approved):

adamkudrna:
> In fullscreen mode, I want all modal body elements to have 100% heights. In scrollMode=modal it works ok, but not for scrollMode=body. I need to set ScrollView inner content height to 100% to have 100% height for all content elements, but it is not possible now. Are you able to fix it? I know not-clean/dirty solution but there might be better one.

@bedrich-schindler I've been investigating it for quite a long time and this is the best I can come up with: I recommend using the `fullscreen` size along with `scrollMode` set to `modal`. Modal content is now 100 % tall (spreads over the whole Modal body) and thanks to its `position: relative` you are able to position your content over it just the way you need.

**Why it works only with the `modal` scrolling mode:** because the usage of `ScrollView` in the `body` scrolling mode would make things more complicated rather than useful.

**Why the content is not centered out of the box:** it may not be what our users want. If you need to center your content, you can do it with a custom element that is absolutely positioned over the whole Modal content. There could be a prop for that but I think it's a very specific use case—at least from my current point of view.

adamkudrna (approved):

PR #177, opened 2020-10-16
Exclude generated docs and tests from npm bundle

adamkudrna:
The bundle shouldn't be 20 MB large 🙈.


bedrich-schindler:
I my opinion, following should be ignored as well:

* dist/docs
* RELEASING.md
* CONTRIBUTING.md

adamkudrna:
> I my opinion, following should be ignored as well:
> 
> * dist/docs
> * RELEASING.md
> * CONTRIBUTING.md

@bedrich-schindler: `dist/docs` has moved to `docs` (which is ignored), `RELEASING.md` is ignored and `CONTRIBUTED.md` is usually part of bundle—as far as I've seen in other npm bundles.

Issue #113, opened 2020-06-19
Require destructuring assignment


mbohal:
Presently we have in `.eslint` set `"react/destructuring-assignment": "off"`.

We need to enable this rule and fix all errors that will come out of it.
To enable it, removing this line should be enough as it is part of the airbnb eslint config.

PR #183, opened 2020-10-30
Extend `Grid` layout with custom rows and columns options and more (#149)

### LINKED ISSUE #149: Extend `Grid` with column options
Originally discussed in #146:

> Yes, now it may be mistaken for the usual grid component that exists in other UI libraries. The “usual grid” also has a different way of usage: you would need cells or grid items to wrap around the content you want to put in the grid. That's because how the grid is built: with flexbox. However, that's not our case. We want a **next-generation grid** component using **the right tool for the job: the grid layout.** Imagine extending the `Grid` component in the future like this:

```jsx
<Grid items="3" size="1fr">
  <div>…</div>
  <div>…</div>
  <div>…</div>
</Grid>
```
> It could go further with non-repetitive layouts like this:

```jsx
<Grid columns="minmax(200px, 300px) 1fr">
  <div>…</div>
  <div>…</div>
</Grid>
```

_Originally posted by @adamkudrna in https://github.com/react-ui-org/react-ui/pull/146#issuecomment-690319553_

adamkudrna:
image

image

image

Closes #149.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,7 +25,8 @@
       "rules": {
         "react/jsx-filename-extension": [1, { "extensions": [".mdx"] }],
         "react/jsx-indent": "off", // Gives false positives in MDX files.
-        "semi": "off" // We don't want to clutter our MDX with semicolons.
+        "semi": "off", // We don't want to clutter our MDX with semicolons.
+        "sort-keys": "off" // This rule only needs to be turned off in a few places (eg. breakpoint lists), but weirdly enough, an inline comment alerts breaking of the `indent` rule.
       }
     }
   ]
--- a/src/gatsby-theme-docz/wrapper.js
+++ b/src/gatsby-theme-docz/wrapper.js
@@ -23,6 +23,15 @@ const Wrapper = ({ children }) => (
           gtag('config', 'UA-173070814-1');
         `}
       </script>
+      <style type="text/css">
+        {`
+          @media (prefers-reduced-motion: no-preference) {
+            html {
+              scroll-behavior: smooth;
+            }
+          }
+        `}
+      </style>
     </Helmet>
     {children}
   </>
--- a/src/lib/components/layout/Grid/GridSpan.jsx
+++ b/src/lib/components/layout/Grid/GridSpan.jsx
@@ -1,0 +1,83 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { generateResponsiveCustomProperties } from './helpers/generateResponsiveCustomProperties';
+import styles from './Grid.scss';
+
+export const GridSpan = ({
+  children,
+  columns,
+  id,
+  rows,
+  ...other
+}) => {
+  if (!children) {
+    return null;
+  }
+
+  return (
+    <div
+      id={id}
+      className={styles.span}
+      style={{
+        ...generateResponsiveCustomProperties(columns, 'column-span'),
+        ...generateResponsiveCustomProperties(rows, 'row-span'),
+      }}
+      {...other}
+    >
+      {children}
+    </div>
+  );
+};
+
+/* Breakpoints are easier to work with when ordered according to their value, not name. */
+/* eslint-disable sort-keys */
+
+GridSpan.defaultProps = {
+  children: null,
+  columns: undefined,
+  id: undefined,
+  rows: undefined,
+};
+
+GridSpan.propTypes = {
+  /**
+   * Items to be aligned in the grid.
+   */
+  children: PropTypes.node,
+  /**
+   * Number of columns to span.
+   */
+  columns: PropTypes.oneOf([
+    PropTypes.number,
+    PropTypes.shape({
+      xs: PropTypes.number,
+      sm: PropTypes.number,
+      md: PropTypes.number,
+      lg: PropTypes.number,
+      xl: PropTypes.number,
+      xxl: PropTypes.number,
+      xxxl: PropTypes.number,
+    }),
+  ]),
+  /**
+   * ID of the root HTML element.
+   */
+  id: PropTypes.string,
+  /**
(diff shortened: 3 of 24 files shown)


bedrich-schindler:
Awesome! That' exactly what I needed.

bedrich-schindler (changes requested):

.eslintrc
```diff
@@ -25,7 +25,8 @@
       "rules": {
         "react/jsx-filename-extension": [1, { "extensions": [".mdx"] }],
         "react/jsx-indent": "off", // Gives false positives in MDX files.
-        "semi": "off" // We don't want to clutter our MDX with semicolons.
+        "semi": "off", // We don't want to clutter our MDX with semicolons.
+        "sort-keys": "off" // This rule only needs to be turned off in a few places (eg. breakpoint lists), but weirdly enough, an inline comment alerts breaking of the `indent` rule.
```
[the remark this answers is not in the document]
adamkudrna: I'd love to do it [locally](https://www.npmjs.com/package/eslint-plugin-markdown#configuration-comments), but it triggers another alert:

image

image

I don't know what to do with that…

.eslintrc
```diff
@@ -25,7 +25,8 @@
       "rules": {
         "react/jsx-filename-extension": [1, { "extensions": [".mdx"] }],
         "react/jsx-indent": "off", // Gives false positives in MDX files.
-        "semi": "off" // We don't want to clutter our MDX with semicolons.
+        "semi": "off", // We don't want to clutter our MDX with semicolons.
```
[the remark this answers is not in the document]
adamkudrna: Sure, the whole block applies only to MDX files:

image

src/gatsby-theme-docz/wrapper.js
```diff
@@ -23,6 +23,15 @@ const Wrapper = ({ children }) => (
           gtag('config', 'UA-173070814-1');
         `}
       </script>
+      <style type="text/css">
+        {`
```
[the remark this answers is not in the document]
adamkudrna: It wouldn't be a valid JSX:

image

src/lib/components/layout/Grid/GridSpan.jsx
```diff
@@ -0,0 +1,85 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { generateResponsiveCustomProperties } from './helpers';
+import styles from './Grid.scss';
+
+export const GridSpan = (props) => {
```
mbohal, line 6: I would prefer the destructuring syntax:

```js
export const GridSpan = ({
    children,
    columns,
    id,
    rows,
    ...other
  }) => {
```

I know it is not used elsewhere, but I find it easier to read. In fact I think we should update other files to use it as well (not in this PR obviously).

src/lib/components/layout/Grid/helpers.js
```diff
@@ -0,0 +1,19 @@
+export const generateResponsiveCustomProperties = (prop, infix) => {
```
mbohal, line 1: I would prefer to have one named export per file where the the filename is the name of the function. So this file should be named `helpers/generateResponsiveCustomProperties.js`

It makes the implementation, but especially the tests much easier to read.

src/lib/components/layout/Grid/helpers.js
```diff
@@ -0,0 +1,19 @@
...
+  if (typeof prop === 'undefined') {
+    return null;
+  }
+
+  const customProperties = {};
```
mbohal, line 6: I would replace it with:

```js
  if (typeof prop !== 'object') {
    return { [`--rui-local-${infix}-xs`]: prop };
  }

  return Object.keys(prop).reduce((acc, breakpoint) => ({
    ...acc,
    [`--rui-local-${infix}-${breakpoint}`]: prop[breakpoint],
  }), {});
```
The return early pattern makes it easier to follow as I can exit the function, well, early as the logical branches do not reconnect.

src/lib/components/layout/Grid/helpers.js
```diff
@@ -0,0 +1,19 @@
+export const generateResponsiveCustomProperties = (prop, infix) => {
+  if (typeof prop === 'undefined') {
```
mbohal, line 2: maybe we should add a unit test…

src/lib/components/layout/Grid/GridSpan.jsx
adamkudrna: > In fact I think we should update other files to use it as well (not in this PR obviously).

Sure, there is an issue for that: #113.

adamkudrna: And thank you for pointing it out, I missed it somehow this time.

src/gatsby-theme-docz/wrapper.js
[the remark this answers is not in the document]
bedrich-schindler: Ok

.eslintrc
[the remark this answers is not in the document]
bedrich-schindler: If it is for MDX only, we can leave it as it is.

src/lib/components/layout/Grid/helpers/__tests__/generateResponsiveCustomProperties.test.js
```diff
@@ -0,0 +1,29 @@
+import { generateResponsiveCustomProperties } from '../generateResponsiveCustomProperties';
+
+describe('generateResponsiveCustomProperties', () => {
+  it('when prop is undefined', () => {
```
bedrich-schindler, line 4: It is up to you. But I would use `test` here, it would make sense then.

bedrich-schindler (approved):

PR #188, opened 2020-11-05
Unify API of form components (#28, #113)

### LINKED ISSUE #28: Unify `error` vs. `helperText` props
In all form fields (`CheckboxField`, `MultipleSelectField`, `RadioField`, `SelectField`, `TextArea`, `TextField`, `Toggle`):

- rename `helperText` to `helpText`
- use `validationText` for invalid/valid/warning validation message (ie. don't share the purpose with `helperText`)
- get rid of the `error` prop
- validate both `*Text` props as string or node to enable passing custom HTML such as links or icons

The new `validationText` would be styled according to the `validationState` prop.

---

**As discussed in #20:**

Mainly, we have to unify differences between usage of `error` vs `helperText`+`validationState` across all components, not only Radio, CheckBox and Toogle. It is weird to have different approaches of passing these values into components. So this is the first part of issue we probably agree on.

Then I am in to have two different props for setting these values, `errorText` and `helpText`, so we can set both independently. With this change, we will have to remove `validationState` prop and remove old `error` prop. And I also tend to use `helpText` instead of `helperText`.

The only thing I am not sure about is if we will need some help icon or not. If yes, then we will have to have prop for help icon, something like `helpIcon`. But it is only weak point I see now. Maybe icon is not necessary. It is totally up to @adamkudrna.

I am still thinking about icons because it is not so handy to pass all these icons like sortingAscIcon, sortingDescIcon, helpIcon (if it is needed) to components all the time. And in case we will need to have helpIcon then I would like to open discussion about`IconContext` component. We already have `TranslationContext` that can pass translations to components globally and it can be same for icons.

_Originally posted by @bedrich-schindler in https://github.com/react-ui-org/react-ui/issues/20#issuecomment-591394870_

> have two different props for setting these values, errorText and helpText
I like these names

> The only thing I am not sure about is if we will need some help icon or not.

Could we make `helpText` and `errorText` accept nodes and/or text and then we could pass in the icons inline? That would open other possibilities such as passing into `helperText` a button that opens documentation modal and such things.

Something like (I didnt test this code):
```jsx
<MyComponent
  helpText={<><Icon type="help"> I'm the help text.</>}
/>
```

_Originally posted by @mbohal in https://github.com/react-ui-org/react-ui/issues/20#issuecomment-591507251_

### LINKED ISSUE #113: Require destructuring assignment

adamkudrna:
- unify API (closes #28)
  - change `error` to `validationText`
  - change `helperText` and `description` to `helpText`
  - make sure both of the new props are supported in all supported components
- `.eslintrc` clean up and related changes (closes #113)


src/lib/components/ui/TextField/TextField.jsx
```diff
@@ -1,153 +1,107 @@
...
+    type,
+    validationState,
+    validationText,
+    value,
+    variant,
+  } = props;
 
-    return style;
-  };
+  const SMALL_INPUT_SIZE = 10;
```
bedrich-schindler, line 34: This could be moved outside of this component.

bedrich-schindler (approved):
I went through changes, tested documentation and it seems it is fixed correctly everywhere. Good job!

PR #192, opened 2020-11-11
Simplify syntax of transferred props (#88)

### LINKED ISSUE #88: Make transferred props more DRY
Related to #86.

It would be handy to combine the definition of props to transfer with `propTypes`. Currently it's white-listed aside like here in `Button`:

```js
  const propsToTransfer = transferProps(
    props,
    ['afterLabel', 'beforeLabel', 'block', 'clickHandler', 'disabled', 'endCorner', 'grouped', 'id',
      'label', 'labelVisibility', 'loadingIcon', 'priority', 'size', 'startCorner', 'type', 'variant'],
  );
```

Code could be also more DRY and easier to write and maintain then.

adamkudrna:
Closes #88.

### DIFF
--- a/src/lib/components/ui/Button/Button.jsx
+++ b/src/lib/components/ui/Button/Button.jsx
@@ -4,27 +4,26 @@ import transferProps from '../../../utils/transferProps';
 import withForwardedRef from '../withForwardedRef';
 import styles from './Button.scss';
 
-export const Button = (props) => {
-  const {
-    afterLabel,
-    beforeLabel,
-    block,
-    clickHandler,
-    disabled,
-    endCorner,
-    forwardedRef,
-    grouped,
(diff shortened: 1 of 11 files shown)


bedrich-schindler (approved):
There is only one thing to think about. I would recommend to have propsToTransfer as the first props of the component, not as last. Reason is that in case of any problem, transfered props will not override explicitly defined props of specific component.

adamkudrna:
> There is only one thing to think about. I would recommend to have propsToTransfer as the first props of the component, not as last. Reason is that in case of any problem, transfered props will not override explicitly defined props of specific component.

@bedrich-schindler But wouldn't there be a conflict of duplicate props in such case? I'll do as you say anyway, you're the expert 🙂.

PR #198, opened 2020-11-13
Fix misaligned help and validation texts of `CheckboxField`, `Radio` and `Toggle` inside `FormLayout`

adamkudrna:
- Fix mentioned form fields
- Improve `transferProps()` function

### DIFF
--- a/src/lib/utils/transferProps.js
+++ b/src/lib/utils/transferProps.js
@@ -1,11 +1,7 @@
-const ALWAYS_BLACKLISTED_PROPS = ['children', 'className', 'forwardedRef', 'staticContext'];
-
-export default (props) => {
-  const propsToTransfer = { ...props };
-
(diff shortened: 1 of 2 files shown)


src/lib/utils/transferProps.js
```diff
@@ -1,11 +1,7 @@
...
-export default (props) => {
-  const propsToTransfer = { ...props };
-
-  ALWAYS_BLACKLISTED_PROPS.forEach((propName) => {
-    delete propsToTransfer[propName];
-  });
-
-  return propsToTransfer;
-};
+export default ({
```
bedrich-schindler, line 1: Just WOW!

bedrich-schindler (approved):

src/lib/utils/transferProps.js
adamkudrna: ❤️ goes to @mbohal!

PR #215, opened 2021-01-21
Extend `RUIProvider` with `globalProps` to pass props to all UI and layout components using provider, create Component Customization page in documentation

### LINKED ISSUE #214: Allow default prop values to be changed via context
Some projects need different prop defaults then others. Presently the only ways to achieve this are:

* by specifying the desired value for every instance
* to wrap the react-ui component ion another component that would specify the desired defaults

A more elegant way would be to allow the defaults to be passed into the components via a React context. It should be done in a similar manner as translations are handled now.

bedrich-schindler:
Closes #214


src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+```js
+import { RUIProvider } from '@react-ui-org/react-ui';
+```
+
+Then wrap application (or its part) with `RUIProvider` with `globalProps` prop object,
+where its keys are names of components and its values are objects with arbitrary
+props you want to pass to a specific component.
+
+Keys conform to actual names of components:
```
mbohal, line 42:
```suggestion
Then wrap application or its part with `RUIProvider` with defined `globalProps` attribute holding an object.
The keys are names of the components and its values are objects with arbitrary
props you want to pass to the specified components.

```

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+<RUIProvider globalProps={{
+  Button: { priority: 'filled' },
+  TextField: { variant: 'filled' },
+  SelectField: { variant: 'filled' },
+}}>
+  //...
+</RUIProvider>
+```
+
+See working example:
```
mbohal, line 54: I like the working example :heart:

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+              label="Button"
+            />
+          </ToolbarItem>
+        </Toolbar>
+      </RUIProvider>
+    );
+  }}
+</Playground>
+
+### Best practices
```
mbohal, line 105: I'm not sure about this section.

For example, when my entire project is just one `<Form>` I can imagine something  like:
```
<RUIProvider globalProps={{
  TextField: { clickHandler: setValue },
}}>
```
I would prefer to leave this up to the discretion of the developer using this feature.

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+to every component you want. It has been made to simplify repeatedly
+set props that are common for your whole project.
+
+For that reason **we do recommend to set only global props which have
+impact on appearance**. Do not use it for props that are used to be
+unique, e.g `clickHandler`,`closeHandler`, `disabled`, `id`, `label`,  etc.
+
+## Translations
+
+Some components may contain tooltips which improve components' accessibility.
```
mbohal, line 117:
```suggestion
Some components may contain texts which improve components' accessibility.
```

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+All texts are in English by default and can be translated to other languages.
+
+### Translation Approaches
```
mbohal, line 120: I would consider moving translation documentation to a separate page. Its close relatation to globalProps is in my view only an implementation detail.

mbohal (approved):

src/lib/components/ui/ScrollView/__tests__/__snapshots__/ScrollView.test.jsx.snap
```diff
@@ -1,80 +1,78 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`rendering renders correctly with a single child 1`] = `
-<WithTranslationContextComponent
```
adamkudrna, line 4: And also here, only deletions — is it OK?

adamkudrna (approved):

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+import {
+  Toolbar,
+  ToolbarItem,
+} from '../../lib/components/layout/Toolbar';
+import Button from '../../lib/components/ui/Button';
+import SelectField from '../../lib/components/ui/SelectField';
+import TextField from '../../lib/components/ui/TextField';
+import { RUIProvider } from '../../lib/provider';
+
+React UI often provides more variants of components, so a user can
```
adamkudrna, line 21: > Many React UI components are available in more variants so users can pick what suits them the best.

…?

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+pick the most suitable variant.
+
+There two ways to configure component:
```
adamkudrna, line 24: > There are two ways …

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+1. Pass props directly to a specific component.
+2. Wrap application (or its part) with `RUIProvider` and pass arbitrary props for
+a specific component through `globalProps` prop.
```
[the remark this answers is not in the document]
adamkudrna: I like Martin's wording of the second option but I wouldn't delete the first one. Only number 2 is in scope of this issue but number 1 is generally still valid, IMO.

src/docs/customize/component-customization.mdx
adamkudrna: Maybe _visually hidden texts_ then…?

src/docs/customize/component-customization.mdx
adamkudrna: Yes, @bedrich-schindler please create a separate page for this and link the pages through.

src/docs/customize/component-customization.mdx
adamkudrna: and their values

src/docs/customize/component-customization.mdx
```diff
@@ -0,0 +1,184 @@
+---
+name: Component Customization
+menu: Customize
+route: /customize/component-customization
+---
+
+# Component Customization
+
+## Global props
```
adamkudrna, line 9: Please stay consistent in capitalizing: _Global Props_, also see other headings below.

src/docs/customize/component-customization.mdx
adamkudrna: Maybe a different wording would help? Something like _however there are component props that are usually unique to their instance_…

src/docs/customize/component-customization.mdx
adamkudrna: Yep this is cool!

adamkudrna (commented):
In the second CR round I'd only suggest some minor changes in the docs. Otherwise, well done!

adamkudrna (changes requested):
(Just to make sure this isn't merged before the docs tweaks are addressed.)

src/docs/customize/component-customization.mdx
mbohal: They are hidden now, but they could be visible in the future.

I think that in this case being more specific does not help in any way. All texts, hidden or visible, have to be translated and when setting up react-ui in multi-language setting I do need to know this information.

src/docs/customize/component-customization.mdx
mbohal: But why?

All props are handled exactly the same way. There is no technical reason for it. As far as non technical reasons that depends on the application. Sometimes it might be a terrible idea and sometimes it can be a simple and elegant solution.  Only the application author knows what props are to be unique and what props  can be shared in the given context.

`usually unique` always depends for whom... if all I ever do are single page three field forms over and over then my definition is very different from someone doing complex administrations.

src/docs/customize/component-customization.mdx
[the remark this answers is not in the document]
mbohal: How about adding the information to the end:
```
…
and passing in the common props through `globalProps` prop.
With or without using `RUIProvider` the component can be configured by passing in the props directly.
The props passed in directly always take precedence over props coming from `RUIProvider`.
```

I find the overloading bit important.

adamkudrna:
@bedrich-schindler One more thing: could you please rename the PR to something more descriptive? It will be used in the changelog. Thanks!

bedrich-schindler:
@mbohal Following is not possible:

```
 <Component
    {...contextGlobalProps}
    translations={contextTranslations}
    {...props}
  />
```

Tests fail with translations set to null (that is component default). If it is this way and null is passed to component then it does not work. 

```
translations={props.translations || contextTranslations}
```

This works because it tries to use translation from props and if falsy content is passed, contextTranslation (that also contains defaults) is used.

src/docs/customize/global-props.mdx
```diff
@@ -0,0 +1,104 @@
...
+import { RUIProvider } from '../../lib/provider';
+
+Sometimes it can be useful to share some configuration props between
+multiple instances of components of the same or different kind.
+
+This can be achieved by wrapping application or its part with `RUIProvider`
+and passing in the common props through `globalProps` prop. With or without
+using `RUIProvider` the component can be configured by passing in the props
+directly. The props passed in directly always take precedence over props
+coming from `RUIProvider`.
```
adamkudrna, line 26:
```suggestion
This can be achieved by wrapping application or its part with `RUIProvider`
and passing in common props through the `globalProps` prop. Regardless of
using `RUIProvider`, the component can still be configured by passing in the props
directly. The props passed in directly always take precedence over props
coming from `RUIProvider`.
```

Just my feelings, use it or not as you see fit.

src/docs/customize/translations.mdx
```diff
@@ -6,22 +6,70 @@ route: /customize/translations
...
 
-1. You can use `TranslationProvider` component and pass translated structure
-   from this file into `TranslationProvider` as `translations` props to
-   overwrite default translations.
+1. You can use `RUIProvider` component and pass translated structure
+   from this file into `RUIProvider` as `translations` prop to
+   overwrite default translations. You have to define all translations for all
+   components, otherwise it will not work.
+2. Other option is to overwrite translations locally using `translations` prop
+   of specific component. You have to define all translations as well.
```
adamkudrna, line 23:
```suggestion
   of specific component. You have to define all translations of that component as well.
```

src/docs/customize/translations.mdx
```diff
@@ -6,22 +6,70 @@ route: /customize/translations
...
 
-2. Other option is to overwrite translations locally using `translations` props
-   of specific component.
+### Local Translations
+
+If component supports translations, `translations` prop and its structure is
+present in props documentation for a specific component.
```
adamkudrna, line 28:
```suggestion
If a component supports translations, `translations` prop and its structure is
present in props documentation for the specific component.
```

adamkudrna (approved):

PR #221, opened 2021-02-03
Make code font family and font size customizable

adamkudrna:
Code font family and font size are now customizable.

These settings are also reflected in docs where the code block appearance has been unified and slightly tweaked to work better with bigger font size of the docs.

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -63,6 +63,12 @@ export default {
         },
       },
     },
+    editor: {
+      fontSize: 1,
+    },
+    fonts: {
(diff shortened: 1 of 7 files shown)


bedrich-schindler:
@adamkudrna What about including this into docs?

bedrich-schindler (approved):

adamkudrna:
> @adamkudrna What about including this into docs?

@bedrich-schindler Currently there are more substantial things missing from the docs like the entire typography etc. The plan is to cover all custom properties with docs in the future and these options will be included too.

PR #230, opened 2021-02-08
Change UI component exports to fix `withForwardedRef` (#229)

### LINKED ISSUE #227: Update dependencies

### LINKED ISSUE #229: Fix exports of components to work correctly with forwardedRef
Components are now exported incorrectly. All components must be wrapped first with `withForwardedRef` and then with rest (e. g. `withProviderContext`).  Otherwise following warning is shown:

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

bedrich-schindler:
Closes #229


adamkudrna:
> Closes #227

@bedrich-schindler I'm not sure this is related to #227… Correct me if I'm wrong.

adamkudrna (approved):
Please just have a look at the issue link and it's perfect.

Issue #236, opened 2021-02-26
Optimize form field custom properties (to slightly reduce bundle size)


adamkudrna:
Not all generated custom properties are actually used by all components.

Issue #224, opened 2021-02-04
Document theming options


adamkudrna:
Preferably generate docs from custom properties definitions.

Issue #163, opened 2020-10-06
Rework theme defaults


adamkudrna:
1. Change default theme to CSS custom property fallback values (See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Custom_property_fallback_values))
   - This allows to omit the default theme without broken appearance.
   - It's still necessary to keep all default values in a single file so users are able to see all available options at once.
2. ~~Use BEM-like notation, eg.:~~ Moved to #243.
   - `--rui_form-field_disabled_label-and-validation-text_color` instead of `--rui-form-field-disabled-label-and-validation-text-color`, or
   - `--rui__form-field--outline--default__border-color` instead of `--rui-form-field-outline-default-border-color`.

adamkudrna:
## Expected result

- Separate design _options_ from _decisions_.
- Bake the decisions into the code and make any further theming steps _optional_.

## Suggested steps to take

1. **Move foundation theme settings** from `theme.scss` to new **design tokens** layer and place it as the first layer in `foundation.scss`. This includes:
  * colors
  * breakpoints
  * spacing
  * borders
  * typography
2. **Figure out** what to do with **basic design decisions:** page (text?) color and background, link color and decoration, list style (?).
3. **Document** foundation settings as **design tokens.**
4. **Move component theme settings** from `theme.scss` to corresponding `_theme.scss` inside component source.
5. **Document component theme options.**

adamkudrna:
Nothing more appears necessary after #243 has been done.

PR #238, opened 2021-02-26
Refactor `Radio`, `CheckboxField` and `Toggle` and make them themeable (#20)

### LINKED ISSUE #20: Refactor Radio, Checkbox and Toggle
- ~~Unify API with other form fields: `error` → `helperText`~~ moved to #28
- Get rid of deprecated SCSS settings and tools (colors, forms, icons)

adamkudrna:
Here it comes 🎊. What has changed? For these components, basically everything.

- **Form fields layout** has been rewritten and extended so all fields are built using the same principles. Their behaviour is unified whether they are used inline or wrapped in `Toolbar` or in `FormLayout`.
- Improve label and help/validation **text wrapping** in context of `FormLayout`.
- `Radio`, `CheckboxField` and `Toggle` have been **completely rewritten** from the ground up to share the same visual principles and theming options with other form components. Default appearance has been made **more generic** and unopinionated.
- **Animated transition** between checked/unchecked states.
- Reported visual **glitches** have been **fixed**.
- Form fields theming is now fully **documented**.
- Finally, last big portions of aged **deprecated code** have been **removed**.

## Before

image
image
image
image
image

## After

Mind the help text wrapping — fields are now inline:

image
image
image
image

Help texts and validation texts make use of available space in context of `FormLayout`:

image

Getting back to the original appearance is possible:

image

### Validation State Styling

image

### Disabled State Styling

image

Closes #20.


webpack.config.js
```diff
@@ -2,8 +2,8 @@ const path = require('path');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
 const VisualizerPlugin = require('webpack-visualizer-plugin');
 
-const MAX_DEVELOPMENT_OUTPUT_SIZE = 1850000;
-const MAX_PRODUCTION_OUTPUT_SIZE = 210000;
+const MAX_DEVELOPMENT_OUTPUT_SIZE = 2000000;
```
mbohal, line 5: Why the size increse? This PR should have a negative imapct:

Screenshot from 2021-02-26 14-21-56

src/docs/customize/theming.mdx
```diff
@@ -422,13 +328,236 @@ Example:
 }}
 </Playground>
 
-### Form Layout
+### Check Fields
+
+Options shared by checkable form controls. This includes
```
mbohal, line 333: I would change `checkable` to `boolean`.

src/docs/customize/theming.mdx
```diff
@@ -87,14 +85,21 @@ Theming options affecting all form components.
 | `--rui-form-field-required-sign-color`               | Color of text appended to required input labels              |
```
mbohal, line 85: It might be beyond the scope of this issue, but I think it would be helpful, if the custom property table would also include data type and example and default if applicable.

I guess I could look in the default theme, but that is not very practical.

mbohal (approved):

webpack.config.js
adamkudrna: It's because of new theming options for refactored components. Until now, the design of `CheckboxField`, `Radio` and `Toggle` was hard-coded to meet the requirements of _our project_. Now the default appearance is less opinionated while it's still perfectly possible to get back to the original design through theming options, just like with any other component.

However, I believe there is a [chance to make the bundle slightly smaller](https://github.com/react-ui-org/react-ui/issues/236) (but not drastically).

src/docs/customize/theming.mdx
adamkudrna: It was under consideration by me, but:

- `boolean` is data type while my point of view is more about design/appearance (refer to `box` fields instead of `text` fields.
- `boolean` would be misleading for `Radio` – its value is rather _checkable_ than `boolean`.

webpack.config.js
mbohal: The new version has less lines of code yet it produces a larger package :thinking:

src/docs/customize/theming.mdx
mbohal: OK, I concur

src/docs/customize/theming.mdx
adamkudrna: Thanks for pointing it out. To be honest, I'm still not convinced of maintaining the theming docs manually, it's a _lot_ of work. There is an idea of generating the docs for it automatically but it's really an early concept based on discussions of @bedrich-schindler and me.

https://github.com/react-ui-org/react-ui/issues/224

webpack.config.js
adamkudrna: Less source SCSS, more output CSS.

src/lib/components/ui/Toggle/README.mdx
```diff
@@ -192,3 +192,18 @@ interfere with the API, and they will be passed to the `input` HTML element.
...
 <Props table of={Toggle} />
+
+## Theming
+
+Head to [Forms Theming](/customize/theming/forms) to see shared form theming
+options. On top of that, the following options are available for Toggle.
+
+| Custom Property                                                    | Description                                    |
+|--------------------------------------------------------------------|------------------------------------------------|
+| `--rui-form-field-check-input-toggle-width`                | Input width (height is shared with other check fields) |
```
bedrich-schindler, line 203: Isn't it possible to make it shorted? Why `check` is in the class?

bedrich-schindler:
I've been trying to go through whole PR, but there are just too many changes in CSS code I have not able to check, otherwise I would have spent hours on that. But I ran it and it works fine as well,.

bedrich-schindler (approved):

src/lib/components/ui/Toggle/README.mdx
adamkudrna: Custom properties naming is a bit (a lot) complicated:

```
- project prefix (rui)
  - global property type (color) | component name (button) | component type (form-field)
    - forms only, optional: form field category (check | box*)
    - all components, optional: variant | priority | feedback state
      - optional: element name
        - optional: element type
          - interaction state
            - property name
```

It's a very approximate illustration but this is about how it works.

From this point of view I should even spread `box` over the box-field-related custom properties (and not delete `check`) to be consistent with what I say here * 🙂.

So it seems I should revisit all form custom properties and either use the convention everywhere — or flatten them all as much as possible without compromising their descriptiveness (box input size vs. check input size etc.).

adamkudrna: Relates to #163.

adamkudrna: OK, I'll unify naming here and change the syntax in #243 so it's more readable.

Issue #218, opened 2021-01-27
event handler props


mbohal:
Presently RUI uses `*Handler` format for passing actions to interactive components. For example we have `clickHandler`, `closeHandler` etc.  This causes the following problems:

* We have to pass this prop manually tot he native element (see: https://github.com/react-ui-org/react-ui/blob/120eaedbe2134c844ba2d987e00313015c549a3c/src/lib/components/ui/TextArea/TextArea.jsx#L99)
* It is inconsistent with other React [synthetic events](https://reactjs.org/docs/events.html) naming convention

I propose we:

* Switch `*Handler` syntax to `on*` syntax
* Pass all synthetic event handlers implicitly via `ransferProps`, that is get rid of `clickHandler` and such

This will lead to:
* removing some props (`clickHandler`, `changeHandler`, maybe others)
* renaming some non event props such as `closeHandler` to `onClose`: https://github.com/react-ui-org/react-ui/blob/456598417add3d890cd65d46ad75ac841431f7b1/src/lib/components/ui/Modal/Modal.jsx#L273

bedrich-schindler:
I vote for renaming `clickHandler` -> `onClick` (as others), I think that this original idea with `<event>Handler` was not good. Otherwise I think that we must preserve those handlers in props. It means to rename them, not to delete them. They cannot be omit from props documentation.

mbohal:
Decided on video call with @bedrich-schindler and @adamkudrna :

We will do it as proposed. Since `clickHandler/onClick` is not mandatory, we do not need to include it explicitly.

PR #245, opened 2021-03-05
Create `FileInputField` component (#219)

### LINKED ISSUE #219: File Input field
Presently RUI does not include a file input (`<input type="file" />`)

This causes:

1. file inputs are not consistent with the rest of the UI
2. rendering of supporting elements such as `label`, `helpTest`, `validationText` is difficult

If solving point 1) would be too difficult or time consuming for now, I think it an acceptable compromise to only solve point 2). That could be achieved by simply wrapping the native file input with the required structures. Depending on the outcome it might be needed to mark the component in documentation as `Labs`, `WIP` or something similar.

In either case need to maintain all accessibility and functionality of the native input.

adamkudrna:
image
image

Closes #219.

### DIFF
--- a/src/lib/components/ui/FileInputField/FileInputField.jsx
+++ b/src/lib/components/ui/FileInputField/FileInputField.jsx
@@ -1,0 +1,157 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import getRootValidationStateClassName from '../../../helpers/getRootValidationStateClassName';
+import { withProviderContext } from '../../../provider';
+import transferProps from '../../../utils/transferProps';
+import withForwardedRef from '../withForwardedRef';
+import styles from './FileInputField.scss';
+
+export const FileInputField = ({
+  changeHandler,
+  disabled,
+  forwardedRef,
+  fullWidth,
+  helpText,
+  id,
+  inFormLayout,
+  isLabelVisible,
+  label,
+  layout,
+  required,
+  validationState,
+  validationText,
+  ...restProps
+}) => (
+  <label
+    className={[
+      styles.root,
+      fullWidth ? styles.isRootFullWidth : '',
+      inFormLayout ? styles.isRootInFormLayout : '',
+      layout === 'horizontal' ? styles.rootLayoutHorizontal : styles.rootLayoutVertical,
+      disabled ? styles.isRootDisabled : '',
+      required ? styles.isRootRequired : '',
+      getRootValidationStateClassName(validationState, styles),
+    ].join(' ')}
+    htmlFor={id}
+    id={`${id}__label`}
+  >
+    <div
+      className={[
+        styles.label,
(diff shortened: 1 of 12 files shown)


src/lib/components/ui/FileInputField/FileInputField.jsx
```diff
@@ -0,0 +1,167 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import getRootValidationStateClassName from '../../../helpers/getRootValidationStateClassName';
+import { withProviderContext } from '../../../provider';
+import transferProps from '../../../utils/transferProps';
+import withForwardedRef from '../withForwardedRef';
+import styles from './FileInputField.scss';
+
+export const FileInputField = ({
+  changeHandler,
```
bedrich-schindler, line 10: I thought we want to swich to `on*` naming. I would suggest to use `onChange` instead. @mbohal?

src/lib/components/ui/FileInputField/README.mdx
```diff
@@ -0,0 +1,215 @@
...
+👉 Refer to the MDN reference for the full list of supported attributes of the
+[file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)
+type.
+
+## Invisible Label
+
+In some cases, it may be convenient to visually hide the field label. The label
+remains accessible to assistive technologies.
+
+While it may be acceptable for login screens with just a few fields or for other
```
bedrich-schindler, line 98: It does not seem to me right to talk about Login screen when we are documenting FileInput

src/lib/components/ui/FileInputField/__tests__/FileInputField.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+      changeHandler={spy}
+      id="test"
+      label="label"
+    />);
+
+    component
+      .find('input')
+      .simulate('change', {
+        preventDefault: () => {},
+        target: { value: 'test' },
```
bedrich-schindler, line 71: It is called `event.target.files` for `type=file`

bedrich-schindler (changes requested):

src/lib/components/ui/FileInputField/FileInputField.jsx
mbohal: This is to be addressed in  https://github.com/react-ui-org/react-ui/issues/218.

I think it better to leave it as is here and then changing everything at the same time. It is a little bit more work, but it will keep thinks consistent and clean.

mbohal (approved):

src/lib/components/ui/FileInputField/__tests__/FileInputField.test.jsx
mbohal: so it should read: `target: { files: ['test.jpg'] },`

PR #254, opened 2021-03-22
Allow running jest in a discrete npm script

mbohal:


package.json
```diff
@@ -105,7 +105,8 @@
     "docs:dev": "rimraf .docz && docz dev",
     "docs:serve": "rimraf .docz && npm run docs:build && docz serve",
     "lib:build": "concurrently \"webpack --mode development --config webpack.config.js\" \"webpack --mode production --config webpack.config.js\"",
-    "lib:test": "stylelint \"src/lib/**/*.scss\" --config stylelint.config.js && eslint src/lib --ext js,jsx && jest -w",
+    "lib:jest": "jest",
```
adamkudrna, line 108: Why is `-w` gone?

>  It may be useful to adjust this in resource limited environments like CIs but the defaults should be adequate for most use-cases.

Is that the reason?

https://jestjs.io/docs/cli#--maxworkersnumstring

adamkudrna (approved):

package.json
mbohal: Firstly `-w` IMHO should not be used. Abrivated flags are nice for interactive use, but scripts are much easier to read if long versions are used. So if something it should be `--maxWorkers=<num>`.

Also we do not use it anywhere. If we ever need to limit resources we can add it, but as far as I'm aware we do not do it. The defaults seem sensible enough to me. Am I wrong? If so I will return it.

bedrich-schindler (approved):

PR #255, opened 2021-03-22
[WIP] Use react Context for modifying form field component behaviour when r…

mbohal:
…endered within `FormLayout` component


mbohal:
I think this PR should not be a BC change.

It can cause warnings caused by setting now removed props, but nothing should break. Even these warnings will only happen when props were set and not used, for example when setting `layout` for a component within `FormLayout`.

src/lib/components/layout/FormLayout/FormLayout.jsx
```diff
@@ -56,20 +56,11 @@ export const FormLayout = (props) => {
...
-          return null;
-        }
-
-        return React.cloneElement(child, {
-          inFormLayout: true,
-          layout: fieldLayout,
-        });
-      })}
+      <FormLayoutContext.Provider
+        value={{ layout: fieldLayout }}
```
bedrich-schindler, line 60: This should be one liner as only one prop is present. Please check and correct another occurances.

src/lib/components/layout/FormLayout/FormLayout.jsx
```diff
@@ -1,8 +1,8 @@
-import flattenChildren from 'react-keyed-flatten-children';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { withProviderContext } from '../../../provider';
 import styles from './FormLayout.scss';
```
bedrich-schindler, line 4: It is small thing. However, you sometimes place `FormLayoutContext` before `styles`, sometimes the other way. Can you make it same for all occurances?

src/lib/components/layout/FormLayout/FormLayoutProvider.jsx
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import FormLayoutContext from './FormLayoutContext';
+
+const FormLayoutProvider = ({
+  children,
+  layout,
+}) => (
+  <FormLayoutContext.Provider
+    value={{ layout }}
```
bedrich-schindler, line 10: Same as above. One liner please.

src/lib/components/layout/FormLayout/FormLayoutCustomField.jsx
```diff
@@ -112,10 +115,6 @@ FormLayoutCustomField.propTypes = {
    * Optional ID of labeled field to keep accessibility features. Only available if `label` is set.
    */
   labelForId: PropTypes.string,
-  /**
-   * Layout of the field, controlled by parent FormLayout.
-   */
-  layout: PropTypes.oneOf(['horizontal', 'vertical']),
```
bedrich-schindler, line 118: We still want to have option to override it locally. For example, I might want to just have one totally separate field with different layout and It is non-sense to use FormLayoutProvider in this case. It applies for all following occurrences.

src/docs/customize/theming/forms.mdx
```diff
@@ -11,7 +11,7 @@ import { Playground } from 'docz'
 import { CheckboxField } from '../../../lib/components/ui/CheckboxField/CheckboxField'
 import { Radio } from '../../../lib/components/ui/Radio/Radio'
 import { SelectField } from '../../../lib/components/ui/SelectField/SelectField'
-import { TextField } from '../../../lib/components/ui/TextField/TextField'
+import TextField from '../../../lib/components/ui/TextField'
```
bedrich-schindler, line 14: I've met the same problem. We should use one or other variant in docs, but mixing them is not good idea. So we need to agree which variant is the correct.

tests/providerTests/inFormLayoutProviderTest.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <FormLayoutProvider {...testedProps}>
+        {Component}
+      </FormLayoutProvider>
+    ));
+
+    assert(dom.container.firstChild);
+  });
+
+  it('renders without FormLayout', () => {
+    const dom = render((
```
bedrich-schindler, line 32: Again. Why multi line and multi brackets? This could be `render(Component)` and it is fine.

src/lib/components/layout/FormLayout/FormLayoutCustomField.jsx
bedrich-schindler: For info. This comment applies only to `layout`, it does not apply for `inFormLayout`.

bedrich-schindler (changes requested):

src/lib/components/layout/FormLayout/FormLayoutCustomField.jsx
mbohal: Current implementation does not support it. And I did not want to introduce BC changes in this PR.

I don't have an opinion about this in general, but since it would cause a change of behaviour I would prefer to not do it in this PR.

@adamkudrna Is it even technically possible?

src/docs/customize/theming/forms.mdx
mbohal: I think that outside the component root directory we should always import via `index.js` as that is the public API of the component.

Are there any reasons why importing directly from the component file is more practical?

src/lib/components/layout/FormLayout/FormLayout.jsx
mbohal: Decided during a phone call with @bedrich-schindler that we do not want to deal with these things on PR. Either there should be an eslint for it or it is at the discretion of the developer.

tests/providerTests/inFormLayoutProviderTest.jsx
mbohal: Decided during a phone call with @bedrich-schindler that we do not want to deal with these things on PR. Either there should be an eslint for it or it is at the discretion of the developer.

src/lib/components/layout/FormLayout/FormLayoutProvider.jsx
mbohal: Decided during a phone call with @bedrich-schindler that we do not want to deal with these things on PR. Either there should be an eslint for it or it is at the discretion of the developer.

src/lib/components/layout/FormLayout/FormLayout.jsx
mbohal: I think ordering works like this:

* Imports are ordered from farthest away to the closes ones. (checked by eslint)
* Imports of the same distance are ordered alphabetically. (not checked by eslint)

It is correct here. If I placed it wrong somewhere I will fix it, but IMHO here is nothing to do. Are there any places in this PR where it is wrong?

src/lib/components/layout/FormLayout/FormLayoutCustomField.jsx
bedrich-schindler: I don't get it. You are the one who is creating BC by removing option to locally overwrite `layout` ... or am I wrong?

src/docs/customize/theming/forms.mdx
bedrich-schindler: It is up to @adamkudrna. In my opinion, we should use index.js exports. The only part where raw/named export is necessary, it proptypes doc rendering.

src/lib/components/layout/FormLayout/FormLayoutCustomField.jsx
mbohal: While before this PR it was possible to set `layout` for a component within `FormLayout` it did nothing - it was ignored.

adamkudrna (approved):
A few small details but generally OK. Good job!

src/docs/customize/theming/forms.mdx
mbohal: I see it is more complicated. Either way this change has no place in this PR. I'm reverting it.

src/lib/components/layout/FormLayout/FormLayout.jsx
mbohal: I also do it by import name (leftmost name) same as @adamkudrna 

I fixed this occurance.

src/lib/components/layout/FormLayout/README.mdx
```diff
@@ -20,7 +20,7 @@ import { FileInputField } from '../../ui/FileInputField/FileInputField'
 import { Radio } from '../../ui/Radio/Radio'
 import { SelectField } from '../../ui/SelectField/SelectField'
 import { TextArea } from '../../ui/TextArea/TextArea'
-import { TextField } from '../../ui/TextField/TextField'
+import TextField from '../../ui/TextField'
```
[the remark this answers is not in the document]
mbohal: fixed

src/lib/components/layout/FormLayout/FormLayout.jsx
```diff
@@ -97,7 +88,10 @@ FormLayout.propTypes = {
    * * `TextField`
    * * `Toggle`
    */
-  children: PropTypes.element,
+  children: PropTypes.oneOfType([
+    PropTypes.element,
+    PropTypes.arrayOf(PropTypes.element),
+  ]),
```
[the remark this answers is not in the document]
mbohal: #313 builds on top of this. 

My changes fixes previously erroneous value.

PR #258, opened 2021-03-26
Drop automatic spacing of standalone form fields (#94)

### LINKED ISSUE #94: Remove automatic spacing from form fields
Related to #41.

Providing form fields with no automatic margins should lead to a better expected behaviour. By far, automatic margins don't work as good as `Toolbar` or `FormLayout`.

Resets of auto margins have to be removed from the afore-mentioned layout components.

adamkudrna:
A layout like `FormLayout` or `Toolbar` should be used to achieve proper spacing between fields.

Closes #94.


mbohal (approved):
I'm not qualified to review this :(

The only thing that I can think of is that I do not see any changes to the documentation in this PR. Is that OK?

adamkudrna:
> The only thing that I can think of is that I do not see any changes to the documentation in this PR. Is that OK?

Thanks for pointing it out. I don't think there is anything to change. I quickly flipped through the docs and I believe I hadn't mentioned it anywhere because we had known that auto spacing would be removed anyway.

PR #259, opened 2021-03-31
Remove bottom margin from last-child elements (#180)

### LINKED ISSUE #180: Remove margin of the last element
Presently there is a margin after a last element of type:

* `<p>`
* `<ol>`
* `<ul>`
* `<li>`

This leads to extra unwanted space. It can be resolved manually, but that is repetitive code. There should be no space by default.

adamkudrna:
Separate shared bottom spacings into categories with configurable values.

New theming options:

- `--rui-spacing-bottom-default`
- `--rui-spacing-bottom-headings`
- `--rui-spacing-bottom-layouts`

image

Closes #180.

### DIFF
--- a/src/docs/foundation/spacing.mdx
+++ b/src/docs/foundation/spacing.mdx
@@ -24,3 +24,45 @@ help you keep your UI consistent.
 \* For the sake of brevity, usage in HTML only illustrates the top margin
 property. See [Spacing helpers](/css-helpers/spacing) for the full list of
 options.
+
+## Shared Spacings
+
+Commonly reused spacings. Bottom spacing is applied unless the element in
+question is a last child of its parent.
+
+| Category | Usage in CSS                    | Usage in SCSS              | Default value     | Shared by              |
+|----------|---------------------------------|----------------------------|-------------------|------------------------|
+| default  | `--rui-spacing-bottom-default`  | `spacing.bottom()`         | `--rui-spacing-5` | Paragraphs, lists etc. |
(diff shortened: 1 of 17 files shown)


bedrich-schindler (changes requested):
It seems logic to me, but when I tried to change those values using Chrome DevTools, only spacing of headings changed. Spacing of elements did not - for example for Toolbar, there is no visual change when I change all values to `--rui-spacing-6` or `1`:

image

Can you check it?

adamkudrna:
@bedrich-schindler I think you did the test in docs context, correct? There is a default spacing around all direct descendants (`> *`) provided by live playground which is probably what you encountered. This spacing is useful most of the times and when you switch it off in dev tools you'll see it works:

Snímek obrazovky 2021-04-08 v 17 39 43

adamkudrna:
@bedrich-schindler … or just head to any configurable component example like [this one](http://localhost:3000/components/layout/toolbar#alignment). Multiple sibling layouts are used there and I just added a few more.

bedrich-schindler:
Ok, if it does not work in playground, it is ok then. We might write some note that some of those modification do not work in playground, so user knows that it is not a bug. But it is for another discussion. Maybe new issue ..

adamkudrna:
> Ok, if it does not work in playground, it is ok then. We might write some note that some of those modification do not work in playground, so user knows that it is not a bug. But it is for another discussion. Maybe new issue ..

You're right, I appended a note to Foundation > Spacing:

image

PR #260, opened 2021-04-08
Use unitless line height to take advantage of inheritance

adamkudrna:
⚠️ Breaking change, you may need to multiply your calculations by `1rem`.


bedrich-schindler (approved):
Hacker!

PR #262, opened 2021-04-16
Update dependencies (#261)

bedrich-schindler:


.eslintrc
```diff
@@ -10,6 +10,7 @@
   },
   "rules": {
     "func-names": "off",
+    "react/destructuring-assignment": "off",
```
bedrich-schindler, line 13: I would say that this is the most stupid ESlint rule I've ever seen. https://eslint.org/docs/rules/prefer-destructuring is OK, but this? Its requirements do not make sense to me so I've turned it off. Also it detects only some lines and some do not.

package.json
```diff
@@ -37,58 +37,58 @@
...
-    "eslint-plugin-mdx": "^1.8.2",
-    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-markdown": "^2.0.1",
+    "eslint-plugin-mdx": "^1.12.0",
+    "eslint-plugin-react": "^7.23.2",
     "eslint-plugin-react-hooks": "^4.2.0",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.6.3",
-    "markdownlint-cli2": "~0.0.13",
+    "markdownlint-cli2": "^0.1.3",
```
adamkudrna, line 78:
```suggestion
    "markdownlint-cli2": "~0.1.3",
```

adamkudrna (approved):
Just a note: please keep in mind that PR title is used in changelog so it should provide some kind of useful information 🙂.

Issue #156, opened 2020-09-29
Improve propTypes validation


adamkudrna:
We would like to validate that:

- mutually exclusive props aren't set at the same time
- parent component of a multi-level component only receives children of certain type (eg. `Toolbar` only accepts `ToolbarItem` and `ToolbarGroup` and nothing else)

Make use of https://github.com/airbnb/prop-types ?

Relates to https://github.com/facebook/prop-types/issues/327#issuecomment-699040889

PR #275, opened 2021-05-12
Allow passing `<ScrollView />` as a prop to modal to facilitate easie…

mbohal:
…r configuration


src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -247,14 +238,7 @@ Modal.defaultProps = {
...
   position: 'center',
-  scrollMode: 'body',
-  scrollViewEndShadowStyle: {
-    background: 'radial-gradient(farthest-side at center bottom, rgba(0, 0, 0, 0.16) 0%, rgba(0, 0, 0, 0.06) 40%, rgba(0, 0, 0, 0.02) 85%, rgba(0, 0, 0, 0) 100%)',
-  },
-  scrollViewShadowSize: '16px',
-  scrollViewStartShadowStyle: {
-    background: 'radial-gradient(farthest-side at center top, rgba(0, 0, 0, 0.15) 0%, rgba(0, 0, 0, 0.05) 60%, rgba(0, 0, 0, 0.02) 85%, rgba(0, 0, 0, 0) 100%)',
-  },
+  scroller: (<ScrollView />),
```
bedrich-schindler, line 241: `<ScrollView />` should contain `scrollViewEndShadowStyle`, `scrollViewShadowSize, `scrollViewStartShadowStyle` defaults from removed `defaultProps`.

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -297,29 +281,10 @@ Modal.propTypes = {
...
-  /**
-   * Custom CSS for the start scrolling shadow.
+   * The React element to be used as a wrapper of the content providing body scrolling functionality.
+   * If set to `null` the entire modal including header and footer will be scrolled.
    */
-  scrollViewStartShadowStyle: PropTypes.shape({
-    background: PropTypes.string,
-    boxShadow: PropTypes.string,
-  }),
+  scroller: PropTypes.node,
```
bedrich-schindler, line 287: I would suggest to try something like https://www.npmjs.com/package/airbnb-prop-types and validate that only `ScrollView` can be passed to `scroller`.

bedrich-schindler, line 287: If we all to pass only `ScrollView` into `scroller`, I would suggest to use `scrollView` prop instead of `scroller`

bedrich-schindler (changes requested):

src/lib/components/ui/Modal/Modal.jsx
mbohal: done

src/lib/components/ui/Modal/Modal.jsx
mbohal: done

mbohal: This is covered by #156  - I added this specific ase to the description there to make sure it does not get forgotten.

src/lib/components/ui/Modal/README.mdx
```diff
@@ -283,7 +283,7 @@ component.
               }]}
               autoFocus={false}
               closeHandler={() => setModalOpen(false)}
-              scrollMode={modalScrollMode}
+              scrollView={modalScrollMode === 'body' ? undefined : null}
```
adamkudrna, line 286: I just don't get it. According to `propTypes`, this can only be a `node` (or `undefined`). What I see here is `undefined` (OK, omit the attribute and use the default value) and `null` (…?). Can somebody please explain to me what's going on here? 👀

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -297,29 +285,11 @@ Modal.propTypes = {
...
-   * Custom CSS for the start scrolling shadow.
+   * The `ScrollView` component to be used as a wrapper of the content to provide body scrolling functionality.
+   * If provided only the modal body will be scrollable.
+   * If set to `null` the entire modal including header and footer will be scrollable.
    */
-  scrollViewStartShadowStyle: PropTypes.shape({
-    background: PropTypes.string,
-    boxShadow: PropTypes.string,
-  }),
+  scrollView: PropTypes.node,
```
adamkudrna, line 292: We certainly want just a single `ScrollView`, if any. Could we narrow it down to a single element?

```suggestion
  scrollView: PropTypes.element,
```

adamkudrna (changes requested):
The idea seems fine to me and it works, cool. I have just a couple of things to clarify.

src/lib/components/ui/Modal/README.mdx
adamkudrna: For the sake of clarity, I'd rename the effect in this example to `modalScrollView` and make it boolean (also see my confusion below). The previously intended extensibility of `modalScrollMode` is no longer valid with the API change here in this PR.

mbohal: > this can only be a node (or undefined)

The `ScrollView` prop is not required. It can therefore be `node`, `null` or `undefined`. They mean the following things:
* `node` use the specified instance of `ScrollView`
* `undefined` do not use any `ScrollView`, scroll the entire modal including header and footer
* `null` use the instance of `ScrollView` defined in `defaultProps`

mbohal: > I'd rename the effect in this example to modalScrollView

done

src/lib/components/ui/Modal/Modal.jsx
mbohal: done

adamkudrna (approved):

PR #276, opened 2021-05-24
Improve prop types definitions to validate children types (#156)

bedrich-schindler:


mbohal (approved):

mbohal:
This PR tests are crashing on exceeding [`MAX_DEVELOPMENT_OUTPUT_SIZE`](https://github.com/react-ui-org/react-ui/blob/eafb63dcdab4b582661a974e2695473156009995/webpack.config.js#L5).

Do we even need to check max development size?

adamkudrna (changes requested):
- Please take a look at React fragments as discussed today.
- Please check author email address in your commit.

bedrich-schindler:
> This PR tests are crashing on exceeding [`MAX_DEVELOPMENT_OUTPUT_SIZE`](https://github.com/react-ui-org/react-ui/blob/eafb63dcdab4b582661a974e2695473156009995/webpack.config.js#L5).
> 
> Do we even need to check max development size?

I think, we need. We want to prevent situation when we accidentally increase bundle size by using a library we haven't check before using it. It also prevents situations when we e.g. use wrong imports that leads to increased size.

mbohal:
@bedrich-schindler I understand that we need max size check for production builds. However, do we need it for dev builds?

bedrich-schindler:
@adamkudrna I have not found a way to make it compatible with React.Fragment. So If React.Fragment is passed as child, prop types validation fails. Neither have I found a word about in its "docs" nor found a way to make it working. :/ 

Last idea is to create an issue in its repo :/

mbohal:
> So If React.Fragment is passed as child, prop types validation fails

That can lead to a lot of false positives. In certain cases it can be practical to create the child elements elsewhere and in such cases the `React.Fragment` is essential.

Example of usage dependent on fragments:

```javascript
    const listItems = () => (
      <>
        <ListItem>List item</ListItem>
        <ListItem>List item that is longer</ListItem>
      </>
    )
    return (
      <List align="end" autoWidth>
        {listItems()}
      </List>
    )
```

So in reality we need not only the correct component to pass the validation, but also  the correct component wrapped in `React.Fragment`. Of course the `React.Fragment`s can be nested…

I'm unsure if we should even merge this without support for `React.Fragment`. I'm afraid it might make usage too painful.

bedrich-schindler:
Closing this PR as new version of implementation has been made from scratch.

Issue #272, opened 2021-05-03
Unify API of components which have UI states variations


adamkudrna:
`type="success"` vs. `variant="success"`

Somewhere it's `type` (eg. `Alert`, `Card`), somewhere `variant` (eg. `Button` – because of conflict with native `type` attribute).

mbohal:
We need to unify:
* `propType` name
* `propType` option list
  * `success`
  * `warning`
  * `danger`
  * `help`
  * `info`
  * `note`

Issue #86, opened 2020-05-26
Enable generating documentation of advanced `propTypes` syntax


bedrich-schindler:
**Added by** @adamkudrna: This includes these useful ways of abstraction:

- `propTypes` assigned as value of a constant in the same file
- `propTypes` defined in another file

---

Based on issue: https://github.com/doczjs/docz/issues/1431

We need to create documentation component that will allow us to render documentation table with all props and its description. This component should be similar to default docz PropType component, but need to be able to renders props that are composed from another proptype definitions.

I would recommend to move this component to separate repository later, because it seems that there are more people with same problem like we have.

bedrich-schindler:
@adamkudrna @mbohal I spent more that one hour reading and trying to somehow read and parse JS comment blocks, but without success. There are few discussions on StackOverflow telling that there is probably no way how to do it in the browser because comments are not available in the DOM.

The only way is to do it is to read it and parse it in Node environment, create metadata and render it from it. But it is extremelly hard and this is what docz do all together with other tools.

I think that the only option is to push docz team to fix it or try to create pull request that fixes it.

To sum up, there is no easy way out except writing it manually ...

bedrich-schindler:
I think that this should be closed as this is no more valid and we are getting rid of TypeScript and prop types in #394 

What do you think @mbohal @adamkudrna?

PR #278, opened 2021-05-28
Unify API of components with color variants (#272)

adamkudrna:
- All components that support color variants now use the same palette: primary, secondary, success, warning, danger (replaces error), help, info, note, light, and dark.
- Color variants are now invoked by `variant` prop in all supporting components (replaces `type` prop in some components).
- Rename `error` color to `danger`.
- Use unified color variants palette also with text utilities (with addition of muted option).
- Remove theming of Card variants since reading definitions of variant colors directly works just fine.
- Improve default color palette with better grading.
- Separate SCSS code from `theme.scss`.

obrazek

obrazek

obrazek


bedrich-schindler:
I have only one small request. Is it possible to extract supported variants (`['primary', 'secondary', 'success', 'warning', 'danger', 'help', 'info', 'note', 'light', 'dark'],`) to some constant, so it is more straightforward and easy to maintain.

bedrich-schindler (approved):

src/lib/components/ui/Card/Card.jsx
```diff
@@ -1,87 +1,49 @@
...
-  variant: 'flat',
+  variant: null,
 };
 
 Card.propTypes = {
   /**
-   * Slot for individual Card elements that build up the inner layout: CardBody, CardFooter, or
+   * Paint border around the card.
+   */
+  bordered: PropTypes.bool,
```
mbohal, line 44: Since `variant` only has visible effect when `bordered=true` I think we can omit the `bordered` prop as it can lead to unexpected result when <Card bordered={false} variant="success" /> will not show any green anywhere.

Having `variant` set to non `null` value would draw the border. Setting `variant` to `null` would draw no border.

Of course this can only work if `variant` is only used to set border color. But since `Card` is such a simple component I think we can make this assumption.

mbohal (changes requested):

src/docs/_components/Placeholder/Placeholder.jsx
```diff
@@ -5,12 +5,14 @@ import styles from './Placeholder.scss';
 export const Placeholder = ({
```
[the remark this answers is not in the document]
mbohal: On second thought, I don't think the name `priority` is good to indicate `filled` vs `outline`.

src/lib/components/ui/Alert/Alert.jsx
```diff
@@ -1,84 +1,56 @@
...
+      </button>
+    )}
+  </div>
+);
 
 Alert.defaultProps = {
   closeHandler: null,
   icon: null,
   id: undefined,
-  type: 'note',
```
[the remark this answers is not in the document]
bedrich-schindler: I am not sure if it is possible to have defaults in those cases. If it is empty, it should have neutral look and I also do not think we have this kind of variant and whether we want default variant of those components because they might not have real use-case. And creating default variant just to have "default appearance" does not sound good to me.

Please @adamkudrna let us know your opinion. I believe Adam has wider view into UI libraries so I will probably follow his recommendation in this case.

src/docs/_components/Placeholder/Placeholder.jsx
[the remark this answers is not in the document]
bedrich-schindler: I agree we should avoid using props with same name with for different purpose. They might be cases when we are unable to use unique name but I hope this is not the case.

However I am not sure that we want to use full set of `priority` as it is for `variant`. `variant` is simple coloring of component, `priority` is another level of visual difference. In my opinion. And therefore it is not necessary to use all of them. But again, I believe in Adam's decision .

src/lib/components/ui/Card/Card.jsx
bedrich-schindler: I am not sure if there is not something deeper in using those props separately. @adamkudrna?

adamkudrna:
> I have only one small request. Is it possible to extract supported variants (`['primary', 'secondary', 'success', 'warning', 'danger', 'help', 'info', 'note', 'light', 'dark'],`) to some constant, so it is more straightforward and easy to maintain.

@bedrich-schindler This is still blocked by #86 (should the constant be placed in a separate file).

src/lib/components/ui/Badge/Badge.jsx
```diff
@@ -73,7 +46,9 @@ Badge.propTypes = {
   /**
    * Color variant to clarify importance and meaning of the badge.
    */
-  type: PropTypes.oneOf(['error', 'help', 'info', 'note', 'success', 'warning', 'light']),
+  variant: PropTypes.oneOf(
+    ['primary', 'secondary', 'success', 'warning', 'danger', 'help', 'info', 'note', 'light', 'dark'],
```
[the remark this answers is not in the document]
adamkudrna: This is unfortunately still blocked by #86.

src/lib/components/ui/Alert/Alert.jsx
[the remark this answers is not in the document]
adamkudrna: I'm for 2 as @mbohal suggests. A default, unstyled appearance (like with Bootstrap) doesn't make sense to me because I don't see any use case for that, as @bedrich-schindler also says. Default appearance may be (and it often is) hidden in implementation, but from the design point of view, such a visual variant makes no sense and should not be suggested or made available to use. Let's go with sensible defaults per component.

src/docs/_components/Placeholder/Placeholder.jsx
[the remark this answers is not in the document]
adamkudrna: > Namely we should rename the variant prop in SelectField TextField and TextArea to priority. This would make it compatible with Button, ButtonGroup and Badge.

Nice touch. Problem is that `variant` of a form field is not the same as `priority` of other components (`Badge`, `Button`). While priority of `Badge`s and `Button`s actually tells a lot of component's _visual priority_ (making it more or less pop out from its context), form field variants really are just _visual alternatives,_ all equal one to each other from visual priority perspective. That brings us to the root of this confusion. While the _property_ means something different, the available _values_ are the same: `filled` and `outline`:

obrazek

To avoid this confusion, we may want to rename the values like this:

- (on the left) `outline` to `bordered` (maybe something better? See below),
- (on the right) `filled` to `bottom-line`.

I wouldn't go for `border-all` and `border-bottom` (or similar) because the difference is in available styling options and in the focus effect rather than just the number of bordered sides.

Also, we may object it may be confusing to call the variant `bordered` as the border can be removed with custom styling (theme). But I haven't come across any better name than `default` which is too vague.

src/lib/components/ui/Card/Card.jsx
adamkudrna: > Since variant only has visible effect when bordered=true

It's no longer the case. I decided to draw border of a `variant` even when `bordered` is not set (and setting it has no effect in such case, the border is already there).

Also, omitting the `bordered` prop is not possible—unless we change the design of _our project_ at the same time.

Current situation in the _project_ is:

- We need flat, border-less cards for large surfaces like the main content area.
- We use bordered `Card` for card listings. The color of the border is neither `light` nor `dark` (see variants), it's just a shade of gray from our gray palette.

We could only get rid of the `bordered` prop if we changed the design to:

1. only use shadow (using the `raised` prop) for card listings, or
2. use `dark` border variant when a border is needed.

From the design point of view, 1) requires heavier shadows to separate white card from white background and 2) would make the UI too cluttered with borders.

So I think we need to keep the `bordered` option as is.

src/lib/components/ui/Alert/Alert.jsx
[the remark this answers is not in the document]
adamkudrna: I'm proposing these `variant` defaults:

- `Alert`: `note` as the lowest (and neutral) visual priority available (`light` and `dark` are meant rather for special purposes).
- `Badge`: `note`, see above.
- `Button`: `primary` because there should be primary buttons (we think).
- `Card`: `null` which makes `variant` optional (correct from the design point of view).

src/lib/components/ui/Badge/Badge.jsx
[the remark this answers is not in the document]
mbohal: I see :(

adamkudrna:
Our findings with @mbohal:

- extract `Paper`/`Surface` from `Card` and simplify API
- make `variant` required for `Card` and default to `light`, merge `bordered` into default appearance
- consider better name for `variant` which seems to be too generic to mean a color

src/lib/components/ui/Card/Card.jsx
adamkudrna: We decided to get rid of the `bordered` prop and make the border mandatory, yet customizable through theming options.

src/docs/_components/Placeholder/Placeholder.jsx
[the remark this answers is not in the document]
adamkudrna: Possible option: make the first variant primary (default) and the other optional (`bottomLine`, boolean). Don't close the door to floating labels that may come in future.

adamkudrna:
⚠️ **REBASED** ⚠️

Latest update: rename `variant` to `color` across components.

image

src/docs/_components/Placeholder/Placeholder.jsx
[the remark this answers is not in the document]
adamkudrna: Having changed the `variant` prop to `color`, I think this is less of a problem now. It may be still confusing that `filled` and `outline` are values used for both `priority` and `variant` props, but I don't see it as a major problem.

src/lib/components/ui/Button/Button.jsx
```diff
@@ -178,6 +127,10 @@ Button.propTypes = {
    * Function to call when the button is clicked.
    */
   clickHandler: PropTypes.func,
+  /**
+   * [Color variant](/foundation/colors#component-colors) to clarify importance and meaning of the button.
+   */
+  color: PropTypes.oneOf(['primary', 'secondary', 'success', 'warning', 'danger', 'help', 'info', 'note', 'light', 'dark']),
```
bedrich-schindler, line 133: It would be better to store this array into constant that will be shared across all components. (It is applicable for all components)

src/lib/components/ui/Alert/Alert.jsx
```diff
@@ -91,6 +64,12 @@ Alert.propTypes = {
    * hidden.
    */
   closeHandler: PropTypes.func,
+  /**
+   * [Color variant](/foundation/colors#component-colors) to clarify importance and meaning of the alert.
+   */
+  color: PropTypes.oneOf(
```
bedrich-schindler, line 70: Check Foundation -> Icons page. You've forgotten to change type to color there.

src/lib/components/ui/Badge/README.mdx
```diff
@@ -43,58 +43,67 @@ lowest:
...
-  <Badge label="365" priority="outline" type="light" />
+  <Badge priority="outline" label="1" color="primary" />
+  <Badge priority="outline" label="2" color="secondary" />
+  <Badge priority="outline" label="3" color="success" />
+  <Badge priority="outline" label="4" color="warning" />
+  <Badge priority="outline" label="5" color="danger" />
+  <Badge priority="outline" label="10" color="help" />
+  <Badge priority="outline" label="34" color="info" />
+  <Badge priority="outline" label="99+" />
+  <Badge priority="outline" label="365" color="light" />
```
bedrich-schindler, line 92: Light color is not visible in the docs.

src/lib/components/ui/Button/README.mdx
```diff
@@ -191,7 +198,7 @@ information or to draw user's attention.
 <Playground>
```
bedrich-schindler, line 198: Light color is not visible in the docs.

bedrich-schindler (approved):

src/lib/components/ui/Button/Button.jsx
adamkudrna: > This is unfortunately still blocked by #86.

src/lib/components/ui/Modal/Modal.jsx
```diff
@@ -252,11 +252,11 @@ Modal.propTypes = {
    */
   actions: PropTypes.arrayOf(PropTypes.shape({
     clickHandler: PropTypes.func.isRequired,
+    color: PropTypes.string,
```
mbohal, line 255: This should list all the colors as enum.

mbohal (approved):

src/lib/components/ui/Button/Button.jsx
adamkudrna: 🤷🏻‍♂️

PR #279, opened 2021-06-01
Update dependencies

bedrich-schindler:


mbohal (approved):

src/lib/components/ui/Button/_tools.scss
```diff
@@ -89,7 +90,7 @@
   width: var(--rui-local-width, auto);
   height: var(--rui-local-height, auto);
   padding: var(--rui-local-padding);
-  font: theme.$font-weight var(--rui-local-font-size)/settings.$line-height settings.$font-family;
+  font: theme.$font-weight math.div(var(--rui-local-font-size), settings.$line-height) settings.$font-family;
```
adamkudrna, line 93: Nono, this is not division but CSS syntax of the `font` shorthand: `<font size>/<line height>`. This needs to be changed to `list.slash()`:

```suggestion
  font: theme.$font-weight list.slash(var(--rui-local-font-size), settings.$line-height) settings.$font-family;
```

https://sass-lang.com/documentation/values/lists#slash-separated-lists
https://sass-lang.com/documentation/modules/list#slash

adamkudrna (changes requested):
Just a small fix and it's done.

src/lib/components/ui/Button/_tools.scss
bedrich-schindler: Should be completed

adamkudrna (approved):

PR #280, opened 2021-06-03
Drop support for automatically setting `TextField` size attribute based on the value of the `max` attribute for when `type=number`

mbohal:
…ed on the value of the `max` for when `type=number`

This automagical functionality does not work when:
* value is not an integer (the decimal places were not taken into account)
* value is negative (the leading`-` was not taken into account)
* value is a specific case of scientific notation (`1e1 = 10`)

I suppose we could find a way to calculate the correct size based on the `max`, `min` and `step` attributes when a standard number is entered. However, I think it is impossible to do for the scientific notation format. So I think dropping auto size setting feature is the only way to fix this.


bedrich-schindler (approved):

adamkudrna (approved):
OK, I agree, I hadn't fully considered all use cases when designing this feature. So:

1. Please fix the PR title to work as changelog item, see https://react-ui.io/contribute/guidelines#git-workflow.
2. Let's drop the feature by merging this PR.
3. I'll file an issue to extend the `size` attribute also for number input types. I believe custom size of number input is even more useful than with other input types.

src/lib/components/ui/TextField/README.mdx
```diff
@@ -206,11 +206,9 @@ normalize rendering across browsers.
...
 the `maxlength` attribute to achieve that effect.
 
-👉 For compatibility with the HTML specification, the `inputSize` option (which
+👉 Note that according to the HTML specification, the `inputSize` option (which
 is translated as `size` attribute of the input element) is
 [not available for the `number` input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Controlling_input_size).
-However, you can use the `max` attribute to determine the desired width of the
-number input instead.
 
 <Playground>
```
adamkudrna, line 213: Please just remove the last two `TextField` examples here—we don't want to demonstrate something is *not* working (I believe).
obrazek

adamkudrna (approved):
I added request for a tiny fix in demo.

PR #282, opened 2021-06-09
Make `id` optional in all components

mbohal:
This is now possible after https://github.com/visionappscz/eslint-config-visionapps/pull/7 was merged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5645,9 +5645,9 @@
       }
     },
     "@visionappscz/eslint-config-visionapps": {
-      "version": "1.4.0",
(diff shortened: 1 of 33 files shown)


adamkudrna (approved):
- I'd change PR title from `Make HTML attribute id optional in all elements` to `Make `id` optional in all components`.
- I'm not sure about dumping the `htmlFor` attribute—why not use the `id` when it's there… But it's a subtle detail I'm ready to leave as is.

mbohal:
@adamkudrna 
> 'm not sure about dumping the htmlFor attribute—why not use the id when it's there

Good point, I have fixed that.

Also I have removed the now unnecessary `id` attributes from the `README.mdx` files. I think the less cluttered the examples are, the better.

adamkudrna (approved):

PR #289, opened 2021-06-28
Introduce `Text` component to control wrapping of text content (#286)

### LINKED ISSUE #286: Control wrapping of content within Toolbar items
Possibly add `wrap` for `ToolbarItems` and `wrapItems` for `ToolbarGroup` and `Toolbar` to allow text wrapping (which is now turned off).

adamkudrna:
obrazek

---

obrazek

Closes #286.

### DIFF
--- a/src/lib/components/ui/Text/Text.jsx
+++ b/src/lib/components/ui/Text/Text.jsx
@@ -1,0 +1,73 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { withProviderContext } from '../../../provider';
+import getRootHyphensClassName from './helpers/getRootHyphensClassName';
+import getRootWordWrappingClassName from './helpers/getRootWordWrappingClassName';
+import styles from './Text.scss';
+
+export const Text = ({
+  blockLevel,
+  children,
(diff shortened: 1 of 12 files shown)


src/lib/components/ui/Text/Text.jsx
```diff
@@ -0,0 +1,73 @@
...
+  wordWrapping: 'normal',
+};
+
+Text.propTypes = {
+  /**
+   * If true, the root HTML element renders as `<div>` instead of `<span>`.
+   */
+  blockLevel: PropTypes.bool,
+  /**
+   * Text content to be sanitized. Can contain HTML.
```
bedrich-schindler, line 50: I think it should be `Can contains HTML`

bedrich-schindler (approved):

src/lib/components/ui/Text/Text.jsx
adamkudrna: Only if we were talking about [cans](https://slovnik.seznam.cz/preklad/cesky_anglicky/plechovka?strict=true) 🙂.

PR #291, opened 2021-07-30
Introduce `Tabs` component (#288)

### LINKED ISSUE #288: Add `Tabs` component

adamkudrna:
obrazek
obrazek

Closes #288.


src/lib/components/ui/Tabs/TabsItem.jsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  isActive: PropTypes.bool,
+  /**
+   * Item's label.
+   */
+  label: PropTypes.string.isRequired,
+  /**
+   * Item's link. Can be a URI or an object with `href` and `onClick` function.
+   */
+  link: PropTypes.oneOfType([
```
bedrich-schindler, line 74: Is object definition necessary here? It would be more simple to have:

```js
link: PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.func,
])
```

I do not see benefit in having mixed definition for a such simple purpose. I don't even know why to have `href` and `onClick` defined at the same time as `href` is never used. What more, this definition is not correct as it allows to have empty definition object resulting in exception even if prop type check is passed.

src/lib/components/ui/Tabs/__tests__/Tabs.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+      <Tabs />,
+    );
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders correctly with children', () => {
+    const tree = shallow(
+      <Tabs id="my-tabs">
+        <li>item 1</li>
```
bedrich-schindler, line 17: We expect to have strict prop type definition later. We want to have `TabItem` here as `li` should never be passed to Tabs.

src/lib/components/ui/Tabs/README.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+<Props table of={TabsItem} />
+
+## Theming
+
+### Tabs Theming
+
+| Custom Property                                      | Description                                                  |
+|------------------------------------------------------|--------------------------------------------------------------|
+| `--rui-Tabs__border-width`                           | Width of decorative bottom border                            |
```
bedrich-schindler, line 208: You talk here about bottom border but property name is without "bottom" in it. Is it correct? Seems like a bug.

bedrich-schindler (changes requested):

src/lib/components/ui/Tabs/TabsItem.jsx
bedrich-schindler, line 74: `link` should be replaced by `href` which must be mandatory string. Then there should be `onClick` function which is optional and could be use to override `href` functionality. It could be then replaced globally by context. Approved by @mbohal

bedrich-schindler (changes requested):

src/lib/components/ui/Tabs/TabsItem.jsx
adamkudrna: Fixed.

src/lib/components/ui/Tabs/__tests__/Tabs.test.jsx
adamkudrna: Fixed.

bedrich-schindler (approved):

mbohal (approved):

PR #293, opened 2021-08-04
Remove flex layout from `Paper` as it prevented margin collapsing of its content

adamkudrna:


src/lib/components/ui/Card/Card.scss
```diff
@@ -7,7 +9,7 @@
 
   display: flex;
   flex-direction: column;
-  width: 100%;
+  min-width: 0; // 1.
```
adamkudrna, line 12: `width` is not necessary and `min-width` should improve behavior in flex and grid layouts.

bedrich-schindler (approved):

PR #301, opened 2021-08-30
Use react-testing-library for testing

mbohal:

PR #309, opened 2021-09-15
Document composition principles and CSS guidelines

adamkudrna:
New sections:

- Contribute > Composition
- Contribute > CSS Guidelines

A PR with refactored CSS to match the new guidelines will follow as soon as #301 is merged.

### DIFF
--- a/src/docs/contribute/css.mdx
+++ b/src/docs/contribute/css.mdx
@@ -1,0 +1,256 @@
+---
+name: CSS Guidelines
+route: /contribute/css
+menu: Contribute
+---
+
+# CSS Guidelines
+
+Basic rules to follow when authoring React UI CSS.
+
+## Coding Style
+
+React UI uses [Stylelint] to avoid errors in CSS and enforce unified coding
+style across all stylesheets.
+
+Configurations used:
+
+- [`stylelint-config-visionapps`][stylelint-config-visionapps] — a more strict
+  extension of popular community preset
+  [`stylelint-config-standard`][stylelint-config-standard].
+
+- [`stylelint-config-visionapps-order`][stylelint-config-visionapps-order] —
+  an opinionated list of CSS properties order.
+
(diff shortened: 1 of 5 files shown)


src/docs/contribute/css.mdx
```diff
@@ -0,0 +1,256 @@
...
+- **Classes that are automatically generated by Sass loops should be handled
+  with care.** With loops, it's easy to produce a lot of CSS and negatively
+  impact performance.
+
+## Postprocessing with PostCSS
+
+All styles are automatically prefixed by [Autoprefixer] plugin for [PostCSS]
+according to [Browserslist] configuration stored in `.browserslistrc`.
+
+[Stylelint]: https://stylelint.io
```
mbohal, line 237: These links do not get rendered in final docs.

Also why are they here? I think we should make a hyperlink out of name used in text, but this is just a bunch of more or less related texts. 

It makes it feel like a CSS tutorial and not a library documentation.

mbohal (approved):

src/docs/contribute/css.mdx
adamkudrna: They do render, it's called [reference links](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links).

mbohal: I dont see them:
Screenshot from 2021-09-16 10-53-08

bedrich-schindler: @mbohal It is not supposed to be rendered, see the link. It is just definition so you do not have to have long links in the main texts and you use those predefined instead.

bedrich-schindler (approved):

src/docs/contribute/css.mdx
mbohal: Everyday I learn something new…

PR #313, opened 2021-09-23
Implement prop type validation of children components (#156)

### LINKED ISSUE #156: Improve propTypes validation

bedrich-schindler:
Closes #156

### DIFF
--- a/src/lib/propTypes/__tests__/elementOfType.test.jsx
+++ b/src/lib/propTypes/__tests__/elementOfType.test.jsx
@@ -1,0 +1,141 @@
+import React from 'react';
+import elementOfType from '../elementOfType';
+
+const TestWrapper = (children) => <div>{children}</div>;
+const TestItem = (children) => <div>{children}</div>;
+
+const executeOptionalElementOfType = (component) => elementOfType(['TestItem'])(
+  { children: component.props.children },
+  'children',
+  'TestWrapper',
+);
+const executeRequiredElementOfType = (component) => elementOfType(['TestItem']).isRequired(
+  { children: component.props.children },
+  'children',
+  'TestWrapper',
+);
+
+describe('elementOfType', () => {
+  it('test valid component which does not require children', () => {
+    const component = (
+      <TestWrapper />
+    );
+
+    const result = executeOptionalElementOfType(component);
+    expect(result).toBeNull();
+  });
+
+  it('test valid component which does not require children with empty React.Fragment', () => {
+    const component = (
+      <TestWrapper>
+        <>
+          <>
+          </>
+        </>
+      </TestWrapper>
+    );
+
+    const result = executeOptionalElementOfType(component);
+    expect(result).toBeNull();
+  });
+
+  it('test valid component which requires children with single child', () => {
+    const component = (
+      <TestWrapper>
+        <TestItem>Item</TestItem>
+      </TestWrapper>
+    );
+
+    const result = executeRequiredElementOfType(component);
+    expect(result).toBeNull();
+  });
+
+  it('test valid component which requires children with multiple children', () => {
+    const component = (
+      <TestWrapper>
+        <TestItem>Item</TestItem>
+        <TestItem>Item</TestItem>
+        <TestItem>Item</TestItem>
+      </TestWrapper>
+    );
+
+    const result = executeRequiredElementOfType(component);
+    expect(result).toBeNull();
+  });
+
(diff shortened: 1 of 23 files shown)


src/lib/propTypes/elementOfType.js
```diff
@@ -0,0 +1,77 @@
...
+  if (isInvalid) {
+    return new Error(`Prop \`${propName}\` of \`${componentName}\` does not contain valid element.`);
+  }
+
+  return null;
+};
+
+const elementOfType = (allowedTypes) => {
+  if (!Array.isArray(allowedTypes) || allowedTypes.length === 0) {
+    throw new TypeError('Prop type validateChildren `elementOfType` requires at least 1 type.');
```
adamkudrna, line 60: Maybe just _Prop type `elementOfType` requires at least 1 type_? 🤔

src/lib/propTypes/__tests__/elementOfType.test.jsx
```diff
@@ -0,0 +1,141 @@
...
+      <TestWrapper>
+        <TestItem>Item</TestItem>
+      </TestWrapper>
+    );
+
+    const result = executeRequiredElementOfType(component);
+    expect(result).toBeNull();
+  });
+
+  it('test valid component which requires children with multiple children', () => {
```
adamkudrna, line 53: which requires ~~children with~~ multiple children ? (multiple times)

src/lib/propTypes/__tests__/elementOfType.test.jsx
```diff
@@ -0,0 +1,141 @@
...
+          <TestItem>Item</TestItem>
+        </>
+      </TestWrapper>
+    );
+
+    const result = executeRequiredElementOfType(component);
+    expect(result).toBeNull();
+  });
+
+  it('test invalid component which requires children but any are passed in', () => {
```
adamkudrna, line 96: _but [ none / no children ] are_ (but I'm not sure)

adamkudrna (approved):
Exciting! Great job! 🚀

adamkudrna:
This is how it get rendered in docs:

obrazek

Maybe we could be more specific in descriptions? Here it could be: _Nested `Button` elements_.

src/lib/propTypes/elementOfType.js
bedrich-schindler: Yeah. I mush have accidentally paste `validateChildren` into the text.

bedrich-schindler:
> This is how it gets rendered in docs:
> 
> obrazek
> 
> Maybe we could be more specific in descriptions? Here it could be: _Nested `Button` elements_.

By the way, I requested new feature in docz so we would be able to change its description. https://github.com/doczjs/docz/issues/1657

mbohal (approved):

bedrich-schindler:
As I have conducted test within real application I have found following problems:

* Most of time, elements passed into `FormLayout` can be wrapped in higher order component as they can be connected to some state. I have decided that we should not check children of  `FormLayout`
* As you can can conditionally render component with`condition && <Component />`, it either returns component or `false`. So we should handle `false` as `null` as `undefined`.

adamkudrna:
> Most of time, elements passed into `FormLayout` can be wrapped in higher order component as they can be connected to some state. I have decided that we should not check children of  `FormLayout`

OK.

> So we should handle `false` as `null` as `undefined`.

I'm a bit lost 👀 . But please do what is necessary.

bedrich-schindler:
> > Most of time, elements passed into `FormLayout` can be wrapped in higher order component as they can be connected to some state. I have decided that we should not check children of  `FormLayout`
> 
> OK.
> 
> > So we should handle `false` as `null` as `undefined`.
> 
> I'm a bit lost 👀 . But please do what is necessary.

See commit as pushed seconds ago. @adamkudrna, @mbohal, please review those changes and approve/reject PR

bedrich-schindler:
We are closing this PR. We've decided that we will not try to validate children as we do not know whether library user will not want to use wrapped version of supported elements which is a thing we are not able to validate in simple way.

PR #315, opened 2021-09-24
Improve Contributing Guidelines to point out all requirements for pull requests

adamkudrna:


CONTRIBUTING.md
```diff
@@ -25,30 +25,49 @@ To run the dev server:
...
+
+5. **Write clear, helpful and descriptive PR names.**
+
+   1. **All rules for commit messages apply** also for PR names.
+   2. **Always check that PR name meets the requirements** above because **PR
+      names are used in changelog**. GitHub automatically truncates long PR
+      names and picks up branch name for multi-commit PRs, so it's necessary to
+      make sure the PR name is what we want to have in the changelog.
+   3. **If an issue exists for your changes, append this text to PR
+      description** in order for the issue to be closed automatically once the
```
mbohal, line 62: What is PR Description? I only see title.

adamkudrna: It's this tiny, easy-to-be-overlooked field (screenshot of new PR screen):

Snímek obrazovky 2021-09-24 v 19 44 34

👀

adamkudrna: They call it _comment_, do you think it's a better name…?

mbohal: Well Im looking at an existing PR and I do not know how to fix it. There is no field called description.

> They call it comment, do you think it's a better name

We should probably use the name GH uses…

adamkudrna: It's really confusing. The field has no label. If it's empty, it says _No description provided_. In the editing mode, the placeholder says _Leave a comment_ 🤷🏻‍♂️. So I'm using both and let's never talk about it again 🙈.

obrazek

PR #329, opened 2021-11-01
Flatten component structure on file system and shuffle components in documentation (#314)

### LINKED ISSUE #314: Component division
Presently we have components split in two groups

* `layout`
  * no `id` prop
  * not visible by themselves
* `ui`
  * do render something
  * do have `id`
  
I think we should consider adding another group for interactive components (form elements, button, etc.). These distinguish themselves from the other `ui` components by:

* logic of working with `id` is not so clear (On second thought I'm not sure if #311 was the step in right direction)
* can be disabled
* support forwarding ref
* support passing attributes to the element itself via `transferProps`

By having them in their own group it would be easier to:

* write docs as we could easily say that something is true for this group only (like passing props via `transferProps`)
* the structure would make it easy to express that some code is only to be used in this group (helpers, propTests, etc.)

 Also I think there should be some documentation describing each group and how it differs from the others.

Also should not `ButtonGroup` be in `layout` group?

mbohal:
Closes #314


src/lib/index.js
```diff
@@ -1,61 +1,61 @@
...
-} from './components/layout/Media';
+} from './components/Media';
 export {
   Toolbar,
   ToolbarGroup,
   ToolbarItem,
-} from './components/layout/Toolbar';
+} from './components/Toolbar';
 
 // UI components
```
adamkudrna, line 32: Let's get rid of these comments and flatten the exports in this file as well.

src/lib/components/ScrollView/README.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 name: ScrollView
-menu: 'UI'
-route: /components/ui/scroll-view
+menu: 'Layouts'
```
adamkudrna, line 3: Can you please move `ScrollView` into the misc. category? I've been struggling with finding the right category for it in my today's proposal, but since we are introducing misc., it's better to have it there.

adamkudrna (changes requested):
Two details and it's ready.

src/lib/index.js
mbohal: done

src/lib/components/ScrollView/README.mdx
mbohal: done

adamkudrna (approved):

PR #336, opened 2021-11-16
Update React to version 17 and update rest of dependencies

bedrich-schindler:


.eslintrc
```diff
@@ -1,23 +1,25 @@
...
-  "rules": {
-    "func-names": "off",
-    "import/prefer-default-export": "off",
-    "react/destructuring-assignment": "off",
-    "react/jsx-props-no-spreading": "off",
-    "react/jsx-one-expression-per-line": "off"
-  },
   "overrides": [
+    {
+      "files": ["**/*.md"],
```
bedrich-schindler, line 11: For unknown reason, those markdown plugins does not work when defined globally. If they are moved under `overrides`  section for MD or MDX, it works as expected. As it is not used by javascript linter by default, I think, it should be OK to have it here.

bedrich-schindler (commented):
For info:

```
Package                    Current          Wanted  Latest  Location
babel-core          7.0.0-bridge.0  7.0.0-bridge.0  6.26.3  @react-ui-org/react-ui
css-loader                   5.2.7           5.2.7   6.5.1  @react-ui-org/react-ui
docz                 2.3.3-alpha.0   2.3.3-alpha.0   2.3.1  @react-ui-org/react-ui
eslint                      7.32.0          7.32.0   8.2.0  @react-ui-org/react-ui
gatsby-plugin-sass           3.2.0           3.2.0   5.1.1  @react-ui-org/react-ui
postcss-loader               4.3.0           4.3.0   6.2.0  @react-ui-org/react-ui
sass-loader                 10.2.0          10.2.0  12.3.0  @react-ui-org/react-ui
style-loader                 2.0.0           2.0.0   3.3.1  @react-ui-org/react-ui
stylelint                  13.13.1         13.13.1  14.1.0  @react-ui-org/react-ui
webpack                     4.46.0          4.46.0  5.64.1  @react-ui-org/react-ui
webpack-cli                 3.3.12          3.3.12   4.9.1  @react-ui-org/react-ui
```

We are not able to update those plugins as they does not work with current version of docz/gatsby. Only exception is stylelint but I was not able to update it as I do not deeply understand those CSS tools. There might be some connection to other tools like postcss and that might be the reason I was not able to make it working.

adamkudrna:
OK, I feel like Docz really starts to be a pain…

.eslintrc
mbohal: Actually I think it can be simplified even further. I tested it and it seems that we use:
```
{
  "extends" : [
    "@visionappscz/eslint-config-visionapps"
  ],
  "env": {
    "browser": true,
    "jest": true
  },
  "overrides": [
    {
      "files": ["**/*.md", "**/*.mdx"],
      "extends": [
        "plugin:markdown/recommended",
        "plugin:mdx/recommended"
      ]
    }
  ],
  "parser": "@babel/eslint-parser",
  "rules": {
    "func-names": "off",
    "import/prefer-default-export": "off",
    "react/destructuring-assignment": "off",
    "react/jsx-props-no-spreading": "off",
    "react/jsx-one-expression-per-line": "off"
  }
}
```

It produces no errors and `*.md(x)` get validated. It seems that the override rules did nothing.

.eslintrc
```diff
@@ -26,5 +28,13 @@
         "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.mdx"] }] // Allow importing docz React components in *.mdx files
       }
     }
-  ]
+  ],
+  "parser": "@babel/eslint-parser",
+  "rules": {
```
mbohal, line 33: I think we only need the line `"import/prefer-default-export": "off",` here.
All other can be removed and all tests still pass.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@react-ui-org/react-ui",
   "description": "React UI is a themeable UI library for React apps.",
-  "version": "0.43.0",
+  "version": "0.44.0-alpha.0",
```
adamkudrna, line 4: Maybe `0.44.0-alpha.1`?

adamkudrna (approved):

package.json
adamkudrna: [SemVer](https://semver.org/#spec-item-9) only speaks of `-alpha` or `-alpha.1`, `-alpha.2`etc. But it's no big deal for me.

bedrich-schindler: E.g. https://github.com/npm/node-semver#prerelease-identifiers. I've also read articles about is as it should end with `.0`

bedrich-schindler: I wanted to discuss it on Monday's meeating but It did not happen.
```suggestion
  "version": "0.44.0-alpha.0",
```

bedrich-schindler:
Should be ready for round 2.

mbohal (approved):

PR #338, opened 2021-11-17
Fix undesired offset of `Grid` when rendered as `dl`, `ol`, or `ul`

adamkudrna:
So doing a fix like this is no longer necessary:

```css
/* Reset list padding until fixed in RUI. */
.root > ul {
  padding-left: 0;
}
```

### DIFF
--- a/src/lib/components/Grid/Grid.scss
+++ b/src/lib/components/Grid/Grid.scss
@@ -51,3 +51,14 @@
   grid-column: span var(--rui-local-column-span, 1); // 2.
   grid-row: span var(--rui-local-row-span, 1); // 2.
 }
+
+/* stylelint-disable selector-no-qualifying-type */
+
+dl.root,
+ol.root,
+ul.root {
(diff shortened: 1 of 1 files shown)


mbohal (approved):

src/lib/components/Grid/Grid.scss
```diff
@@ -51,3 +51,14 @@
   grid-column: span var(--rui-local-column-span, 1); // 2.
   grid-row: span var(--rui-local-row-span, 1); // 2.
 }
+
```
bedrich-schindler, line 54: Just a question, but what about other elements? Documentation says: "HTML tag to render. Can be any valid HTML tag of your choice, usually a block-level element.". So wouldn't be better to reset styles of all those elements, or of all elements (`*`)?

adamkudrna: Honestly, I spent some time thinking about it — and I don't know. Let's go with `*` and see.

PR #342, opened 2021-12-07
Refactor components which use multiple class names to use `classNames` util to unify approach of writing class names and to avoid empty spaces in output; refactor all helper files to use named exports, move all helper files from `services` and `utils` into `helpers` folder and rename all `helpers` folders on component level to `_helpers` (#270)

### LINKED ISSUE #270: create helper function for adding classNames
Presently we use the following syntax:
```
<div className={['class1', 'class2'].join(' ')} />
```

This produces empty attribute `className` if classes are added conditionally and all conditions evaluate to `flase`.

We need to create a helper function `className()` in `src/lib/services`. The usage should be:

```
<div {...className(['class1 || null', 'class2'])} />
```

it should return either `{ className: 'class1 class2' } or {} if the class list is empty.

All usages in RUI should be updated.

This function should be also available to projects depending on RUI and must be therefore well documented.

bedrich-schindler:
Closes #270


src/lib/utils/classNames.js
```diff
@@ -0,0 +1,8 @@
+export default (...classNames) => {
+  const filteredClassNames = classNames.filter(
```
mbohal, line 2: I'm not sure if this test is maybe too strict.

I was thinking more along the lines of 
```js
 const filteredClassNames = classNames.filter((className) =>  !!className);
```

Apart the unlikely case of having classes with numeric names the only argument against your solution I can think of is speed. I run a quick test in chrome:

```js
var iterations = 1000000;
console.time('Function #1');
for(var i = 0; i < iterations; i++ ){
    const test = (className) => typeof className === 'string' && className.trim().length > 0;

    test('');
    test(false);
    test(1);
    test('name');
};
console.timeEnd('Function #1')

console.time('Function #2');
for(var i = 0; i < iterations; i++ ){
    const test = (className) => !!className;

    test('');
    test(false);
    test(1);
    test('name');
};
console.timeEnd('Function #2')
VM19564:11 Function #1: 23.919189453125 ms
VM19564:22 Function #2: 5.922119140625 ms
```

So `!!` is somewhat faster, but both are pretty fast.

I would probably use `!!` but I do not feel it too strongly about it.

src/lib/utils/__tests__/classNames.js
```diff
@@ -0,0 +1,33 @@
+import classNames from '../classNames';
+
+describe('classNames', () => {
+  it('returns filtered class names', () => {
+    const result = classNames(
```
mbohal, line 5: We should probably also test with two class names.
Maybe also possibly with class name containing space?

mbohal (approved):

src/lib/utils/classNames.js
```diff
@@ -0,0 +1,8 @@
+export default (...classNames) => {
```
mbohal, line 1: I'm not sure about the placement of the file. In the issue it said it should be in `src/lib/services` and made available by export in `src/lib/index.js`.

That way we could separate helper functions:
* internal in `src/lib/utils` 
* exported in `src/lib/services`

mbohal (changes requested):

src/lib/utils/classNames.js
bedrich-schindler: Agree, it is faster. The idea has been to allow only non-empty string so that `null`, `undefined`, `<bool>`, `<number>`, ... are filtered out. Same for empty string which should not be part of class names as it creates multiple empty spaces which we have wanted to omit. And in that case there is no faster way than `typeof className === 'string' && className.trim().length > 0` check. But if we agree that we are that responsible that we do not use empty strings in conditions (I hope I removed all of them) then we can omit that.

src/lib/utils/classNames.js
bedrich-schindler: We will have to clean it out. Currently, there are `utils`,`helpers` and `services`. All three are used as "helpers".

# Current state:

## Services:
_https://github.com/react-ui-org/react-ui/blob/master/src/lib/services/elementPositionService.js_

## Utils:
_https://github.com/react-ui-org/react-ui/blob/master/src/lib/utils/transferProps.js_

## Helpers:
_https://github.com/react-ui-org/react-ui/tree/master/src/lib/helpers_
* Contains CSS helper function that return class names
* Contains `resolveContextOrProp.js`

# Proposal

In my opinion, we should have only two types of those functions. Private and public. Service is in my opinion advanced composition of logic which is not present in RUI yet. I'm proposing to have private helpers, so that all existing functions from services and utils will have to be moved under the helpers. And then, we could have publicly exposed functions like `classNames` which we shall call utils.

Nevertheless we have section `CSS helpers` in documentation. In analogy to that name, we should consider to have `JS helpers`. In that case, it might seem better to have `classNames` function in `helpers` and rest of functions under `utils`.

I would like to avoid creating `services` as it should be reserved for real services which (IMHO) does not exist yet.

What do you think @adamkudrna @mbohal? After the decision, I will create documentation page for this function.

src/lib/utils/__tests__/classNames.js
bedrich-schindler: Why not. I just wanted to cover all typical cases that can accidentally be passed into it but we test it as well.

bedrich-schindler:
@adamkudrna @mbohal Ready for round two!

src/lib/utils/classNames.js
mbohal: Please:
* move `src/lib/helpers` to `src/lib/components/_helpers`
* move `src/lib/hooks` to `src/lib/components/ScrollView/hooks`
* move `src/lib/helpers/getElementsPositionDifference` to `src/lib/components/ScrollView/_helpers`

mbohal (approved):

src/docs/js-helpers/classnames.mdx
```diff
@@ -0,0 +1,37 @@
...
+route: /js-helpers/class-names
+---
+
+# Classnames
+
+import { Playground } from 'docz'
+import { classNames} from '../../lib';
+
+The `classNames` helper joins all classnames you pass into the function as
+single classname. It automatically filters out empty strings and values that
```
adamkudrna, line 13: > values that are not strings

Hmm, but `styles.root` is an object, isn't it…? 🤔

src/lib/components/List/List.jsx
```diff
@@ -12,32 +13,32 @@ export const List = ({
     return null;
   }
 
-  let alignClass = '';
+  let alignClass;
```
adamkudrna, line 16: Wait, I was setting the type here 🙂

src/lib/utils/__tests__/classNames.js
```diff
@@ -0,0 +1,34 @@
...
+      false,
+    );
+
+    expect(result).toEqual('class-1 class-2 class-3');
+  });
+
+  it('returns undefined if all class names are filtered out', () => {
+    const result = classNames(
+      ' ',
+      ' ',
```
adamkudrna, line 23: How about `''`?

src/lib/utils/__tests__/classNames.js
```diff
@@ -0,0 +1,34 @@
...
+      0,
+      1,
+      null,
+      undefined,
+      true,
+      false,
+    );
+
+    expect(result).toEqual('class-1 class-2 class-3');
```
adamkudrna, line 17: I think we should also test objects like `styles.root` — that's what we actually use the most.

src/lib/utils/classNames.js
```diff
@@ -0,0 +1,8 @@
+export const classNames = (...classes) => {
```
adamkudrna, line 1: 8 lines of code instead of [one dependency](https://github.com/JedWatson/classnames)! 🎉 I'm curious how far we get with it.

src/lib/utils/classNames.js
adamkudrna: @adamkudrna has no idea what you are talking about. But he thinks our most frequented use case `styles.*` is somehow omitted here… (although it works!)

adamkudrna (changes requested):
8 lines of code instead of [one dependency](https://github.com/JedWatson/classnames)! 🎉 I'm curious how far we get with it.

Just one major thing:

- [x] I think we should test and document the use case of `styles.*` syntax.

src/docs/js-helpers/classnames.mdx
bedrich-schindler: @adamkudrna It not an object. If you import and use something like `styles.root` from SCSS file, webpack loader generates styles and at the same time, it replaces object with string so that `styles.root` becomes a classname. It the magic performed by loaders.

src/lib/components/List/List.jsx
bedrich-schindler: What do you mean? Before `classNames()`, you had had to explicitly pass empty string otherwise `[].join(' ')` would generate wrong classnames. Now, we can omit those empty string as undefined variable is correctly processed by `classNames()`.

src/lib/utils/__tests__/classNames.js
bedrich-schindler: We can but it is not necessary in my opinion. Because '' is subset of ' '. But if you want, we can add it, but it is unnecessary as we do not have fully covered tests.

src/lib/utils/__tests__/classNames.js
bedrich-schindler: See comment above. It is related to loader.

src/lib/utils/classNames.js
bedrich-schindler: I have already made changes in those folders and files. But I still haven't changed the detection using typeof. Should I change it or not? I would like to have this type of change instead of what Martin wrote.

adamkudrna: If I get it correctly, we agreed on keeping `typeof` today.

src/lib/components/List/List.jsx
adamkudrna: Just a joke, sorry 🙂. I mean that `''` was suggesting something like `let string: alignClass = ''`.

src/lib/utils/__tests__/classNames.js
adamkudrna: OK, didn't know that.

adamkudrna (approved):

bedrich-schindler:
>  I think we should test and document the use case of styles.* syntax.

I am still not sure about that. As I've described it earlier, this is pure function solving problem with filtering classnames and it is independent despite your software stack. If you use e.g. webpack with bunch off CSS loaders and you will use `classNames(styles.rootClassName, styles.dangerColor)` instead of `classNames("rootClassName", "dangerColor")`, it is still the same because those loaders translate `styles.rootClassName` to  string of `rootClassName` and function retrieves string during runtime, not an object or whatever it is. If you use another bundler, approach could be different but the logic is still the same. Function accepts strings or otherwise it will not work. I think that we do not want to describe none of possible implementations as the function works independently and it is totally up to you how you use it.

mbohal:
>  I think we should test and document the use case of styles.* syntax.

I agree with @bedrich-schindler . This function filters strings. En  example with `styles.root` would be needlessly complicated and it would not communicate anything that is not already described.

adamkudrna:
OK, you convinced me.

PR #347, opened 2021-12-20
Automatically check for broken links in the docs on a regular basis

adamkudrna:


.github/workflows/broken-links-check.yml
```diff
@@ -0,0 +1,18 @@
+name: Broken links check
+on:
+  schedule:
+    - cron: '0 0 * * 0' # weekly
```
mbohal, line 4: Why here? We should check them during build and fail releasing if there are broken links.

mbohal (changes requested):

.github/workflows/broken-links-check.yml
adamkudrna: This action checks all links in the final (rendered) docs website, no matter where they are defined. Since we don't have branch deployments, we have to run the check against the production environment.

Another approach would be to check only MD source files like we have it here: https://github.com/lmc-eu/cookie-consent-manager/blob/main/.github/workflows/markdown-link-check.yaml. There is a schedule too, but I can imagine it could be run against PRs.

Considering both, I find this approach safer. I doesn't prevent just **some** broken links in PRs, but it catches **all** broken links in production.

mbohal: I guess I understand, but I still dislike it.

mbohal (approved):

.github/workflows/broken-links-check.yml
adamkudrna: Approval with reservations 🙂.

Consider it a monitoring or surveillance. We can still add source files check anytime later. Both check are useful IMO.

Issue #360, opened 2022-01-14
make `translations` part of `globalProps`


mbohal:
In [RUIProvider](https://github.com/react-ui-org/react-ui/blob/master/src/lib/provider/RUIProvider.jsx) the props `globalProps` and 'translations' are separate.

However on the [component level](https://github.com/react-ui-org/react-ui/blob/master/src/lib/components/Modal/Modal.jsx), `translations` is just a prop like nay other.

We should remove the `translations` prop from the [RUIProvider](https://github.com/react-ui-org/react-ui/blob/master/src/lib/provider/RUIProvider.jsx) so as not to have these two approaches mixed.

PR #365, opened 2022-04-05
Remove the `translations` prop from `RUIProvider` as the same functionality can now be achieved with `globalProps` (#360)

mbohal:
Closes: #360 

The aim is to simplify usage by having only one way to pass props to the components through the `globalProps`.


bedrich-schindler:
I am not approving this. I have already expressed my position in this issue some time ago. The decision about this problem (and this issue) was even though made without my knowledge during my time off. If @adamkudrna approves it, the only thing is that I must accept it.

Even though translations seems like standard props to someone, I see them as very specific case. Typically, translation dictionaries are JSON files, so you can easily translate and pass whole translation dictionary via provider without touching `globalProps`. It is easy for everybody to do it this way. If we want, we can simply support multiple languages by providing multiple JSON files. Even MaterialUI separates globalProps (they call it theme) and translations. Translations are then passed to component via `localeText` object same as we does it using `translations` prop. See https://mui.com/x/react-data-grid/localization/#locale-text.

adamkudrna:
> you can easily translate and pass whole translation dictionary via provider without touching `globalProps`. It is easy for everybody to do it this way.

I agree that not mixing translations with `globalProps` seems more straightforward to me as a user/developer. When all I want is to change translations, I appreciate I don't have to learn the concept of `globalProps`.

> However on the [component level](https://github.com/react-ui-org/react-ui/blob/master/src/lib/components/Modal/Modal.jsx), translations is just a prop like nay other.

(see #360)

It may be so from the implementation point of view, but it appears to be better to keep translations separate from other props from user's point of view.

But I don't remember the discussion behind this decision. Also I may not have realized all consequences at that time.

mbohal:
>  When all I want is to change translations, I appreciate I don't have to learn the concept of globalProps

It could be rephrased to: "When I want is to change translations, why do I need to learn new concept  apart of globalProps"

And if I'm familiar with globalProps I will use them as there is no reason why not (nor technical nor in docs). And then I have to keep track of whether `globalProps` or `translations` will take priority. Also having two approaches that do not work well together is not good.

How about we remove the translations  from the component `props` and pass in `translations` via context as before? That way we loose the ability to change the text on per instance bases, but it is clear that translations and props are isolated concepts. I think we only need translations for titles of buttons that are not passed in as a components from above. As these buttons are by definition not configurable I don't think we need to be able to configure their `title`.

@bedrich-schindler @adamkudrna would you like this better?

mbohal:
@bedrich-schindler @adamkudrna OK, here is another attempt. I have removed the translations from individual component's props.

bedrich-schindler:
@mbohal It's OK. The only thing is that we have HoC component that simplifies work with globalProps in the way that it always injects correct subtree of globalProps. For translations, we have to do it now manually and we are (wrongly) able to take a look into another translations namespace. It's sad we do use same HoC that already knows name of component and therefore is is able to inject correct props. But I am approving this as it is to be able to close it finally.

bedrich-schindler (approved):

mbohal:
Agreed with @bedrich-schindler via videocall, that fixing the mentioned drawbacks is not practical as `translations` are not part of the individual components public API (props).

src/lib/components/ScrollView/ScrollView.jsx
```diff
@@ -34,9 +38,10 @@ export const ScrollView = (props) => {
     scrollbar,
     shadowColor,
     shadowSize,
-    translations,
   } = props;
 
+  const { translations } = useContext(RUIContext);
```
adamkudrna, line 43: Why is this written in two different ways? There is context consumer in `Modal` and direct assignment here.

src/docs/customize/translations.mdx
```diff
@@ -9,41 +9,9 @@ route: /customize/translations
...
-```jsx
-<Modal translations={{ close: 'Close' }}>
-  // ...
-</Modal>
-```
-
-### Global Translations
-
-To define translations globally, you need to import `RUIProvider` first:
+To use custom translations, you need to import `RUIProvider` first:
```
adamkudrna, line 14: Did you mean `RUIContext`?

adamkudrna (changes requested):
As far as I understand it, the concept looks OK to me. However, the **docs need to be fixed**, components with translation are not working now.

src/lib/components/ScrollView/ScrollView.jsx
mbohal: Because `ScrollView` uses the functional syntax and `Modal` uses the class syntax.

src/docs/customize/translations.mdx
mbohal: I think it is correct as it is.

adamkudrna: You're right. There is both context and provider but only provider is publicly exported.

src/lib/components/Alert/README.mdx
```diff
@@ -13,10 +13,7 @@ import {
 import Icon from '../../../docs/_components/Icon'
 import Placeholder from '../../../docs/_components/Placeholder'
 import Button from '../Button'
-import {
-  AlertWithContext as Alert,
-  Alert as ParsableAlert,
-} from './Alert'
+import { Alert } from './Alert'
```
adamkudrna, line 16: Even better! Thanks.

adamkudrna (approved):

PR #366, opened 2022-04-05
Unify the propTypes `label` to be always string and `value` to be alwyas either string or number

mbohal:
I must say I would prefer to always have it as string only, but we have agreed in video call that this is the desired way.


bedrich-schindler (approved):

adamkudrna (commented):
I agree with `TextArea` changes to make its API unified with `TextField`, but does anyone know why we decided to change `Badge` API…? I know its label is always rendered as string in the end, but developer experience is somewhat worse I think:

```jsx
// before

<Badge label={count} />

// after

<Badge label={`${count}`} />

// or

<Badge label={count.toString()} />
```

mbohal:
@adamkudrna 

> does anyone know why we decided to change Badge API

The issue was to unify all `label` props. I think `label` in `Badge` same as any other `label`. It is a string to be shown to the user.


> but developer experience is somewhat worse I think

Same as for all other labels (i.e. `TextField`). We can open it up and allow `label` to be either string or number, but it should be done for all `label`s. We already allow `value` to be either string or number (which I do not like) so it is not like it would make it any worse.  Honestly I don't really care as long as it is the same for all labels.

adamkudrna (approved):

PR #368, opened 2022-04-06
Rename `Link` component to `TextLink` and pass in the display value via `label` prop instead of `children`

mbohal:
This has been agreed on a video call. The reason is to allow for easier future implementation of `beforeLabel` and such props. Apparently no-text links are too different to try to have one component for both.

Passing it content via props will make it harder to use with translation libraries such as https://react.i18next.com/.

**before:**
```jsx
<Trans
  components={[<Link to={routes.home} />]}
  i18nKey="error.messageWithLinkHome"
/>
```
**after:**
```jsx
<Trans
  components={[<Link to={routes.home} label={t('routes.homePageName')} />]}
  i18nKey="error.messageWithLinkHome"
/>
```


bedrich-schindler:
It is not possible to comment it directly as its content wasn't changed, but please check styles of the component. I think that CSS custom properties should be renamed according to new name.

mbohal:
> I think that CSS custom properties should be renamed according to new name.

I was thinking, that in case we add another type of link (i.g. `ImageLink`) we might want to share these.

Also there is the file `src/lib/styles/theme/_links.scss` - should this be renamed as well?

@adamkudrna should I rename them?

bedrich-schindler (approved):

bedrich-schindler:
@mbohal The only thing I am not sure about is naming of CSS properties. Rest is fine. So follow @adamkudrna suggestion

adamkudrna:
> > I think that CSS custom properties should be renamed according to new name.
> 
> I was thinking, that in case we add another type of link (i.g. `ImageLink`) we might want to share these.
> 
> Also there is the file `src/lib/styles/theme/_links.scss` - should this be renamed as well?
> 
> @adamkudrna should I rename them?

Please do refactor component's styles but keep `styles/theme/_links.scss` (which is shared).

```
// TextLink/_theme.scss

$color: var(--rui-TextLink__color);
$text-decoration: var(--rui-TextLink__text-decoration);
$hover-color: var(--rui-TextLink--hover__color);
$hover-text-decoration: var(--rui-TextLink--hover__text-decoration);
```

1. I.e. 1) remove `link-` prefix from Sass variable names and 2) change `Link` to `TextLink` in custom property names.
2. Adjust **component's** custom property names in `src/lib/theme.scss` according to 1 ⬆️.

adamkudrna (changes requested):
Aside from tiny fixes, please change some variable names as suggested in previous comment.

adamkudrna (approved):

PR #371, opened 2022-04-26
Require Node 16 which is current LTS

adamkudrna:
Also switch to npm 8 and Ubuntu 20.04.

Downgrade `stylelint-webpack-plugin` to v2 because v3 requires Webpack 5.


.github/workflows/test.yml
```diff
@@ -1,36 +1,26 @@
...
 
 jobs:
   build_and_test:
     name: Build and test
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [ 14, 15 ]
+        node: [ 16, 18 ]
```
bedrich-schindler, line 11: As we spoke earlier, I assume that only node 16 will be tested, right?

bedrich-schindler (approved):
I just checked if it makes sense without deeper knowledge and it seems OK to me, but I did not test it.

PR #372, opened 2022-04-26
Update GitHub actions to their latest versions

adamkudrna:


.github/workflows/test.yml
```diff
@@ -1,36 +1,27 @@
...
 
 jobs:
   build_and_test:
     name: Build and test
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [ 16, 18 ]
+#        node: [ 16, 18 ]
+        node: [ 16, 17 ]
```
bedrich-schindler, line 12: I think that 17 should not be part of the testing.

bedrich-schindler (approved):

.github/workflows/test.yml
adamkudrna: It's not, it was just a test.

Issue #232, opened 2021-02-15
Refactor subgrid workarounds once it's supported in necessary browsers


adamkudrna:
Once supported:

- form fields in `FormLayout` can use subgrid instead of `display: contents` with two mandatory `div`s,
- `CheckboxField` and `Toggle` can reflect `labelPosition="before"` inside `FormLayout`.

adamkudrna:
**UPDATE September 2023:** subgrid is now supported in all major browsers including Chrome joining as last in August's version 117. However, there is some fieldset/legend magic that prevents subgrid from being used on `<fieldset>` element (as per the spec (sic!), the magic adds an additional virtual box which breaks the subgrid).

As a result, subgrid cannot be used on `<fieldset>` elements until the Chrome issue is resolved:

https://bugs.chromium.org/p/chromium/issues/detail?id=1473242

PR #375, opened 2022-05-11
Display labels of inline form fields in the same `FormLayout` column as other labels (#323)

### LINKED ISSUE #323: Display labels of inline form fields in the same `FormLayout` column as other labels
Affects `CheckboxField` and `Toggle`. (`Radio` is a combination of box field and inline fields and already works this way.)

adamkudrna:
Affects `CheckboxField` and `Toggle`.

Closes #323.

obrazek

### DIFF
--- a/src/lib/components/FormLayout/README.mdx
+++ b/src/lib/components/FormLayout/README.mdx
@@ -200,20 +200,17 @@ with CSS custom properties.
 
 ### Limitations
 
+#### Label Position
+
+Label position of inline form fields (CheckboxField, Toggle) is ignored in
+horizontal FormLayout.
+
 #### Modals
 
 Please note the `auto` and `limited` label width options may not function
 correctly in combination with other auto layout mechanisms, e.g. the auto-width
-[Modal](/components/modal). It's just too much of magic that doesn't quite
+[Modal](/components/modal). It's just too much of magic that does not quite
 work together yet 🎩.
-
-#### Inline Form Fields
-
-CheckboxField and Toggle always display their labels after inputs inside
-FormLayout. In other words, their labels never appear in a column with other
-fields' labels even when they have `labelPosition` set to `before`. This feature
-[will be implemented][rui-232] once [CSS subgrid][subgrid] is fully supported in
-[browsers that we support](/getting-started/browsers-and-devices).
 
 ## Custom Fields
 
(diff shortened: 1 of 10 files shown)


src/lib/components/FormLayout/README.mdx
```diff
@@ -402,6 +402,7 @@ This is a demo of all components supported by FormLayout.
             </FormLayoutCustomField>
             <CheckboxField
```
mbohal, line 403: In `/src/lib/components/FormLayout/README.mdx` it says:

> CheckboxField and Toggle always display their labels after inputs inside FormLayout. In other words, their labels never appear in a column with other fields' labels even when they have labelPosition set to before. This feature https://github.com/react-ui-org/react-ui/issues/232 once [CSS subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid) is fully supported in [browsers that we support](https://github.com/react-ui-org/react-ui/blob/master/getting-started/browsers-and-devices).

I think this needs to be updated.

(sorry for wrong place, but GH does not support PR wide discussions)

mbohal (approved):
I'm not really knowledgeable enough to responsibly review the code :(

src/lib/components/FormLayout/README.mdx
adamkudrna: Fixed, thanks for pointing it out!

src/lib/styles/tools/form-fields/_inline-field-layout.scss
```diff
@@ -62,12 +70,45 @@
...
+            grid-template-columns: subgrid;
+            grid-column: span 2;
+
+            @supports not (grid-template-columns: subgrid) {
+                display: contents;
+            }
+        }
+
+        &.rootLayoutHorizontal .field {
+            display: contents;
```
bedrich-schindler, line 83: Just to be sure. I remember that the last time there was an issue with `display: contents;`, we met it few days/week ago. I just want to be sure it is fine in this contexts.

bedrich-schindler (approved):
Checked it, tried it. Works perfect.

Just quick remark which it is up to you. I tried to combine FormLayout with labelPosition during the testing. Don't we want to explicitly mention how this is expected to behave?

src/lib/styles/tools/form-fields/_inline-field-layout.scss
adamkudrna: Good point 🙂. This is OK because, as opposed to the case with the cart table, we don't inherit any styling from the `.field` class. The only thing it does is defining the layout which is exactly what we need to destroy by `display: contents`.

adamkudrna:
> Just quick remark which it is up to you. I tried to combine FormLayout with labelPosition during the testing. Don't we want to explicitly mention how this is expected to behave?

Good point. I added that to the _Limitations_ section.

PR #378, opened 2022-06-17
Introduce `Popover` component (#45)

### LINKED ISSUE #45: Create `Popover` component
Something like:

image

adamkudrna:
obrazek
obrazek

## Theming

obrazek
obrazek

## Advanced Positioning

Visit https://floating-ui.com/docs/react-dom to learn how to integrate.

---

Closes #45.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@babel/preset-env": "^7.16.0",
     "@babel/preset-react": "^7.16.0",
     "@babel/register": "^7.16.0",
+    "@floating-ui/react-dom": "^0.7.1",
     "@nejcm/docz-theme-extended": "^2.0.14",
     "@testing-library/jest-dom": "^5.15.0",
     "@testing-library/react": "^12.1.2",
--- a/src/lib/components/Popover/README.mdx
+++ b/src/lib/components/Popover/README.mdx
@@ -1,0 +1,293 @@
+---
+name: Popover
+menu: 'Miscellaneous'
+route: /components/popover
+---
+
+# Popover
+
+import {
+  useFloating,
+  autoUpdate,
+  flip,
+} from '@floating-ui/react-dom'
+import {
+  Playground,
+  Props,
+} from 'docz'
+import { Placeholder } from '../../../docs/_components/Placeholder/Placeholder'
+import { Button } from '../Button/Button'
+import { ButtonGroup } from '../ButtonGroup/ButtonGroup'
+import { SelectField } from '../SelectField/SelectField'
+import { Toolbar } from '../Toolbar/Toolbar'
+import { ToolbarItem } from '../Toolbar/ToolbarItem'
+import { Popover } from './Popover'
+
+Popover displays additional information without interrupting user flow.
+
+## Basic Usage
+
+To implement the Popover component, you need to import it first:
+
+```js
+import { Popover } from '@react-ui-org/react-ui';
+```
+
+And use it:
+
+<Playground>
+{() => {
+  const [isPopoverOpen, setIsPopoverOpen] = React.useState(false);
+  // All inline styles in this example are for demonstration purposes only.
+  return (
+    <div
+      style={{
+        display: 'grid',
+        placeContent: 'center',
+        minWidth: '20rem',
+        minHeight: '10rem',
+      }}
+    >
+      <div style={{position: 'relative'}}>
+        <Button
+          aria-describedby={isPopoverOpen ? 'my-popover' : undefined}
+          label="Want to see a popover? Click me!"
+          onClick={() => setIsPopoverOpen(!isPopoverOpen)}
+        />
+        {isPopoverOpen && (
+          <Popover id="my-popover">
+            Hello there!
+          </Popover>
+        )}
+      </div>
+    </div>
+  );
+}}
+</Playground>
+
+See [API](#api) for all available options.
+
+## Placement
+
+By default, Popover is placed relative to the closest parent element with
+`position: relative` or `position: absolute`. Available placements are: top,
+right, bottom, and left. Additionally, all basic placements can be aligned to
+the center (default, no suffix), start (e.g. `top-start`), or end (e.g.
+`bottom-end`). Check Popover [API](#api) for the complete list of accepted
+values.
+
+<Playground>
+{() => {
+  const [align, setAlign] = React.useState('');
+  // All inline styles in this example are for demonstration purposes only.
+  return (
+    <>
+      <Toolbar align="baseline">
+        <ToolbarItem>
+          <span id="alignment-options-label">Alignment:</span>
+        </ToolbarItem>
+        <ToolbarItem>
+          <ButtonGroup aria-labelledby="alignment-options-label">
+            <Button
+              color={align === '-start' ? 'dark' : 'primary'}
+              label="start"
+              onClick={() => setAlign('-start')}
+            />
+            <Button
+              color={align === '' ? 'dark' : 'primary'}
+              label="center"
+              onClick={() => setAlign('')}
+            />
+            <Button
+              color={align === '-end' ? 'dark' : 'primary'}
+              label="end"
+              onClick={() => setAlign('-end')}
+            />
+          </ButtonGroup>
+        </ToolbarItem>
+      </Toolbar>
+      <div
+        style={{
+          display: 'grid',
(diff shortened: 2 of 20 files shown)


bedrich-schindler (approved):
I am just amazed! Works good with almost no code except CSS. Briliant solution! <3

src/lib/components/Popover/README.mdx
```diff
@@ -0,0 +1,294 @@
...
+    const [isPopoverOpen, setIsPopoverOpen] = React.useState(false);
+    const [placement, setPlacement] = React.useState('top');
+    const {
+      x,
+      y,
+      reference,
+      floating,
+      strategy,
+    } = useFloating({ placement });
+    const placementOptions = [
```
mbohal, line 181: For the sake of readability I would consider this syntax:
```js
    const placementOptions = [
      'top',
      'top-start',
      'top-end',
      'right',
      'right-start',
      'right-end',
      'bottom',
      'bottom-start',
      'bottom-end',
      'left',
      'left-start',
      'left-end',
    ];
    // All inline styles in this example EXCEPT Popover `style` are for
    // demonstration purposes only.
    return (
      <>
        <SelectField
          label="Placement"
          onChange={e => setPlacement(e.target.value)}
          options={placementOptions.map((el) => ({
            label: el,
            value: el,
          }))}
          value={placement}
        />
```

package.json
```diff
@@ -44,6 +44,7 @@
     "@babel/preset-env": "^7.16.0",
     "@babel/preset-react": "^7.16.0",
     "@babel/register": "^7.16.0",
+    "@floating-ui/react-dom": "^0.7.1",
```
mbohal, line 47: I think this should be in the `dependencies` section?
Maybe `peerDependencies` if it is not mandatory.

src/lib/components/Popover/README.mdx
```diff
@@ -0,0 +1,294 @@
...
+</Playground>
+
+## Advanced Positioning
+
+While the basic setup can be sufficient in some scenarios, dropping a Popover
+usually won't be so easy. To handle all possible situations and edge cases
+automatically, including smart position updates, we recommend to involve an
+external library designed specifically for this purpose.
+
+ℹ️ The following example is using an external library [Floating UI] (formerly
```
mbohal, line 145: I would remove the `formerly
Popper.js` part. I dont think it brings any value and as time goes it will be less and less relevant.

src/lib/components/Popover/README.mdx
```diff
@@ -0,0 +1,294 @@
...
+            Left side
+          </Popover>
+        </div>
+      </div>
+    </>
+  );
+}}
+</Playground>
+
+## Advanced Positioning
```
mbohal, line 138: I'm confused here. I thought that when I use the `floating-ui` library I don't have to worry about placement at all. For example, I was expecting that as I turn tablet or scroll the library would do its best to keep the popover visible.

In the demo It seems to me, that the Basic and Advanced examples behave the same.

mbohal (changes requested):

src/lib/components/Popover/README.mdx
adamkudrna: I knew there is a way! 👍🏻

package.json
adamkudrna: I think it's correct.

- It's only used in the docs as an example implementation—that's why `devDependencies`.
- It's just a **recommended** library to use along with `Popover`. You can use RUI with plain `Popover` (with no library) or even without `Popover`.

There is an [`optionalDependencies` field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies) in npm, but it does a different thing: installation just will not fail if such dependencies cannot be found or installed.

In my opinion, it's an **opt-in dependency**, therefore I think it's correct just to list it in the docs.

src/lib/components/Popover/README.mdx
adamkudrna: No strong opinion there, but people may know Popper.js mainly from Bootstrap which makes use of it. I can remove it if you insist.

src/lib/components/Popover/README.mdx
adamkudrna: Thanks for pointing it out. I tried to showcase such an example but it imposes so many questions and gotchas (beware of, do this, don't do that…) that I ended up having just the simplest example possible.

I will give it one more try.

bedrich-schindler: Personally, I am OK with present state. We just need to show how it is possible to implement it with some library (in our case - recommended Floating UI), but I do not see any benefit from showing advanced example. Different people needs deferent behavior and it is up to them how they want to have their instance configured.

src/lib/components/Popover/README.mdx
mbohal: I do insist

_(I'm rarely givent the opportunity to insist, so I want to make the best of it)_ :smile_cat:

package.json
mbohal: OK, I get it now.
Somehow I thought the integration with `floating-ui` is tighter. I guess the real issue is the docs then and that is addressed in another comment.

Thank you

src/lib/components/Popover/README.mdx
adamkudrna: OK, no more tries.

mbohal: Sorry, but I'm still confused.

Does using floating-ui actually change the behaviour? If yes, the example should reflect that and make it obvious. If not why bother mentioning it?

mbohal (changes requested):

mbohal (changes requested):

src/lib/components/Popover/README.mdx
mbohal: Also I think we should change the section title from `Advanced Positioning` to `Positioning with External Library`

adamkudrna: Finally I succeeded making a better example that is not overwhelming yet demonstrates the idea. What do you think now?

Snímek obrazovky 2022-06-25 v 16 22 16

bedrich-schindler (changes requested):

bedrich-schindler:
I am curious about this behavior:

image
image

I would expect Popover to be at the top of the screen. What happen if Popover is at the right side and there is not enough space for Popover? Have you try it?

src/lib/components/Popover/README.mdx
bedrich-schindler: It's much better now! I think that @mbohal will be OK with that :)

adamkudrna:
> I am curious about this behavior

It works as expected. Popover is placed to the other side when its content is too long to fit the room at the suggested side. (Remember it works only with Floating UI. The basic implementation puts Popover where you tell it to, no flipping, no computing, no nothing.)

But this is the kind of detail I'd like to avoid in RUI docs… 😐 

obrazek

src/lib/components/Popover/README.mdx
```diff
@@ -0,0 +1,261 @@
...
+  return (
+    <div
+      style={{
+        display: 'grid',
+        placeContent: 'center',
+        minWidth: '20rem',
+        minHeight: '10rem',
+      }}
+    >
+      <div style={{position: 'relative'}}>
```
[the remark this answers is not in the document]
adamkudrna: Hmm, interesting idea! 🤔 I put it down in a separate commit. Let me know what you think!

bedrich-schindler (approved):

src/lib/components/Popover/README.mdx
```diff
@@ -0,0 +1,333 @@
...
+
+How do you know you may need PopoverWrapper?
+
+- You are **not** rendering Popover in a React portal.
+- You are using Popover in a complex layout and it does not pop up where you
+  need it.
+- You are using Floating UI with `absolute` positioning strategy (see
+  [Advanced Positioning](#advanced-positioning) below) and your Popover keeps to
+  be misplaced.
+- You have no idea what CSS `position` is and just want to get it working.
```
mbohal, line 169: my use case is finally covered :)

mbohal (approved):

src/lib/components/Popover/README.mdx
mbohal: Yes, this is waaaaaay better.

PR #387, opened 2022-06-25
Decompose `Modal` into `Modal`, `ModalHead`, `ModalBody`, `ModalContent` and `ModalFooter` and extend its documentation (#306)

### LINKED ISSUE #306: Break down `Modal` into composable pieces
The aim is to get rid of `ScrollView` passed through API and to offer better flexibility.

Also: review `autoFocus` functionality. At least its docs need to be updated. Maybe it should be removed entirely and developers should implement it on their own.

Example:

```jsx
<Modal>
  <ModalBody>
    <ModalContent>
      ...
    </ModalContent>
  </ModalBody>
</Modal>
```

With `ScrollView`:

```jsx
<Modal>
  <ModalBody scrollable>
    <ScrollView>
      <ModalContent>
        ...
      </ModalContent>
    </ScrollView>
  </ModalBody>
</Modal>
```

Design tokens that need to be configurable are:

    * alignment: (start, center)
    * corner rounding radius
    * shadow
    * backdrop style
    * outline

bedrich-schindler:
Closes #306


src/lib/components/Modal/ModalBody.jsx
```diff
@@ -0,0 +1,58 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * ID of the root HTML element.
+   */
+  id: PropTypes.string,
+  /**
+   * If `true`, body is scroll-compatible and its child must be instance of `ScrollView`.
+   */
+  isScrollViewCompatible: PropTypes.bool,
```
bedrich-schindler, line 53: Not sure about the naming. If `ScrollView` is present as child, this must be set to `true`. But I can imagine that it can be named differently. @adamkudrna?

src/lib/components/Modal/ModalContent.jsx
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { withGlobalProps } from '../../provider';
+import { isChildrenEmpty } from '../_helpers/isChildrenEmpty';
+import styles from './ModalContent.scss';
+
+export const ModalContent = ({
```
bedrich-schindler, line 7: If `ModalContent` is used without `ModalHead` and `ModalContent` (or in combination with `ModalHead` only), padding of the component is not centered:

image

This is why I added `mt-4` into `Blocking modal without title`. @adamkudrna, I think we should add styles for all suitable combinations. Content, Head + Body (with Content), Content (with Content) + Footer. Are you able to add it into this PR? Or would you rather solve it in different PR?

adamkudrna: It's not much straightforward once `ModalBody` and `ModalContent` are separate components but this should work:

```scss
// ModalBody.scss

.root:first-child:not(.isRootScrollable) {
  padding-top: settings.$padding-y;
}
```

All other combinations are fine I think.

src/lib/components/Modal/__tests__/ModalContent.test.jsx
```diff
@@ -0,0 +1,45 @@
...
+        ),
+      },
+      (rootElement) => {
+        expect(within(rootElement).getByText('content text 1'));
+        expect(within(rootElement).getByText('content text 2'));
+      },
+    ],
+    [
+      {
+        id: 'id',
```
adamkudrna, line 29: There is a reusable `idPropTest.js` if I'm not mistaken, can we use it where applicable?

src/lib/components/Modal/__tests__/ModalHead.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    render((
+      <ModalHead
+        {...mandatoryProps}
+        onClose={spy}
+      >
+        Modal content
+      </ModalHead>
+    ));
+
+    userEvent.click(screen.getByText('×'));
```
adamkudrna, line 58: Maybe `getByRole` would be more robust…?

src/lib/components/Modal/README.mdx
```diff
@@ -12,22 +12,40 @@ import {
...
+  Button,
+  FormLayout,
+  Modal,
+  ModalBody,
+  ModalContent,
+  ModalFooter,
+  ModalHead,
+  ScrollView,
+  TextField,
+} from '../..'
```
adamkudrna, line 25: `} from '¯\_(ツ)_/¯'`

src/lib/components/Modal/README.mdx
```diff
@@ -38,18 +56,35 @@ And use it:
...
+              closeButtonRef={modalCloseButtonRef}
+              primaryButtonRef={modalPrimaryButtonRef}
             >
-              <p>
-                Do you really want to delete the user <code>admin</code>?
-                This cannot be undone.
-              </p>
+              <ModalHead
+                onClose={() => setModalOpen(false)}
+                title="Delete the user?"
```
adamkudrna, line 64: +1

src/lib/components/Modal/README.mdx
```diff
@@ -74,13 +109,167 @@ See [API](#api) for all available options.
...
+
+- `Modal` is root component which can contain `ModalHead`, `ModalBody` and
+`ModalFooter`.
+- `Modal` must always contain at least `ModalBody`.
+- `ModalBody` can contain `ModalContent` or `ModalContent` wrapped
+with `ScrollView`.
+- `ModalContent` contains desired content.
+- `ModalFooter` contains one or more `Button` components representing modal actions.
+
+Using different combinations, you can compose different kind of modals, e.g.
```
adamkudrna, line 139:
```suggestion
Using different combinations, you can compose different kinds of modals, e.g.
```

src/lib/components/Modal/README.mdx
```diff
@@ -273,86 +619,48 @@ component.
...
+                onClose={() => setModalOpen(false)}
+                title="Delete the user?"
+              />
+              <ModalBody isScrollViewCompatible={modalScrollView}>
+                {
+                  modalScrollView
+                    ? (
+                      <ScrollView
+                        customEndShadowStyle={{
+                          background: 'radial-gradient(farthest-side at center bottom, rgba(0, 0, 0, 0.16) 0%, rgba(0, 0, 0, 0.06) 40%, rgba(0, 0, 0, 0.02) 85%, rgba(0, 0, 0, 0) 100%)'
```
adamkudrna, line 638: So the scrolling shadow defaults seem to be the only thing we are losing.

Could we offload the definition into constants so this code is less cluttered maybe…?

obrazek

src/lib/components/Modal/README.mdx
```diff
@@ -74,13 +109,167 @@ See [API](#api) for all available options.
...
+  `autofocus` prop to `false`.
 
 - **Avoid stacking** of modals. While it may technically work, the modal is just
   not designed for that.
 
+## Composition
+
+Modal is decomposed into following components:
+
+- `Modal`
```
adamkudrna, line 123: As for the formatting:

- Component names are not wrapped with `code` in RUI docs to make the output less cluttered (there is already plenty of other `code` elements). This applies to all component name occurences on this page.
- Component names in the first list here could be turned into links to their API—if you agree.

obrazek

src/lib/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -146,27 +159,125 @@ describe('functionality', () => {
...
+            onClick={() => {}}
+            ref={ref}
+          />
+        </ModalFooter>
+      </Modal>
+    ));
+
+    expect(document.activeElement).toEqual(within(container.firstChild).getByRole('button'));
+  });
 });
```
adamkudrna, line 283: This one is a little bit complicated… For the record, I might have missed something here 👀. Other tests look good!

adamkudrna (changes requested):
Great job! 👏🏻 There are still a few details but I think it's on a good way. We will know more once we release it in the alpha.

src/lib/components/Modal/README.mdx
adamkudrna: Or! Can we define the shadow constants here in the `Modal` (and make them modifiable through props) and pass them down to `ScrollView` via context?

src/lib/components/Modal/__tests__/Modal.test.jsx
bedrich-schindler: You mean whole section go functional tests or just last test?

src/lib/components/Modal/README.mdx
bedrich-schindler: Definitely. Please check it in the next round and suggest further changes :)

src/lib/components/Modal/__tests__/Modal.test.jsx
bedrich-schindler: I improved naming of last two tests of autofocus functionality if it helps.

src/lib/components/Modal/README.mdx
bedrich-schindler: You are right, we are losing default configuration of ScrollView which is used within Modal.

After I read your comment, I wrapped `children` in `ModalBody` into `RUIProvider` to be able to override defaults for `ScrollView`. We use `RUIProvider` as Global Props, but It can be used anywhere. But I reverted it. Because it would override preferences of end user. Custom context does not problem neither.

There are following ways how to deal with this problem:

1. Leave it as it i.
2. Save default configuration into the constant and export it to allow user to pass this configuration using spread rest. Something like `<ScrollView {...ModalBody.scrollViewDefaultProps} />`
3. To merge `ModalBody` and `ModalContent` and implement `ScrollView` internally as it was in original implementation.
4.  To create `ModalScrollableContent` which will be only wrapper for `ScrollView`. We will then be able to configure in using Global Props. 


I kinda like 4 as it would allow Global Props configuration separately for `ScrollView` and `ModalScrollableContent`.

src/lib/components/Modal/Modal.jsx
```diff
@@ -1,52 +1,46 @@
...
-    if (autoFocus && this.childrenWrapperRef && this.childrenWrapperRef.current) {
-      const childrenWrapperElement = this.childrenWrapperRef.current;
-      const childrenElements = childrenWrapperElement.querySelectorAll('*');
-      const formFieldEl = Array.from(childrenElements).find(
-        (element) => ['INPUT', 'TEXTAREA', 'SELECT'].includes(element.nodeName),
-      );
+    if (autoFocus) {
+      if (this.childrenWrapperRef?.current != null) {
+        const childrenWrapperElement = this.childrenWrapperRef.current;
+        const childrenElements = childrenWrapperElement.querySelectorAll('*');
```
mbohal, line 31: I think we can use one the `React.Children.*` methods here (https://reactjs.org/docs/react-api.html#reactchildren).
probably `toArray()` or `map()`.

That would allow us to get rid of the ref.

src/lib/components/Modal/Modal.jsx
```diff
@@ -1,52 +1,46 @@
...
-import { transferProps } from '../_helpers/transferProps';
-import {
-  Toolbar,
-  ToolbarItem,
-} from '../Toolbar';
-import Button from '../Button';
-import ScrollView from '../ScrollView';
 import styles from './Modal.scss';
 
 export class Modal extends React.Component {
```
mbohal, line 8: I would consider refactoring it to be a functional component.

src/lib/components/Modal/README.mdx
```diff
@@ -239,19 +488,119 @@ Modal can be aligned either to top or center of the screen.
...
 
-## Scrolling Long Content
+## Key Control
+
+Modal can be controlled either by mouse or keyboard. To enhance user
+experience, primary action can be fired by pressing `Enter` key and modal
+can be closed by pressing `Escape` key.
+
+To enable it, you just need to pass reference to the buttons using
+`primaryButtonRef` and `modalCloseButtonRef` props on Modal. Advantage
```
mbohal, line 498: wrong name of the prop: `modalCloseButtonRef` -> `closeButtonRef`

src/lib/components/Modal/Modal.jsx
```diff
@@ -243,52 +142,42 @@ export class Modal extends React.Component {
...
-    color: PropTypes.oneOf(['primary', 'secondary', 'success', 'warning', 'danger', 'help', 'info', 'note', 'light', 'dark']),
-    disabled: PropTypes.bool,
-    feedbackIcon: PropTypes.node,
-    id: PropTypes.string,
-    label: PropTypes.string.isRequired,
-    onClick: PropTypes.func.isRequired,
-  })),
-  /**
-   * If `true`, focus the first action in the footer when the modal is opened.
+   * If `true`, focus the first input element in the modal or primary button when the modal is opened.
```
mbohal, line 157: I would change:
> … or primary button when the modal is opened

to:
> … or the button referenced by the `primaryButtonRef` prop when the modal is opened

src/lib/components/Modal/ModalBody.jsx
```diff
@@ -0,0 +1,58 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * ID of the root HTML element.
+   */
+  id: PropTypes.string,
+  /**
+   * If `true`, body is scroll-compatible and its child must be instance of `ScrollView`.
+   */
+  withScrollView: PropTypes.bool,
```
mbohal, line 53: Would it be possible to just check if the child is an instance of `ScrollView` or `ModalContent` via `React.Children.only()` method (https://reactjs.org/docs/react-api.html#reactchildren)?

src/lib/components/Modal/ModalHead.jsx
```diff
@@ -0,0 +1,83 @@
...
+    <h3
+      className={styles.title}
+      {...(id && { id: `${id}__title` })}
+    >
+      {title}
+    </h3>
+    {onClose && (
+      <RUIContext.Consumer>
+        {({ translations }) => (
+          <button
```
mbohal, line 29: I thought we were going to use composition for this as well.

In other words, I though, that  the usage would be more along the lines of:
```jsx
<ModalHead id="my-modal-head">
  My Modal Title
  |
  <Button
    ref={modalCloseButtonHead}
    onClick={someHandler} 
    label="Close"
  />
</ModalHead>
```

src/lib/components/Modal/ModalHead.jsx
```diff
@@ -0,0 +1,83 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import {
+  RUIContext,
+  withGlobalProps,
+} from '../../provider';
+import styles from './ModalHead.scss';
+
+export const ModalHead = ({
```
mbohal, line 9: In https://github.com/react-ui-org/react-ui/blob/master/src/lib/components/Card/README.mdx we refer to the `Card` _header_. I think we should keep this the same across all components. I think I like `Header` better, but I don't really care.

src/lib/components/Modal/ModalFooter.jsx
```diff
@@ -0,0 +1,54 @@
...
+export const ModalFooter = ({
+  children,
+  id,
+  justify,
+}) => (
+  <div
+    className={styles.root}
+    id={id}
+  >
+    <Toolbar
```
mbohal, line 19: I think I would not include `Toolbar`, I would just render the `children` directly.

```jsx
<Toolbar
  dense
  justify={justify}
>
  {children}
</Toolbar>
```

I can envision usages where the footer might prefer other ways of organizing the content. For example the user might prefer to use some custom layout component

While I agree this would be nice for our use cases I think it would be too opinionated for RUI.

The same might go for the `justify` prop.

src/lib/components/Modal/README.mdx
```diff
@@ -239,19 +488,119 @@ Modal can be aligned either to top or center of the screen.
...
+its content scrollable independently of the page itself. In this case,
+`withScrollView` must be set to `true` on ModalBody.
 
 <Playground>
   {() => {
     const [modalOpen, setModalOpen] = React.useState(false);
     const [modalScrollView, setModalScrollView] = React.useState(true);
+    const modalCloseButtonRef = React.useRef();
+    const modalPrimaryButtonRef = React.useRef();
+    const modalContent = (
```
mbohal, line 530: I would consider putting all content into one looong line. This makes reading the code (both in IDE and in the live preview) a pain.

mbohal (changes requested):

src/lib/components/Modal/Modal.jsx
```diff
@@ -243,52 +142,42 @@ export class Modal extends React.Component {
 }
 
 Modal.defaultProps = {
-  actions: [],
   autoFocus: true,
```
mbohal, line 145: I would use `autofocus`, I think it is one word: https://www.merriam-webster.com/dictionary/autofocus

bedrich-schindler: `autoFocus` is prop defined on all HTML input element in React with `F`, see https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes

src/lib/components/Modal/README.mdx
bedrich-schindler: We can't due to linter restrictions, tests would fail. And the second thing is that is untouched part written originally by @adamkudrna, I just split Modal into multiple components.

src/lib/components/Modal/ModalHead.jsx
bedrich-schindler: I don't care as well. Let it discuss it on Monday with @adamkudrna

src/lib/components/Modal/ModalHead.jsx
bedrich-schindler: Discuss this on Monday as well. We can't just decompose whole modal as it will not work for styles. And button you show in example is not the button we want to use there.

src/lib/components/Modal/ModalBody.jsx
bedrich-schindler: Yeah, it would. I was thinking and actually implementing this, but ... You can configure your own ScrollView or wrap it into some another component like MyCustomizedScrollView and it would stop working then. It is connected to the ongoing discussion above.

src/lib/components/Modal/Modal.jsx
bedrich-schindler: Unfortunately not. Generally, I don't want to touch code which functionality is proven and tested, but that is not primary reason. Primary reason is that we have to detect HTML element, not React element. Second reason is that It would require recursive calling as inputs can be place deeper in DOM tree, while querySelectorAll work like a charm and according to some articles its performance in not that bad.

src/lib/components/Modal/README.mdx
bedrich-schindler: The problem is that internal implementation of Modal's ScrollView has different appearance than it should have. I would also discuss it on Monday

src/lib/components/Modal/Modal.jsx
mbohal: OK, makes sense

src/lib/components/Modal/ModalBody.jsx
mbohal: > You can configure your own ScrollView or wrap it into some another component like MyCustomizedScrollView and it would stop working then.

I agree, but I dislike the `withScrollView` prop very much. If possible I would like to find a good solution to this as changing it later would be a BC change.

Some other non-great ideas how we could get rid of it.

1. We could check for presence of `ModalContent`. This component has very limited API so there is little need to wrap it. If the child is not `ModalContent` we now it is some version of `ScrollView` since nothing else is allowed.
2. We could check for `ScrollView` if `withScrollView` prop is undefined. That way the user does not have to set `withScrollView` if  she uses the `ScrollView` component unwrapped.

It would be ideal if we could modify the CSS to make it work regardless of weather it contains `ScrollView` or not. It would be much easier to modify `ScrollView` based on its parent via context then the other way around…

src/lib/components/Modal/README.mdx
mbohal: > We can't due to linter restrictions

I was assuming there was some way to include an inline ignore directive. There is apparently not: https://github.com/markdownlint/markdownlint/issues/16

Lets don't worry about it then

src/lib/components/Modal/Modal.jsx
mbohal: I was looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

Apparently I was looking at the wrong doc page.

src/lib/components/Modal/ModalHead.jsx
bedrich-schindler: @adamkudrna ?

src/lib/components/Modal/ModalHead.scss
```diff
@@ -0,0 +1,38 @@
...
+    align-items: baseline;
+    padding: settings.$padding-y spacing.of(4) settings.$padding-y settings.$padding-x;
+}
+
+.root h1,
+.root h2,
+.root h3,
+.root h4,
+.root h5,
+.root h6 {
```
bedrich-schindler, line 19: @adamkudrna I do not like this 💩 code, but I wrote it until you will have a time to look at it. However, I think we need to somehow solve it, otherwise its size and margins would break it

src/lib/components/Modal/README.mdx
```diff
@@ -363,58 +934,47 @@ component.
...
-                priority="flat"
-              />
-            </Modal>
-          )}
-        </div>
-      </>
-    );
-  }}
-</Playground>
+<!-- markdownlint-disable MD024 -->
```
bedrich-schindler, line 941: We have duplicate headings because of Composition section. Any idea how we want to solve it? if we add API postfix, it should be added across whole documentation.

src/lib/components/Modal/ModalHead.jsx
adamkudrna: I agree, let's go with `ModalHeader`.

src/lib/components/Modal/README.mdx
adamkudrna: In #393 I'm suggesting to support three scrolling modes:

- `auto` mode that turns on scrolling for `ModalBody`, provides basic scrolling functionality.
- `custom` mode that expects `ScrollView` (or anything else if necessary) with any design props.
- `none` mode leaving scrolling to `Modal` itself.

Also, I'm introducing customizable inner separators to make design more than acceptable no matter the scrolling mode.

obrazek

src/lib/components/Modal/ModalBody.jsx
adamkudrna: I tried and I failed. We must always tell `ModalBody` what is responsible for scrolling, there is no other way. I'm suggesting more scrolling modes in #393 so it feels more like a feature than a workaround.

src/lib/components/Modal/README.mdx
adamkudrna: I think it's alright the way it is!

adamkudrna (changes requested):
Cool! To sum it up:

- I ran the `README` through Grammarly and pushed fixes as a fixup.
- I agree with renaming `ModalHead` to `ModalHeader`.
- I did a minor redesign and API changes in #393.

src/lib/components/Modal/ModalBody.jsx
bedrich-schindler: Solved in #393

src/lib/components/Modal/ModalHead.jsx
```diff
@@ -0,0 +1,61 @@
...
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * ID of the root HTML element.
+   */
+  id: PropTypes.string,
+  /**
+   * Horizontal alignment (distribution) of individual buttons.
+   */
+  justify: PropTypes.oneOf(['center', 'space-between', 'stretch']),
```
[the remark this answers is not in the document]
bedrich-schindler: @adamkudrna Do you think we need it? The way I look at this is that even examples shown in the docs are artificially created examples just to demonstrate the possibility and I am not so happy with that. Besides that `space-between` can simulate `left`, and  `right` is not IMHO necessary, maybe in RTL environment?

So you are really sure we want to open it up like this? I am afraid of creating non-sense layouts because of those options (e.g. I have never seen title on the right and I would not expect that in LTR environments and I frankly do not know whether RTL just switches direction of text or whether the alignment changes to the right as well.)

src/lib/components/Modal/ModalFooter.jsx
```diff
@@ -0,0 +1,67 @@
...
+import { withGlobalProps } from '../../provider';
+import { classNames } from '../../utils/classNames';
+import styles from './ModalFooter.scss';
+
+export const ModalFooter = ({
+  children,
+  id,
+  justify,
+}) => {
+  const justifyClass = (value) => {
```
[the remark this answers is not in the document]
bedrich-schindler: Except the Toolbar I copied it from :D Moving it into helpers

src/lib/components/Modal/README.mdx
```diff
@@ -74,13 +114,433 @@ See [API](#api) for all available options.
...
+    const [modalOpen, setModalOpen] = React.useState(null);
+    const modalPrimaryButtonRef = React.useRef();
+    const modalCloseButtonRef = React.useRef();
+    return (
+      <>
+        <Button
+          label="Launch blocking modal without title"
+          onClick={() => {
+            setModalOpen(1);
+            setTimeout(() => setModalOpen(null), 2500);
```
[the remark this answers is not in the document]
bedrich-schindler: It would require setting up interval which would update state with countdown, so the example would be even more messy. I would expect that 2,5 sec is short enough to not to implement something like that. What @mbohal thing about it?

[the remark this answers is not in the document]
mbohal: I agree with @bedrich-schindler an example that is too complicated is a bad example.

src/lib/components/Modal/Modal.jsx
```diff
@@ -243,52 +142,43 @@ export class Modal extends React.Component {
...
+   *
+   * At least `ModalBody` is required.
    */
   children: PropTypes.node,
   /**
-   * ID of the root HTML element. It also serves as a base for nested elements:
-   * * `<ID>__content`
-   * * `<ID>__closeModalHeaderButton`
-   * * `<ID>__closeModalFooterButton`
+   * Reference to close button element. It is used to close modal when Escape key is pressed.
```
mbohal, line 172: It should say `…used to close modal when Escape key is pressed or the backdrop is clicked`

src/lib/components/Modal/ModalHead.jsx
```diff
@@ -0,0 +1,61 @@
...
+} from '../../provider';
+import { classNames } from '../../utils/classNames';
+import styles from './ModalHead.scss';
+
+export const ModalHead = ({
+  children,
+  id,
+  justify,
+}) => {
+  const justifyClass = (value) => {
```
mbohal, line 14: This also could be moved into a _helper.

src/lib/components/Modal/ModalHead.jsx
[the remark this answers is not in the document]
mbohal: I think the options should be the same as in `ModalFooter` as @adamkudrna suggests because:

1. I can envision header layout such as: 
    ```
    X | My Modal Title
    ```
2. having the same options leads to less developer confusion and less need to consult the docs
3. Code can be shared <-- I agree this is a small benefit


Note: In order to support RTL we should not use `left` / `right`, but we should use `start` / `center` (as we already do in `ModalFooter`).

src/lib/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -117,27 +104,49 @@ describe('functionality', () => {
   });
 
   it.each([
-    () => userEvent.click(screen.getByText('submit')),
-    () => userEvent.keyboard('{enter}'),
-  ])('call submit action (%#)', (action) => {
+    () => userEvent.keyboard('{esc}'),
+    () => userEvent.click(screen.getByTestId('id')),
+  ])('do not call close modal using `closeButtonRef` (%#)', (action) => {
```
mbohal, line 109: I would consider changing:
> do not call close modal using `closeButtonRef`

to 
>  do not call close modal using `closeButtonRef` when button disabled

src/lib/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -146,27 +155,126 @@ describe('functionality', () => {
...
+      (rootElement) => {
+        expect(document.activeElement).toEqual(within(rootElement).getByRole('combobox'));
+      },
+    ],
+  ])('autofocuses form field element (%#)', (child, assert) => {
+    const dom = render((
+      <Modal>
+        <ModalBody>
+          {child}
+        </ModalBody>
```
mbohal, line 257: I think we should add a primary button here to test that form elements have precedence.

mbohal (changes requested):

src/lib/components/Modal/ModalTitle.jsx
```diff
@@ -0,0 +1,45 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import {
+  withGlobalProps,
+} from '../../provider';
+import styles from './ModalTitle.scss';
+
+export const ModalTitle = ({
```
mbohal, line 8: I'm unsure about this component.

Would it be possible to not have it and pass in a `node` to have it more open?

Specifically I'm interested in this situation:
```
<Icon /> <h2>My Title</h2>                                                     <ModalCloseButton />
```
I think the current solution does not make this possible. If `node` could be passed in, we could simply do:

```jsx
<ModalHeader>
  <span>
    <Icon />
    <h2>My Title</h2>
  </span>
  <ModalCloseButton />
</ModalHeader>
```

bedrich-schindler: Note: This has been implemented by @adamkudrna in PR that was merged into this PR. Not sure how to handle it as it was already approved. In such case, @adamkudrna as author, should change it here.

@mbohal Originally, I implemented it like you, but it is more complicated than you think. Adam would describe it better, but due to some CSS things (specificity or selectors or something like that), we would not be able to obtain correct appearance of headings and the only option is dirty option with `!important` and disabling stylelint. Correct way it own component which internally solves styles. I was not able to do it, so Adam came up with `ModalTitle`. And I think, it is right way. Another libraries implement it same way, maybe because of same problems.

In case you want advanced layout, use custom layout with custom components. But I guarantee you that 98% people will not need it. Personally, I can't remember situation when anything else than text was used as title of modal - if we speak about serious web apps.

src/lib/components/Modal/ModalHead.jsx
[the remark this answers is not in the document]
bedrich-schindler: Not happy with that, but ok, lets have it same.

src/lib/components/Modal/ModalTitle.jsx
mbohal: > it is more complicated than you think

Yah, I though so…

> In case you want advanced layout, use custom layout with custom components

Sure, that would work.

Another solution might be for the `ModalTitle` not to render the `<hx>`, but just a `<span>`. The user would then pass into it the contents, typically `<h2>My title<h2>`. This would also remove the need for the `level` prop. The downsides are:

* allows users to not use the heading (`<hx>`) tags in modal titles
* in the typical use case with `<h2>` it is slightly more verbose

src/lib/components/Modal/ModalHead.jsx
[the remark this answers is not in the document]
adamkudrna: @mbohal Thanks for a nice summary. 👍🏻 

Although I'm not a fan of it, Apple prefers putting the close button on the left:

obrazek

src/lib/components/Modal/README.mdx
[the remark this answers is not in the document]
adamkudrna: OK

src/lib/components/Modal/ModalTitle.jsx
adamkudrna: I've carefully considered what @mbohal wrote and I still think this is the right solution.

1. We need `ModalTitle` to be able to style the headline.
2. We want to output the best possible markup out of the box, without making developers think about the problematics of modal headlines. Those who need it can easily adjust heading level or replace `ModalTitle` with a custom component.
   obrazek
3. Placing an icon along the headline is already possible (though it requires some styling):
   Snímek obrazovky 2022-08-01 v 17 57 20

adamkudrna (commented):
I think we are close! Please just push your (already resolved) changes, extend `ModalHeader` API and it's done.

src/lib/components/Modal/__tests__/ModalHead.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+  it.each([
+    [
+      {},
+      (rootElement) => {
+        expect(within(rootElement).getByText('content text'));
+      },
+    ],
+    ...idPropTest,
+    [
+      { justify: 'center' },
```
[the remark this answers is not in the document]
bedrich-schindler: I will also move it to separate file within `propTests` folder

src/lib/components/Modal/__tests__/Modal.test.jsx
bedrich-schindler: Fix it in another occurrences.

src/lib/components/Modal/ModalTitle.jsx
bedrich-schindler: If @mbohal is ok with it, please mark this conversation as resolved.

src/lib/components/Modal/README.mdx
```diff
@@ -994,9 +994,9 @@ opened.
 | `--rui-Modal__padding-y`                             | Block padding of individual modal components                 |
 | `--rui-Modal__background`                            | Modal background (including `url()` or gradient)             |
 | `--rui-Modal__box-shadow`                            | Modal box shadow                                             |
-| `--rui-Modal__separator__width`                      | Width of separator between modal head, body, and footer      |
-| `--rui-Modal__separator__color`                      | Color of separator between modal head, body, and footer      |
-| `--rui-Modal__head__gap`                             | Modal head gap between children                              |
+| `--rui-Modal__separator__width`                      | Width of separator between modal header, body, and footer    |
+| `--rui-Modal__separator__color`                      | Color of separator between modal header, body, and footer    |
+| `--rui-Modal__header__gap`                           | Modal header gap between children                            |
```
adamkudrna, line 999: Could you please move this line above `--rui-Modal__footer__background`? I think I messed up the order doing the redesign… 😐 Thanks!

src/lib/theme.scss
```diff
@@ -881,7 +881,7 @@
     --rui-Modal__box-shadow: none;
     --rui-Modal__separator__width: var(--rui-border-width);
     --rui-Modal__separator__color: var(--rui-color-gray-100);
-    --rui-Modal__head__gap: var(--rui-spacing-2);
+    --rui-Modal__header__gap: var(--rui-spacing-2);
```
adamkudrna, line 884: Could you please move this line above `--rui-Modal__footer__background`? 🙏🏻

adamkudrna (approved):
Cool! LGTM. 👍🏻

src/lib/components/Modal/Modal.jsx
```diff
@@ -1,145 +1,152 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, {
+  useEffect, useRef,
```
mbohal, line 3: `useRef` should be on a new line

src/lib/components/Modal/Modal.jsx
```diff
@@ -1,145 +1,152 @@
...
+        const childrenWrapperElement = childrenWrapperRef.current;
         const childrenElements = childrenWrapperElement.querySelectorAll('*');
         const formFieldEl = Array.from(childrenElements).find(
           (element) => ['INPUT', 'TEXTAREA', 'SELECT'].includes(element.nodeName),
         );
 
         if (formFieldEl) {
           formFieldEl.focus();
-          return;
+          return () => {};
```
mbohal, line 113: ~~This is exactly the same case as in @dacerondrej PR we dealt with yesterday. I think it would be better to return `undefined` here because:~~
1. ~~I think i better communicates the intention~~
2. ~~I would expect it to be more performant (I dont know though)~~

On second thought I think the `return` should not be here at all. As it is, in case I have form elements to autofocus, the `keyDown` event handler does not get cleared on unmount.

src/lib/components/Modal/ModalTitle.jsx
mbohal: @bedrich-schindler @adamkudrna I'll trust your judgment on this.

mbohal (approved):

PR #400, opened 2022-08-08
Update supported browser versions

adamkudrna:
obrazek

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@react-ui-org/react-ui",
-      "version": "0.46.0",
+      "version": "0.47.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


bedrich-schindler:
@adamkudrna Note not relevant for this repo, but we should update it on related projects as well.

bedrich-schindler:
@adamkudrna How it is reflected in package-lock.json? I see only changes in this file:

image

And my next question is, why is version of the repository updated there?

adamkudrna:
@bedrich-schindler Browser versions database (against which Browserslist queries are run) is what gets updated with `npx browserslist@latest --update-db`, that's the `caniuse-lite` package.

The version number is a little bit mystery, it had already changed [here](https://github.com/react-ui-org/react-ui/commit/84c8bbac4be7b4e8d24707f7b932290d9d8217d9)…

PR #401, opened 2022-08-09
Add support for `<optgroup>` to the `Select` component (#395)

### LINKED ISSUE #395: Add support for `<optgroup>` to the `Select` component

dacerondrej:
Closes #395

### DIFF
--- a/src/lib/components/Radio/Radio.jsx
+++ b/src/lib/components/Radio/Radio.jsx
@@ -50,34 +50,37 @@ export const Radio = ({
       <div className={styles.field}>
         <ul className={styles.list}>
           {
-            options.map((option) => (
-              <li key={option.value}>
-                <label
-                  className={styles.option}
-                  htmlFor={id && `${id}__item__${option.value}`}
-                  id={id && `${id}__item__${option.value}__label`}
-                >
-                  <input
-                    {...transferProps(restProps)}
-                    className={styles.input}
-                    checked={restProps.onChange
-                      ? (value === option.value) || false
-                      : undefined}
-                    disabled={disabled || option.disabled}
-                    id={id && `${id}__item__${option.value}`}
-                    name={id}
-                    type="radio"
-                    value={option.value}
-                  />
-                  <span
-                    className={styles.optionLabel}
-                    id={id && `${id}__item__${option.value}__labelText`}
+            options.map((option) => {
+              const key = option.key ?? option.value;
+              return (
+                <li key={key}>
+                  <label
+                    className={styles.option}
+                    htmlFor={id && `${id}__item__${key}`}
+                    id={id && `${id}__item__${key}__label`}
                   >
-                    { option.label }
-                  </span>
-                </label>
-              </li>
-            ))
+                    <input
+                      {...transferProps(restProps)}
+                      className={styles.input}
+                      checked={restProps.onChange
+                        ? (value === option.value) || false
+                        : undefined}
+                      disabled={disabled || option.disabled}
+                      id={id && `${id}__item__${key}`}
+                      name={id}
+                      type="radio"
+                      value={option.value}
+                    />
+                    <span
+                      className={styles.optionLabel}
+                      id={id && `${id}__item__${key}__labelText`}
+                    >
+                      { option.label }
+                    </span>
+                  </label>
+                </li>
+              );
+            })
           }
         </ul>
         {helpText && (
@@ -134,6 +137,9 @@ Radio.propTypes = {
    * * `<ID>__item__<VALUE>`
    * * `<ID>__item__<VALUE>__label`
    * * `<ID>__item__<VALUE>__labelText`
+   *
+   * If `key` in the option definition object is set,
+   * then `option.key` is used instead of `option.value` in place of `<VALUE>`.
    */
   id: PropTypes.string,
   /**
@@ -154,9 +160,13 @@ Radio.propTypes = {
   layout: PropTypes.oneOf(['horizontal', 'vertical']),
   /**
    * Set of options to be chosen from.
+   *
+   * For generating unique IDs the `option.value` is normally used. For cases when this is not practical or
+   * the `option.value` values are not unique the `option.key` attribute can be set manually.
    */
   options: PropTypes.arrayOf(PropTypes.shape({
     disabled: PropTypes.bool,
+    key: PropTypes.string,
     label: PropTypes.string.isRequired,
     value: PropTypes.oneOfType([
       PropTypes.string,
--- a/src/lib/components/SelectField/README.mdx
+++ b/src/lib/components/SelectField/README.mdx
@@ -78,9 +78,6 @@ See [API](#api) for all available options.
 - Use **clear, calm error messages** when there's a problem with what they
   entered.
 
-- In case of a large amount of options, consider **grouping related items
-  together** by nesting them (implements the `optgroup` HTML element).
-
 👉 We use the **native `select`** HTML element to improve user experience on
 mobile devices by using the native select of the platform.
 
@@ -233,6 +230,70 @@ Full-width fields span the full width of a parent:
           onChange={(e) => setFruit(e.target.value)}
           options={options}
           value={fruit}
+          variant="filled"
+        />
+      </>
+    );
+  }}
+</Playground>
+
+## Grouping Related Options
+
+In case of a large amount of options, consider **grouping related items
+together** by nesting them (implements the `optgroup` HTML element).
+
+<Playground>
+  {() => {
+    const [crop, setCrop] = React.useState('apple');
+    const options = [
+      {
+        label: 'Fruits',
+        options: [
+          {
+            label: 'Apple',
+            value: 'apple',
+          },
+          {
+            label: 'Banana',
+            value: 'banana',
+          },
+          {
+            label: 'Grapefruit',
+            value: 'grapefruit',
+          },
+        ],
+      },
+      {
+        label: 'Vegetables',
+        options: [
+          {
+            label: 'Beetroot',
+            value: 'beetroot',
+          },
+          {
+            label: 'Carrot',
+            value: 'carrot',
+          },
+          {
+            label: 'Tomato',
+            value: 'tomato',
+          },
+        ],
+      },
+    ];
+    return (
+      <>
+        <SelectField
+          label="Your favourite crop"
+          onChange={(e) => setCrop(e.target.value)}
+          options={options}
+          value={crop}
+        />
+        <SelectField
+          label="Your favourite crop"
+          onChange={(e) => setCrop(e.target.value)}
+          options={options}
+          value={crop}
           variant="filled"
         />
       </>
--- a/src/lib/components/SelectField/README.mdx
+++ b/src/lib/components/SelectField/README.mdx
@@ -78,6 +78,9 @@ See [API](#api) for all available options.
 - Use **clear, calm error messages** when there's a problem with what they
   entered.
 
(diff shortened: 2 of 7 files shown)


src/lib/components/SelectField/_components/OptionField.jsx
```diff
@@ -0,0 +1,44 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export const OptionField = ({
+  id,
+  option,
+}) => (
+  <option
+    disabled={option.disabled}
+    id={id && `${id}__item__${option.value}`}
```
bedrich-schindler, line 10: Typically, we compose IDs only within one component. It means that this component should only accept `id` and `${id}__item__${option.value}` composition should be made in `SelectField` (also because of fact that it must be documented in props, so it must be in sync). This is the way how it is done in another components (see Modal for example).

src/lib/components/SelectField/_components/OptionField.jsx
```diff
@@ -0,0 +1,44 @@
...
+   * Serves as a prefix for individual options:
+   * * `<ID>__item__<VALUE>`
+   */
+  id: PropTypes.string,
+  /**
+   * Set of options to be chosen from.
+   *
+   * Both option or optgroup are accepted.
+   */
+  option: PropTypes.shape({
```
bedrich-schindler, line 34: There is inconsistence between code and documentation. I think `Both option or optgroup are accepted.` should removed from here as it is not true.

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -177,15 +191,18 @@ SelectField.propTypes = {
...
-    label: PropTypes.string.isRequired,
-    value: PropTypes.oneOfType([
-      PropTypes.string,
-      PropTypes.number,
-    ]),
-  })).isRequired,
+  options: PropTypes.oneOfType([
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string,
```
bedrich-schindler, line 200: Label [mandatory according to W3C](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup#attr-label).

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -125,6 +130,15 @@ SelectField.defaultProps = {
   variant: 'outline',
 };
 
+const optionPropType = PropTypes.shape({
```
bedrich-schindler, line 133: Great idea to create definition of `option`. I would create new file `prototypes.js` and import in from there as you can share this definition with SelectField and Option then.

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -69,16 +70,20 @@ export const SelectField = ({
...
-                >
-                  {option.label}
-                </option>
-              ))
+              options.map((option) => {
+                if ('options' in option) {
+                  return (
+                    <optgroup label={option.label} key={option.label}>
+                      {option.options.map((optgroupOption) => (
+                        <OptionField id={id} key={optgroupOption.label} option={optgroupOption} />
```
bedrich-schindler, line 78: It is just note and it is up to you whether to change it or not (I don't want you to change every line of code you have written :D). But typically, we try to have at most one prop per line, in case there are more, we try to spread it across more lines, exactly one prop per line. I think we have ESlint rule for that in RACOM projects, but it might not be implemented there. @mbohal 

And when it is not enforced by ESlint, we typically do not comment those things during code review. But it is only way how we can hand over you info about this.

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -177,15 +191,18 @@ SelectField.propTypes = {
...
    */
-  options: PropTypes.arrayOf(PropTypes.shape({
-    disabled: PropTypes.bool,
-    label: PropTypes.string.isRequired,
-    value: PropTypes.oneOfType([
-      PropTypes.string,
-      PropTypes.number,
-    ]),
-  })).isRequired,
+  options: PropTypes.oneOfType([
```
bedrich-schindler, line 197: I do not see changes in tests. Please add test to test options with optgroups. And you also need to add section into documentation (README.md) about grouping.

src/lib/components/SelectField/_components/OptionField.jsx
bedrich-schindler, line 34: @mbohal @adamkudrna I would rather spread those `option` props as I do not see benefit from this, what do you think?

bedrich-schindler (changes requested):

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -69,16 +70,20 @@ export const SelectField = ({
...
-                  key={option.value}
-                  value={option.value}
-                >
-                  {option.label}
-                </option>
-              ))
+              options.map((option) => {
+                if ('options' in option) {
+                  return (
+                    <optgroup label={option.label} key={option.label}>
```
mbohal, line 76: Here `label` as is used as `key`. In the `Radio` component, we use `value` as `key`.

I'm not sure what is better:
* `label` can be translated and can contain strange characters. On the other hand, two `<options>` or `<optgroups>` with same `label` make no sense. It can probably happen in case of a missing translation or something like that.
* `value` can be same on two `<options>`. It is strange, but i can imagine this happening more ofthen then the same `label`
* we could use `nanoid`, but that is not elegant

mbohal (changes requested):

src/lib/components/SelectField/_components/OptionField.jsx
adamkudrna: I agree 👍🏻 .

src/lib/components/SelectField/_components/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './OptionField';
```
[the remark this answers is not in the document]
adamkudrna: Me too (can I say that…?).

adamkudrna (commented):
I'd like to see the updated docs too.

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -69,16 +70,28 @@ export const SelectField = ({
...
+                          {...optgroupOption}
+                          {...(id && { id: `${id}__item__${optgroupOption.value}` })}
+                        />
+                      ))}
+                    </optgroup>
+                  );
+                }
+                return (
+                  <Option
+                    key={option.label}
```
bedrich-schindler, line 89: Whenever possible, use `option.value` instead of `option.label`, because `value` is required to be unique. Label should be also unique to make sense, but it is not strong requirement. `value` is safer.

I am 100% OK with key={option.label} which is used for optgroup, but it's better that installing some dependency like nanoid. Decision should be made by @mbohal, who has already commented this few lines above.

bedrich-schindler (approved):

src/lib/components/SelectField/SelectField.jsx
mbohal: see: https://github.com/react-ui-org/react-ui/pull/401/commits/54c7af01f3b60452b67f71ba5c5b4e96a9e1afad#r943341569

src/lib/components/SelectField/SelectField.jsx
mbohal: > value is required to be unique

I'm not sure that is correct. I could not find any mention of it.

Honestly I have no strong opinion about which one is correct - I think they are both equally wrong. Either one will work in 99% of cases. So I agree we should use `value` for `<option>` and `label` for `<optgroup>`.

To deal with the 1% of cases we could allow for passing in the `key` manually by adding a new attribute to the existing `label`, `value`, `disabled` in the option definition object.

mbohal: > Should I also allow passing in the key for <optgroup>

Yes please. And I would add it to he `Radio` ccomponent as well. To have it uniform.

dacerondrej: Can I include it into the single commit or should I create separate commit for `Radio` component update?

mbohal: > Can I include it into the single commit or should I create separate commit

I think a dedicated commit would be better

src/lib/components/Radio/Radio.jsx
```diff
@@ -134,6 +137,8 @@ Radio.propTypes = {
    * * `<ID>__item__<VALUE>`
    * * `<ID>__item__<VALUE>__label`
    * * `<ID>__item__<VALUE>__labelText`
+   *
+   * If `key` in the option definition object is set, then value of the `key` attribute is used instead of `<VALUE>`.
```
mbohal, line 141: I would make it clearer by changing it to:

>  If `key` in the option definition object is set, then  `option.key` is used instead of `option.value` in place of `<VALUE>`.

src/lib/components/Radio/Radio.jsx
```diff
@@ -154,9 +159,12 @@ Radio.propTypes = {
   layout: PropTypes.oneOf(['horizontal', 'vertical']),
   /**
    * Set of options to be chosen from.
+   *
+   * If needed, `key` of each option can be set manually.
```
mbohal, line 163: I would make it clearer by changing it to:

> For generating unique IDs the `option.value` is normally used. For cases when this is not practical or the `option.value` values are not unique the `option.key` attribute can be set manually.

mbohal (approved):
:heart:

src/lib/components/SelectField/SelectField.jsx
dacerondrej: @adamkudrna is the documentation ok according to you?

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -177,15 +196,40 @@ SelectField.propTypes = {
...
-      PropTypes.string,
-      PropTypes.number,
-    ]),
-  })).isRequired,
+  options: PropTypes.oneOfType([
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        key: PropTypes.string,
+        label: PropTypes.string.isRequired,
+        options: PropTypes.arrayOf(PropTypes.shape({
```
adamkudrna, line 212: I see a nice opportunity for TS here 🙂.

src/lib/components/SelectField/README.mdx
```diff
@@ -240,6 +237,70 @@ Full-width fields span the full width of a parent:
   }}
 </Playground>
 
+## Grouping Related Options
+
+In case of a large amount of options, consider **grouping related items
+together** by nesting them (implements the `optgroup` HTML element).
```
adamkudrna, line 243: Now when I see it live and in context, I'd change the texts after all. I believe this will work better:

1. Please keep the original hint in the _General Guidelines_ section, just shorter. The advice still has its place there.

   ```
   In case of a large amount of options, consider **grouping related items
   together** by nesting them.
   ```

2. Please change the text here so it works more like API description:

   ```suggestion
   For a large amount of options you can group related items together
   by nesting them (implements the `optgroup` HTML element).
   ```

adamkudrna (changes requested):
Just a tiny detail in the docs, otherwise it looks great! 👍🏻

src/lib/components/SelectField/SelectField.jsx
adamkudrna: Almost! Please see below.

adamkudrna (approved):

Issue #402, opened 2022-08-10
Allow only spacing values in spacing props


adamkudrna:
To write:

```jsx
<Grid columnGap={3} />
```

instead of:

```jsx
<Grid columnGap="var(--rui-spacing-3)" />
```

mbohal:
Since this is a breaking change, I have moved it up to the v1 milestone.

PR #403, opened 2022-08-10
Prefer `defaultProps` over theme defaults (#308)

### LINKED ISSUE #308: Prefer `defaultProps` over theme defaults
All components should prefer `defaultProps` over theme custom properties for global defaults. The original aim was to be able to set global defaults which is now possible with global props.

Guide (add to docs!):

- Both global props and custom properties can define shared appearance across instances.
- **Global props can be overridden** by component API in component instance.
- **Custom properties cannot be overridden** via component API.

Affects mainly (but maybe not only):

- ~~`FormLayout`~~ ⛔️ False alarm, it's OK the way it is.
- ~~`Grid`~~ ✅
- ~~`ScrollView`~~ ✅
- ~~`TextField`~~ ⛔️ False alarm, it's OK the way it is.

adamkudrna:
Removed custom properties:

- `--rui-Grid__column-gap` in favour of the `columnGap` prop default of `Grid`
- `--rui-Grid__row-gap` in favour of the `rowGap` prop default of `Grid`

Closes #308.


src/lib/components/Grid/Grid.jsx
```diff
@@ -54,13 +54,13 @@ Grid.defaultProps = {
   alignItems: undefined,
   autoFlow: undefined,
   children: null,
-  columnGap: undefined,
-  columns: undefined,
+  columnGap: 'var(--rui-spacing-4)',
```
adamkudrna, line 57: I filed #402 to suggest limiting this kind of props only to allowed [spacing values](https://react-ui.io/foundation/spacing).

bedrich-schindler (approved):

mbohal (approved):
As far as I'm qualified to judge it seems fine.

hubacekj (approved):

PR #405, opened 2022-08-11
Move ref forwarding from HoC to components, unify imports (#385)

### LINKED ISSUE #385: `forwardedRef` is used instead of `ref`
I would like to clarify following:

* `ref` is prop used to reference origin HTML component.
* `forwardedRef` is prop used in many our components, but it is supposed to be used as internal implementation only and must not be used as prop.

The thing is that we have HoC  `src/lib/components/withForwardedRef.jsx` that wraps original component with `React.forwardRef`. This redirects `ref` to `forwardedRef`, so every component wrapped by this HoC accepts `ref`, not `forwardedRef`.

```js
export default (WrappedComponent) => {
  const withForwardedRef = (props, ref) => (
    <WrappedComponent {...props} forwardedRef={ref} />
  );

  withForwardedRef.displayName = `withForwardedRef(${WrappedComponent.name || WrappedComponent.name})`;

  return React.forwardRef(withForwardedRef);
};
```

```js
export const Button = ({
  forwardedRef,
  ...restProps
}) => {

export const ButtonWithGlobalProps = withForwardedRef(withGlobalProps(Button, 'Button'));

export default ButtonWithGlobalProps;
```

First problem is that API shows `forwardedRef` instead of `ref`:

image

Other problem is that code examples contain different kind of imports. Sometimes default export (the correct one) is used, sometimes named export is used (the wrong one). It results in fact that almost all examples contain `forwardedRef`, but `ref` must be used in implementation, otherwise it won't work.

hubacekj:
Apart from unifying both exports and imports in the library and documentation and unifying the use of `transferProps` function, here is a proposed solution to forwarding refs directly in the components. All test passed and everything works as expected. However, there are some warnings regarding the forwarding of refs in the HoC `withGlobalProps` even though it should be implemented correctly according to React documentation [https://reactjs.org/docs/forwarding-refs.html](url). I think that React has no way of knowing whether the component passed to `withGlobalProps` function uses `React.forwardRef`, so it gives a warning.

Closes #385


bedrich-schindler:
I see merge commit `Merge remote-tracking branch 'origin/master' into bc/385` which is not allowed in out git workflow. Rebase instead. If any help is required, please contact Martin. I am not much available today for calls.

src/lib/components/Button/Button.jsx
```diff
@@ -6,31 +6,32 @@ import { getRootColorClassName } from '../_helpers/getRootColorClassName';
...
+    priority,
+    size,
+    startCorner,
+    type,
+    color,
+    ...restProps
+  } = props;
+
   const context = useContext(ButtonGroupContext);
+  const buttonRef = ref || React.createRef();
```
bedrich-schindler, line 34: I have a question. Why don't you use `ref` directly. We don't want to create reference which won't be used. If ref is not required, we should just omit it.

src/lib/provider/withGlobalProps.jsx
```diff
@@ -1,21 +1,38 @@
...
+    forwardedRef,
+    ...rest
+  }) => (
     <RUIContext.Consumer>
       {({ globalProps }) => {
         const contextGlobalProps = globalProps ? globalProps[componentName] : null;
 
         return (
           <Component
+            ref={forwardedRef}
```
bedrich-schindler, line 16: `ref={forwardedRef}` should be placed after `restProps`

bedrich-schindler (changes requested):

src/lib/components/CheckboxField/CheckboxField.jsx
```diff
@@ -5,25 +5,25 @@ import { classNames } from '../../utils/classNames';
...
+    label,
+    labelPosition,
+    required,
+    validationState,
+    validationText,
+    value,
+    ...restProps
+  } = props;
   const context = useContext(FormLayoutContext);
+  const checkboxFieldRef = ref || React.createContext();
```
mbohal, line 26: It should say `createRef()`

src/lib/components/Button/Button.jsx
mbohal: I agree with @bedrich-schindler I think it is OK to let iot default to `undefined`

src/lib/provider/withGlobalProps.jsx
mbohal: I'm not sure.:

1. In https://reactjs.org/docs/forwarding-refs.html they have it before `...rest`.
2. The only potential unwanted override could be  if the `props` contained a `ref` prop along with the `forwardedRef` prop. i don't think that is technically possible because of this line: `return React.forwardRef((props, ref) => (<WithGlobalPropsComponent {...props} forwardedRef={ref} />));`

mbohal (approved):
I did not check every file, but I trust that @bedrich-schindler did.
What I seen seems OK

src/lib/provider/withGlobalProps.jsx
bedrich-schindler: @mbohal It's not true. They do not have any recommendation about the order. In some examples, they have it before rest props, in another after props. They don't give a damn about the order.

But_
1) We agreed on Monday that we want to have first rest props, then "local" props, so we cannot accidentally override it. 
2) In components, we always use `transferProps` which truncates those unwanted props from the object. This is not standard component, so I would preserve order I suggest and which we use in the rest of components. I am not sure whether to use transferProps in this HoC, I might leave it there without but it is up to you.

mbohal: I have no strong feelings about it, I just wanted to point to the example where it was before.

>  I might leave it there without

I agree

adamkudrna (commented):
Please don't wait for my review, this change is a little bit out of my expertise 🙂.

bedrich-schindler (approved):

Issue #354, opened 2021-12-22
Rename breakpoints so they scale better


adamkudrna:
`x2l`, `x3l` etc. might be a better choice.

Why:

- `xxxl` implies next custom breakpoints would be `xxxxl`, `xxxxxl` etc. Not much readable!
- `2xl` notation (starting with a digit) requires quotes which makes breakpoints syntax inconsistent.

PR #407, opened 2022-08-17
Limit `Grid` gap props to accept only available spacing values (#402)

### LINKED ISSUE #402: Allow only spacing values in spacing props

adamkudrna:
obrazek

## Migration Path

For `columnGap` and `rowGap`, reduce all values (responsive or not) in the format `var(--rui-spacing-X)` to corresponding numbers (just `X`), e.g.:

```jsx
// Before
<Grid columnGap="var(--rui-spacing-3)">

// After
<Grid columnGap={3}>
```

(Please mind the correct value type.)

Closes #402.

### DIFF
--- a/src/lib/components/Grid/__tests__/Grid.test.jsx
+++ b/src/lib/components/Grid/__tests__/Grid.test.jsx
@@ -18,6 +18,16 @@ const responsiveBreakpoints = {
   x2l: 'placeholder-x2l',
   x3l: 'placeholder-x3l',
 };
+
+const responsiveSpacingBreakpoints = {
+  xs: 1,
+  sm: 2,
+  md: 3,
+  lg: 4,
+  xl: 5,
+  x2l: 6,
+  x3l: 7,
+};
 /* eslint-enable sort-keys */
 
 const responsiveStyles = (infix) => ({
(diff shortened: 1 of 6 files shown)


src/lib/components/Grid/__tests__/Grid.test.jsx
```diff
@@ -18,6 +18,16 @@ const responsiveBreakpoints = {
...
   x3l: 'placeholder-x3l',
 };
+
+const responsiveSpacingBreakpoints = {
+  xs: 1,
+  sm: 2,
+  md: 3,
+  lg: 4,
+  xl: 5,
+  x2l: 6,
```
mbohal, line 28: I think we should use the more common convention `2xl` insteaed of `x2l`. In terms of ordering both solutions are equally non-elegant since we have `xs`as the first element anyway.

mbohal (approved):

src/lib/components/Grid/__tests__/Grid.test.jsx
adamkudrna: > 2xl notation (starting with a digit) requires quotes which makes breakpoints syntax inconsistent.

See #354 where the naming changed.

mbohal: Ok then I guess. :(

src/lib/components/Grid/_helpers/generateResponsiveCustomProperties.js
```diff
@@ -1,14 +1,22 @@
-export const generateResponsiveCustomProperties = (prop, infix) => {
+const prepareValueByType = (value, type) => {
+  if (type === 'spacing') {
+    return `var(--rui-spacing-${value})`;
+  }
+
+  return value;
+};
+
+export const generateResponsiveCustomProperties = (prop, infix, type) => {
```
bedrich-schindler, line 9: If `type` is not mandatory, please change it to `(prop, infix, type = null)`.

bedrich-schindler (approved):

hubacekj (approved):

PR #408, opened 2022-08-19
Add support for ref forwarding in ScrollView (#406)

### LINKED ISSUE #406: Add support for `ref` forwarding in `ScrollView`
There are some use cases where you may need to listen in the ancestor for a scroll event on the scrolling container inside the `ScrollView` component.
See for instance `Button` for implementation.

hubacekj:
Closes #406

### DIFF
--- a/src/lib/components/ScrollView/ScrollView.jsx
+++ b/src/lib/components/ScrollView/ScrollView.jsx
@@ -11,6 +11,7 @@ import {
   withGlobalProps,
 } from '../../provider';
 import { classNames } from '../../utils/classNames';
+import { transferProps } from '../_helpers/transferProps';
 import { getElementsPositionDifference } from './_helpers/getElementsPositionDifference';
 import { useLoadResize } from './_hooks/useLoadResizeHook';
 import { useScrollPosition } from './_hooks/useScrollPositionHook';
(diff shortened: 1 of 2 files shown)


src/lib/components/ScrollView/ScrollView.jsx
```diff
@@ -55,7 +57,7 @@ export const ScrollView = (props) => {
   const scrollPositionStart = direction === 'horizontal' ? 'left' : 'top';
   const scrollPositionEnd = direction === 'horizontal' ? 'right' : 'bottom';
   const scrollViewContentEl = useRef(null);
-  const scrollViewViewportEl = useRef(null);
+  const scrollViewViewportEl = ref ?? React.createRef(null);
```
bedrich-schindler, line 60: I am not sure, but I think there should be `useRef`. `useRef` remembers its state even if component is re-rendered, `React.createRef` just create reference every single pre-render.

hubacekj: There is a problem with calling a `useRef` hook conditionally. I can get around it by introducing an auxiliary variable for the hook and using it in the condition instead. Is it the correct way to address this problem?

bedrich-schindler (approved):

PR #420, opened 2022-08-22
Use `transferProps` uniformly (#404)

### LINKED ISSUE #404: use transferProps uniformly
We need to:
1. Add `transferProps` to docs for all components:
    1. link to MDN where all props are documented
    2. info on what HTML element will the props be applied
2. remove from `props` of all components all props that are only passed tot he HTML element as this can be achieved with `transferProps`

dacerondrej:
Closes #404

### DIFF
--- a/src/lib/components/Alert/README.mdx
+++ b/src/lib/components/Alert/README.mdx
@@ -204,6 +204,17 @@ click on the close button:
   }}
 </Playground>
 
+## Forwarding HTML Attributes
+
+In addition to the options below in the [component's API](#api) section, you
+can specify [React synthetic events] or **any HTML attribute you like.** All
+attributes that don't interfere with the API are forwarded to the root `<div>`
+HTML element. This enables making the component interactive and helps to improve
+its accessibility.
+
+👉 Refer to the MDN reference for the full list of supported attributes of the
+[div] element.
+
 ## API
 
 <Props table of={Alert} />
@@ -235,3 +246,6 @@ Where:
   (see alert [color variants](#color-variants) and [API](#api)),
 - `<PROPERTY>` is one of `color` (color of text), `foreground-color` (color of
   border, icon, links, and emphasis), or `background-color`.
+
+[React synthetic events]: https://reactjs.org/docs/events.html
+[div]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#attributes
--- a/src/lib/components/Button/Button.jsx
+++ b/src/lib/components/Button/Button.jsx
@@ -25,7 +25,6 @@ export const Button = React.forwardRef((props, ref) => {
     priority,
     size,
     startCorner,
-    type,
     color,
     ...restProps
   } = props;
@@ -56,7 +55,6 @@ export const Button = React.forwardRef((props, ref) => {
       disabled={resolveContextOrProp(context && context.disabled, disabled) || !!feedbackIcon}
       id={id}
       ref={ref}
-      type={type}
     >
       {startCorner && (
         <span className={styles.startCorner}>
--- a/src/lib/components/Modal/Modal.jsx
+++ b/src/lib/components/Modal/Modal.jsx
@@ -5,15 +5,16 @@ import React, {
 } from 'react';
 import { createPortal } from 'react-dom';
 import { withGlobalProps } from '../../provider';
+import { transferProps } from '../_helpers/transferProps';
 import { classNames } from '../../utils/classNames';
 import styles from './Modal.scss';
 
 const preRender = (
   children,
   childrenWrapperRef,
-  id,
   closeButtonRef,
   position,
+  restProps,
   size,
 ) => {
   const sizeClass = (modalSize) => {
@@ -47,7 +48,6 @@ const preRender = (
   return (
     <div
       className={styles.backdrop}
-      id={id}
       onClick={() => {
         if (closeButtonRef?.current != null) {
           closeButtonRef.current.click();
@@ -56,6 +56,7 @@ const preRender = (
       role="presentation"
     >
       <div
+        {...transferProps(restProps)}
         className={classNames(
           styles.root,
           sizeClass(size),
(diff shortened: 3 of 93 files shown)


src/lib/components/Button/Button.jsx
```diff
@@ -25,7 +25,6 @@ export const Button = React.forwardRef((props, ref) => {
     priority,
     size,
     startCorner,
-    type,
```
dacerondrej, line 28: Here I removed `type` from `props`, but I kept it in `defaultProps` and `propTypes`. The reason is that the default value for `type` is different among browsers ("button" in IE and "submit" in other browsers) which can lead into different behaviour. Is it ok?

src/lib/components/Modal/Modal.jsx
```diff
@@ -46,8 +47,8 @@ const preRender = (
 
   return (
     <div
+      {...transferProps(restProps)}
```
dacerondrej, line 50: According to you, is it ok to leave `transferProps` on the root backdrop `div`? I couldn't decide whether to use it as it is or to use it in child `div` for modal.

bedrich-schindler:
I haven't finished review of this PR as I am not sure whether we discussed and approved this task together (Task was created by Martin recently). On Monday, I would like to get in touch with @adamkudrna and discuss whether it is ok or not, because I personally disagree with some changes.

src/lib/components/Modal/Modal.jsx
bedrich-schindler: We decided collectively on our meeting to move it to the child div that represents modal window.

src/lib/components/Modal/README.mdx
```diff
@@ -1017,6 +1017,17 @@ opened.
 
 <!-- markdownlint-disable MD024 -->
 
+## Forwarding HTML Attributes
+
+In addition to the options below in the [component's API](#api) section, you
+can specify [React synthetic events] or **any HTML attribute you like.** All
+attributes that don't interfere with the API are forwarded either to the native
+HTML `<div>` or to the `<button>` in case of `ModalCloseButton` component. This
```
bedrich-schindler, line 1025: This would require some modification. For. example HTML tag of  `ModalTitle` can be changed, but it is heading by default. What do you think @adamkudrna ?

bedrich-schindler: What more, we need to inform end-user that `Modal` will transfer props to different element than root element

src/lib/components/Table/Table.jsx
```diff
@@ -1,25 +1,18 @@
...
-      sort,
-    } = this.props;
+export const Table = ({
+  columns,
+  id,
+  rows,
+  sort,
+  ...restProps
+}) => {
+  const renderHeaderCell = (column) => {
```
bedrich-schindler, line 15: Just personal note, because I would like to solve it in different issue. This change means that `renderHeaderCell` and `renderBodyCell` will be created and re-rendered every time when Table is re-rendered, because you are defining those components in-line. So the least change must always generate new components and render them all, React is not able to compare them between frames as they are generated each time. This differs from class methods, if I am not mistaken. This might become performance issue for larger data sets.

We can leave it here as it is and solve it later OR you can create **internal/private** components in `_components/TableHeaderCell` and `_components/TableBodyCell` to solve it now.

cc @adamkudrna

src/lib/components/TextLink/TextLink.jsx
```diff
@@ -5,34 +5,18 @@ import { transferProps } from '../_helpers/transferProps';
...
 
-TextLink.defaultProps = {
-  id: undefined,
-};
-
 TextLink.propTypes = {
-  /**
-   * Link's `href` attribute.
-   */
-  href: PropTypes.string.isRequired,
```
bedrich-schindler, line 31: I think we must preserve this prop. @adamkudrna ?

bedrich-schindler:
Tests are passing but you should remove tests of props of props you removed.

bedrich-schindler (changes requested):

bedrich-schindler (changes requested):
Tests are passing but you should remove tests of props of props you removed.

src/lib/components/Modal/README.mdx
dacerondrej: Please take a look at changes made.

src/lib/components/Table/Table.jsx
dacerondrej: Internal/private components were created. Can you please check if it is ok? @bedrich-schindler

dacerondrej:
I tried to fix up the Modal API which is not showing up as Adam reported, but without any success. Does anybody know what could cause this problem?

src/lib/components/Alert/README.mdx
```diff
@@ -204,6 +204,17 @@ click on the close button:
   }}
 </Playground>
 
+## Forwarding HTML Attributes
+
+In addition to the options below in the [component's API](#api) section, you
+can specify [React synthetic events] or **any HTML attribute you like.** All
+attributes that don't interfere with the API are forwarded to the native HTML
```
mbohal, line 211: I dislike the phrase:

> …are forwarded to the native HTML `div`

I think the phrase `native HTML*` makes sense in case of interactive elements that we wrap. In case we transfer the props to the root `div` of some presentation component I would use something along the lines:

> …are forwarded to the root  `div` HTML element

mbohal (approved):

src/lib/components/Alert/README.mdx
bedrich-schindler: I would like to point out it as not always root element (ScrollView, Modal). I am OK with both variants. @adamkudrna?

bedrich-schindler (approved):

src/lib/components/Alert/README.mdx
mbohal: > But this is about Alert, right?

No, this is applicable to many components. most non form components I think.

> I would like to point out it as not always root element

The text has to be altered to reflect how transferProps are handled for the given component. by definition it can not be same for all components.

dacerondrej: Please, check the last commit. I kept original formulation `native HTML *` for components where `transferProps` are applied to `<button>` or `<input>` HTML element. For remaining components was used the formulation suggested by @mbohal.

mbohal: I like it.

PR #424, opened 2022-09-14
Refactor CSS class names according to new guidelines (#310)

### LINKED ISSUE #310: Refactor CSS class names according new guidelines
Guidelines: https://react-ui.io/contribute/css

Also check tests:
https://github.com/react-ui-org/react-ui/pull/301#discussion_r708129668

hubacekj:
Closes #310

### DIFF
--- a/src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
+++ b/src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
@@ -3,6 +3,7 @@ import {
   render,
   within,
 } from '@testing-library/react';
+import { blockPropTest } from '../../../../../tests/propTests/blockPropTest';
 import { childrenEmptyPropTest } from '../../../../../tests/propTests/childrenEmptyPropTest';
 import { Button } from '../../Button';
 import { ButtonGroup } from '../ButtonGroup';
@@ -13,14 +14,7 @@ const mandatoryProps = {
 
 describe('rendering', () => {
   it.each([
-    [
-      { block: true },
-      (rootElement) => expect(rootElement).toHaveClass('isRootBlock'),
-    ],
-    [
-      { block: false },
-      (rootElement) => expect(rootElement).not.toHaveClass('isRootBlock'),
-    ],
+    ...blockPropTest,
     ...childrenEmptyPropTest,
     [
       { children: <Button label="label text" /> },
@@ -36,27 +30,27 @@ describe('rendering', () => {
     ],
     [
       { priority: 'filled' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityFilled'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityFilled'),
     ],
     [
       { priority: 'outline' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityOutline'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityOutline'),
     ],
     [
       { priority: 'flat' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityFlat'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityFlat'),
     ],
     [
       { size: 'small' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootSizeSmall'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootSizeSmall'),
     ],
     [
       { size: 'medium' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootSizeMedium'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootSizeMedium'),
     ],
     [
       { size: 'large' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootSizeLarge'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootSizeLarge'),
     ],
   ])('renders with props: "%s"', (testedProps, assert) => {
     const dom = render((
--- a/src/lib/components/Card/__tests__/Card.test.jsx
+++ b/src/lib/components/Card/__tests__/Card.test.jsx
@@ -9,6 +9,7 @@ import { ScrollView } from '../../ScrollView';
 import { Card } from '../Card';
 import { CardBody } from '../CardBody';
 import { CardFooter } from '../CardFooter';
+import { densePropTest } from '../../../../../tests/propTests/densePropTest';
 
 const mandatoryProps = {
   children: <CardBody>card body content</CardBody>,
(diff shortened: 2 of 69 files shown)


mbohal:
Za mě to vypadá dobře, ale moc tomu nerozumím, nechám na @adamkudrna

mbohal (approved):

src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
```diff
@@ -38,27 +32,27 @@ describe('rendering', () => {
     ...idPropTest,
     [
       { priority: 'filled' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityFilled'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityFilled'),
```
bedrich-schindler, line 35: What about replacing `filled` tests with shared prop test as you did with `block`? 

I see that Badge.jsx has reduced set of options, so it might be right to do it this way. 

What do you think @adamkudrna ?

src/lib/components/Card/__tests__/Card.test.jsx
```diff
@@ -36,11 +36,11 @@ describe('rendering', () => {
     ...colorPropTest,
     [
       { dense: true },
-      (rootElement) => expect(rootElement).toHaveClass('rootDense'),
+      (rootElement) => expect(rootElement).toHaveClass('isRootDense'),
```
bedrich-schindler, line 39: Might be replaced with shared test as `dense` is used in multiple props.

src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
```diff
@@ -38,27 +32,27 @@ describe('rendering', () => {
...
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityOutline'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityOutline'),
     ],
     [
       { priority: 'flat' },
-      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('rootPriorityFlat'),
+      (rootElement) => expect(within(rootElement).getByRole('button')).toHaveClass('isRootPriorityFlat'),
     ],
     [
       { size: 'small' },
```
bedrich-schindler, line 46: What about replacing `size`  tests with shared prop test as you did with block?

Only Modal has different set, but other components share the same set of size options.

What do you think @adamkudrna ?

bedrich-schindler (approved):
This have to be reviewed primarily by @adamkudrna. I suggested some changes by creating "shared prop tests" for cases where those shared tests are not currently used. But it might be done in different issue, I was just inspired by @hubacekj  who created new shared test for `block`

src/lib/components/Card/__tests__/Card.test.jsx
hubacekj: This is a bit tricky. Dense prop is used in `Card`, `Toolbar` and `ToolbarGroup`. In toolbar related components we use a single SCSS file to re-use code, which means that there is `.isDense` class for both components (@adamkudrna is it okay this way? Or should it be `.isToolbarDense` for `Toolbar` and `.isGroupDense` for `ToolbarGroup`?). In `Card` we use `isRootDense` because this class is used only for the root element. It does make sense to me this way but it makes the `densePropTest` usable either for the toolbar related components only or the `Card` component only.

src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
hubacekj: If you look closely, all components that could use `priorityPropTest` have tests related to `priority` prop described differently (such as `expect(rootElement)...` and `expect(within(rootElement)...` . So I don't think it would be helpful, I had tried it before I realised this. That's the reason why I left it this way.

src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
hubacekj: Those components that need a simple `rootElement` test do use a common `sizePropTest`. Others that are described differently have their own tests.

src/lib/components/Card/__tests__/Card.test.jsx
adamkudrna: The code of `Toolbar` seems a little bit obsolete in more ways…

1. I'm aware of the inconsistence of `.isDense`, `.isAlignedToTop` etc. Back in those days, the idea was to save some CSS and share everything shareable. However it's inconsistent with the rest of the code base. I think now it's a great chance to fix it. That means being **more explicit in modifier class names**, i.e. have separate class names for individual sub-components:

   ```css
   .isToolbarAlignedToTop,
   .isGroupAlignedToTop {
       align-items: flex-start;
   }
   ```

   or:

   ```css
   :is(.isToolbarDense, .isGroupDense),
   :is(.isToolbarDense, .isGroupDense) .group {
       gap: theme.$gap-dense;
   }
   ```

   or:

   ```css
   :is(.isToolbarNowrap, .isGroupNowrap) > .item:not(.isItemFlexible) {
       flex: 0 1 auto;
   }
   ```
2. According to the latest convention, `alignClass` and `justifyClass` **helper functions** should be placed in the `_helpers` directory, although class name splitting above will make this more wordy.
3. Looking at the modifier classes, I think we could further unify their names instead of attempting to build sentence-like names: `.isToolbarAlignedToTop` ➡️ `.isToolbarAlignTop`, etc. The pattern would be:
   1. `.is<ELEMENT><PROP-NAME><PROP-VALUE>` for `oneOf` prop types, e.g. `.isRootColorPrimary`,
   2. `.is<ELEMENT><PROP-NAME>` for truthy boolean prop types, e.g. `.isRootBlock`.

   If you agree, I'd do it in a separate PR (or a commit at least) and also [document it](https://react-ui.io/contribute/css#class-naming-rules) in part 4.

adamkudrna (changes requested):
Nice!

- [x] Looks like only the `Toolbar` requires more attention.
- [x] I also checked all styles _not_ changed in this PR and it turns out `ModalHeader` and `ModalFooter` need to be updated too.
- [ ] Additional unification of modifier class names syntax is to be considered and done (if agreed) in a separate PR.

src/lib/components/ButtonGroup/__tests__/ButtonGroup.test.jsx
hubacekj: Did you mean `size` and `priority` prop tests? I don't see how it would benefit us in these cases. Configurable `sizePropTest` would be done only to add its support for `Modal` and `ButtonGroup`. I can't think of any other way to implement it than just to move the test descriptions into an if statement inside a function returning the right tests. I believe it would be pretty much the same amount of code in exchange for it being less readable. It's similar for `priorityPropTest`.
I made `Toolbar` related tests configurable, though. It came in handy there!

src/lib/components/Card/__tests__/Card.test.jsx
hubacekj: ad 1 & 2: Should be done.
ad 3: Ok, I'm going to create new issue for it.

src/lib/components/Toolbar/Toolbar.jsx
```diff
@@ -4,6 +4,8 @@ import { withGlobalProps } from '../../provider';
 import { transferProps } from '../_helpers/transferProps';
 import { classNames } from '../../utils/classNames';
 import { isChildrenEmpty } from '../_helpers/isChildrenEmpty';
+import { getJustifyClassName } from './_helpers/getJustifyClassName';
+import { getAlignClassName } from './_helpers/getAlignClassName';
```
adamkudrna, line 8: 🔤

webpack.config.js
```diff
@@ -3,7 +3,7 @@ const StyleLintPlugin = require('stylelint-webpack-plugin');
 const VisualizerPlugin = require('webpack-visualizer-plugin');
 
 const MAX_DEVELOPMENT_OUTPUT_SIZE = 2900000;
-const MAX_PRODUCTION_OUTPUT_SIZE = 330000;
+const MAX_PRODUCTION_OUTPUT_SIZE = 335000;
```
adamkudrna, line 6: So this is the price of conformance to guidelines… 🙂

adamkudrna (approved):
Great job, thank you!

PR #427, opened 2022-09-26
Fix console warning due to using ref in PropTypes just for documentation (#426)

### LINKED ISSUE #426: Fix console warning due to using `ref` in `PropTypes` just for documentation
Image

hubacekj:
Closes #426

### DIFF
--- a/src/lib/components/Button/Button.jsx
+++ b/src/lib/components/Button/Button.jsx
@@ -103,7 +103,6 @@ Button.defaultProps = {
   id: undefined,
   labelVisibility: 'xs',
   priority: 'filled',
-  ref: undefined,
   size: 'medium',
   startCorner: null,
   type: 'button',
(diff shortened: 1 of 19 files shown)


mbohal (approved):

bedrich-schindler:
It is subjective opinion, but I would prefer "If you provide [ref] prop, it is forwarded ..." instead of "If you provide either [object ref or callback ref], it is forwarded ...". Why? While `ref` prop was removed from API table (due to React error), I would keep it as simple as it gets, because if you need to use `ref`, you already know how does it work. Besides that, object ref is not official naming so I was confused for couple of seconds.

But if it is OK for @adamkudrna and @mbohal, I am OK with that.

hubacekj:
> I would prefer "If you provide [ref] prop, it is forwarded ..." instead of "If you provide either [object ref or callback ref], it is forwarded ..."

I was trying to avoid the word prop as it is what React was complaining about in the first place. I agree that my unnecessarily complicated expression might be confusing, though. Would it be ok with you and others to use just [ref] as a link to the documentation?

mbohal:
> I was trying to avoid the word prop

I think that is good thinking and we should avoid the word `prop`

The simpler sentence:

I think that `If you provide [ref], it is forwarded ...` is also good.

adamkudrna (approved):
Looks good! The only thing I was missing was a screenshot of the problem. There was no description in the issue, so I fixed it.

PR #432, opened 2022-11-22
Prevent page from scrolling when  is open (#397)

### LINKED ISSUE #397: Prevent page from scrolling when `Modal` is open
Currently, when `Modal` is opened, we can scroll page behind the modal. Other libraries like Material UI, Ant design prevent page from scrolling when modal is displayed. We agreed to implement same behavior for `Modal` component.

hubacekj:
I am not that confident of this implementation. Any suggestions are welcome. @adamkudrna 

Closes #397


dacerondrej (approved):

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@react-ui-org/react-ui",
-      "version": "0.47.0",
```
bedrich-schindler, line 9: Why is this change here? It should have been already changed during the release, shouldn't it? (cc @adamkudrna)

If it fixes something, please do it in separate commit.

src/lib/components/Modal/Modal.jsx
```diff
@@ -97,9 +97,14 @@ export const Modal = ({
   };
 
   useEffect(() => {
+    const bodyElement = document.getElementsByTagName('body')[0];
+    bodyElement.style = 'overflow-y: hidden';
```
bedrich-schindler, line 101: I am not happy with this implementation. To be honest, I do not now how to do it properly as this is recommended approach on CSS tricks, but I am afraid this would not work in all cases. 

For example. Some applications scroll whole body and some applications have div with size (top,bottom,left,right=0 or width/height=100%), so this solution would cover only some cases.

@adamkudrna, can you review it and help with this? You are the only one with such knowledge to do it properly, it is too much for me.

package-lock.json
hubacekj: It came from a regular npm install, I am not sure why it is there either.

src/lib/components/Modal/Modal.jsx
bedrich-schindler: The problem is that our implementation (where RUI is used) does not have scrollable body, so this change wouldn't help us. We might need to come with another solution :/

hubacekj: @bedrich-schindler @adamkudrna The only other viable solution I can see is listening to all scroll events when `Modal` is open and unless the event target is any of the `Modal` children ignore it.

mbohal: Agreed upon on a video call:

* If possible we would like to make all scroll events affect the modal body and not the page underneath regardless of the event source (keyboard, mouse, touchpad etc.). Technical solution is unknown.
* keyboard scroll events must take into account form fields (e.g. `<textarea>`)
* we want to use the `<dialog>` component
* when no active form element or primary button is found and `autofocus` prop is `true` we want to focus the `<dialog>` element

Relevant links:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations
* https://github.com/whatwg/html/wiki/dialog--initial-focus,-a-proposal

Alternate solution is to pass in the page scroll prevention function as a prop.

adamkudrna, line 101: How about:

```suggestion
    window.document.body.style.overflowY = 'hidden';
```

(`window` is optional, but let's keep it the same as below)

adamkudrna: I checked our idols, MUI and Bootstrap 🎉 , and they both prevent body from scrolling. That makes me think:

1. Yes, I'd like to prevent body from scrolling by default.
2. It should opt-out so I can disable it when I don't like it, and…
3. … it could be customizable so I can pass a custom function as @mbohal suggests.

adamkudrna, line 101: I'm afraid the blocking feature needs to be smarter, at least it can cover the varying scrollbar width across OS's…

https://github.com/twbs/bootstrap/blob/main/js/src/util/scrollbar.js#L34

hubacekj: Migration to native `<dialog>` element will be implemented in https://github.com/react-ui-org/react-ui/issues/461

src/lib/components/Modal/README.mdx
dacerondrej, line 1: I would simplify the code here in the `Playground` - `children` of all three modals look similar. It can be moved to a variable.

```suggestion
const modalChildren = (
      <>
        <ModalHeader>
          <ModalTitle>Delete the user?</ModalTitle>
          <ModalCloseButton onClick={() => setModalOpen(false)} />
        </ModalHeader>
        <ModalBody>
          <ModalContent>
            <p>
              Do you really want to delete the user <code>admin</code>? This
              cannot be undone.
            </p>
          </ModalContent>
        </ModalBody>
        <ModalFooter>
          <Button
            color="danger"
            label="Delete"
            onClick={() => setModalOpen(false)}
            ref={modalOpen ? modalPrimaryButtonRef : null}
          />
          <Button
            color="secondary"
            label="Close"
            onClick={() => setModalOpen(false)}
            priority="outline"
            ref={modalOpen ? modalCloseButtonRef : null}
          />
        </ModalFooter>
      </>
);
```

src/lib/components/Modal/README.mdx
```diff
@@ -1007,6 +1007,154 @@ independent of the page itself. This can be done in three ways using the
...
+
+You can choose the mode in which Modal prevents scroll of the page underneath. Default
+mode prevents scroll on `<body>` element. If you choose `off`, there will be no scroll
+prevention. If you need more flexibility, define your own methods `start` (called
+on Modal's mount) and `reset` (called on Modal's unmount) wrapped by an object
+and handle scroll prevention yourself.
+
+<Playground>
+  {() => {
+    const [modalOpen, setModalOpen] = React.useState(null);
```
dacerondrej, line 1020: Considering the fact that TypeScript is currently being introduced to react-ui, I dislike that `modalOpen` takes values:

- `null`
- `false`
- 1, 2, 3

src/lib/components/Modal/Modal.jsx
```diff
@@ -78,54 +79,120 @@ export const Modal = ({
...
+      childrenFocusableElements.push(childrenWrapperElement);
+    }
+
+    const firstFocusableElement = childrenFocusableElements[0];
+    const lastFocusableElement = childrenFocusableElements[childrenFocusableElements.length - 1];
+
+    if (firstFormFieldEl) {
+      firstFormFieldEl.focus();
+    } else if (primaryButtonRef?.current != null) {
+      primaryButtonRef.current.focus({ focusVisible: true });
```
hubacekj, line 127: The `focusVisible` option works only in FF. It would be nice to have similar behaviour (to indicate we moved focus in JS by showing outline - by default it does not show on buttons) in other browsers too. Any suggestions @adamkudrna ?

src/lib/components/Modal/Modal.jsx
```diff
@@ -78,54 +79,120 @@ export const Modal = ({
...
-    if (e.key === 'Escape' && closeButtonRef?.current != null) {
-      closeButtonRef.current.click();
+  useEffect(() => {
+    if (!autoFocus) {
+      return () => {};
     }
 
-    if (e.key === 'Enter' && e.target.nodeName !== 'BUTTON' && primaryButtonRef?.current != null) {
-      primaryButtonRef.current.click();
+    // If `autoFocus` is set to `true`, following code finds all focusable elements and among them
```
adamkudrna, line 97: > following code

How about moving this into a function with a self-explanatory name like `focusFirstFocusableElement(autoFocus, enableFocusTrap)`? Or maybe even external helpers? The code here starts to get a little bit long and more difficult to read.

Let @bedrich-schindler or @mbohal decide, it's just my personal feeling.

src/lib/components/Modal/Modal.jsx
```diff
@@ -78,54 +79,120 @@ export const Modal = ({
...
+        && e.target.nodeName !== 'TEXTAREA'
+        && e.target.nodeName !== 'A'
+        && primaryButtonRef?.current != null
+      ) {
+        primaryButtonRef.current.click();
       }
+
+      // Following code enforces focus to stay inside Modal
+
+      if ((e.key !== 'Tab' && e.keyCode !== 9) || !focusEnforce) {
```
adamkudrna, line 149: I think these key codes (not only here) deserve their constants, at least the numeric one.

src/lib/components/Modal/Modal.jsx
```diff
@@ -78,54 +79,120 @@ export const Modal = ({
...
+      const scrollbarWidth = Math.abs(window.innerWidth - window.document.documentElement.clientWidth);
+      const prevOverflow = window.document.body.style.overflow;
+      const prevPaddingRight = window.document.body.style.paddingRight;
+
+      window.document.body.style.overflow = 'hidden';
+      if (Number.isNaN(parseInt(prevPaddingRight, 10))) {
+        window.document.body.style.paddingRight = `${scrollbarWidth}px`;
+      } else {
+        window.document.body.style.paddingRight = `calc(${prevPaddingRight} + ${scrollbarWidth}px)`;
+      }
```
adamkudrna, line 184: Cool 🙂

adamkudrna (commented):
Nicely done! 👏🏻 My major comments are:

- focus trap vs. focus enforce,
- code readability because of long hooks.

src/lib/components/Modal/Modal.jsx
bedrich-schindler: Why not. I just want to mention that if we want to make it nicer, I would refactor whole component. One small extraction to the helper wouldn't help so much and it would be even more messy because nobody would know "the style in which is component implemented".

bedrich-schindler: Tell me whether you will refactor it or not. If yes, I will review it after the component is refactored.

src/lib/components/Modal/Modal.jsx
```diff
@@ -184,6 +253,10 @@ Modal.propTypes = {
     // eslint-disable-next-line react/forbid-prop-types
     current: PropTypes.any,
   }),
+  /**
+   * Enforces focus to stay inside of Modal. Works only if autoFocus is not disabled.
+   */
+  focusEnforce: PropTypes.bool,
```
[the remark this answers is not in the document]
hubacekj: Since I think that focus trap should be enabled by default, I'd prefer disableFocusTrap. Option to disable it should take some action for users that know what they are doing.

src/lib/components/Modal/Modal.jsx
```diff
@@ -184,6 +253,10 @@ Modal.propTypes = {
     // eslint-disable-next-line react/forbid-prop-types
     current: PropTypes.any,
   }),
+  /**
+   * Enforces focus to stay inside of Modal. Works only if autoFocus is not disabled.
```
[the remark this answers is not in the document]
hubacekj: I wanted to express the fact that `autoFocus` is enabled by default and it needs to be explicitly disabled, which then also disables focus trap making no sense provided the focus is not moved into the `Modal`.

src/lib/components/Modal/Modal.jsx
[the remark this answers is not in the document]
mbohal: We do not use imperative verbs for boolean props. We use `raised`, `muted`, `disabled` etc. (adjectives I gues).

However I would prefer naming the prop `focusTrap` as that in my view is descriptive enough. We can set it default true, that should cause no confusion.

@hubacekj I see where you are coming from, but we try to stay away from names in negative. It makes the code harder to read when you have double and triple negatives such as `if (!disableFocusTrap)…`. It is a convention we maintain across code bases (I think at least).

src/lib/components/Modal/README.mdx
```diff
@@ -825,21 +826,157 @@ To enable it, you just need to pass a reference to the buttons using
...
+In case there are none, Modal itself is focused.
+
+For most use cases, to make Modal accessible to all users, it should not be possible
+to move focus on elements outside of it.
 
-Autofocus is enabled by default, so if you want to control the focus of
-elements manually, set the `autoFocus` prop on Modal to `false`.
+This behavior is enabled by default and works only if autoFocus is enabled. We strongly
+recommend you keep it this way. If you need to move focus outside of Modal, set
+`focusEnforce` to `true`.
```
mbohal, line 845: Should it not say:

> set `focusEnforce` to `true`.

Or am I just confused?

src/lib/components/Modal/Modal.jsx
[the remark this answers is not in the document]
adamkudrna: > enabled by default

… and that's just fine. We set it to `true` by default and keep the name.

src/lib/components/Modal/Modal.jsx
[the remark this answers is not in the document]
adamkudrna: From my point of view, it's the same story as with `if (!disableFocusTrap)`, just in words instead of code. The default value should have no impact on the API itself.

I'm not going to push here, but I'd prefer to have it more straightforward, which means easier to understand and easier to maintan.

src/lib/components/Modal/Modal.jsx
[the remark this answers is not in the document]
adamkudrna: Also, from user's point of view, the API would soon become confusing if there was a mix of `enable*` and `disable*` options. My personal opinion is that authors of such APIs just react on the requests coming (like "we have it there but now we need to turn it off somewhere") and don't think much about usability in the downstream. Having the API consistent no matter the default value helps users (= library consumers) adopt and use the components.

src/lib/components/Modal/Modal.jsx
adamkudrna: > f you think the above use case is ok to leave up to the browser, I will remove the focus-visible option, because it would then work differently in chrome, edge and FF.

I agree.

src/lib/components/Modal/Modal.jsx
hubacekj: in the meantime I came across the information that `e.keyCode` is no longer supported in favour of `e.key`. I removed the numeric one. Given that, I don't see much benefit in creating a file with key name constants, since we only use them here and even if we might use them elsewhere in the future, it adds complexity for the sake of complexity and makes the code less readable. In general, I understand the benefits of having it in one place, in this case it seems to me like using a sledgehammer to crack a nut, though. Also, ’Escape’, ’Tab’, ‘Enter’ are self-explanatory enough.

src/lib/components/Modal/README.mdx
hubacekj: Looks like I'm the one who's confused, corrected.

src/lib/components/Modal/Modal.jsx
hubacekj: Should be ready. @bedrich-schindler

hubacekj:
Should be ready for hopefully the final round. @bedrich-schindler @adamkudrna

src/lib/components/Modal/Modal.jsx
adamkudrna: As long as there are no magic numbers, I am fine 🙂.

src/lib/components/Modal/Modal.jsx
```diff
@@ -16,116 +17,57 @@ const preRender = (
...
-      if (primaryButtonRef?.current != null) {
-        primaryButtonRef.current.focus();
-      }
-    }
+  useModalFocus(
+    autoFocus,
+    childrenWrapperRef,
+    closeButtonRef,
+    focusTrap,
+    primaryButtonRef,
```
adamkudrna, line 67: Maybe the refs could be listed together (without `focusTrap` among them)?

src/lib/components/Modal/README.mdx
```diff
@@ -1007,6 +1144,150 @@ independent of the page itself. This can be done in three ways using the
 off to prevent the modal from scrolling to the end immediately after being
 opened.
 
+## Prevent scrolling underneath the Modal
```
adamkudrna, line 1147: Please use Title Case (see other headings).

```suggestion
## Prevent Scrolling Underneath the Modal
```

src/lib/components/Modal/__tests__/Modal.test.jsx
adamkudrna, line 1: Wow, I've hardly ever seen UI tests as detailed as this 👏🏻. Nice.

src/lib/components/Modal/_hooks/useModalFocus.js
```diff
@@ -0,0 +1,96 @@
...
+        if (e.key === 'Escape' && closeButtonRef?.current != null) {
+          closeButtonRef.current.click();
+        }
+
+        if (
+          e.key === 'Enter'
+        && e.target.nodeName !== 'BUTTON'
+        && e.target.nodeName !== 'TEXTAREA'
+        && e.target.nodeName !== 'A'
+        && primaryButtonRef?.current != null
```
adamkudrna, line 62: The indentation looks weird 🤔 (hey ESlint!).

src/lib/components/Modal/Modal.jsx
adamkudrna, line 1: Now the code is beautiful and easy to read 👍🏻.

adamkudrna (approved):
Great job, thank you! 👏🏻

src/lib/components/Modal/Modal.jsx
```diff
@@ -16,116 +17,57 @@ const preRender = (
...
-        }
+      {...transferProps(restProps)}
+      className={classNames(
+        styles.root,
+        getSizeClassName(size, styles),
+        getPositionClassName(position, styles),
+      )}
+      onClick={(e) => {
+        e.stopPropagation();
       }}
```
adamkudrna, line 39: Is it safe? There is no way out…

src/lib/components/Modal/_hooks/useModalFocus.js
```diff
@@ -0,0 +1,96 @@
...
+      // focusable elements. In case there are none, childrenWrapperElement (Modal itself) is focused.
+
+      const childrenWrapperElement = childrenWrapperRef.current;
+
+      if (childrenWrapperElement == null) {
+        return () => {};
+      }
+
+      const childrenFocusableElements = Array.from(
+        childrenWrapperElement.querySelectorAll('button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'),
```
bedrich-schindler, line 29: Are we sure that `'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'` is final list of focusable elements? It is more question that my requirement, but what about allowing user to override this list of elements?

Something like this:

```
Modal.defaultProps = {
  autoFocusableElemenes: [
    'button',
    '[href]',
    'input',
    ...
  ]
}
```

cc @mbohal @adamkudrna @hubacekj

src/lib/components/Modal/_hooks/useModalFocus.js
bedrich-schindler: Is this of elements final? Can we double check it with others? I am not sure if for example input type=button or things like this will work. @mbohal any idea or do you think that @hubacekj listed all options?

src/lib/components/Modal/_hooks/useModalScrollPrevention.js
```diff
@@ -0,0 +1,34 @@
...
+        return () => {};
+      }
+
+      if (preventScrollUnderneath === 'default') {
+        const scrollbarWidth = Math.abs(window.innerWidth - window.document.documentElement.clientWidth);
+        const prevOverflow = window.document.body.style.overflow;
+        const prevPaddingRight = window.document.body.style.paddingRight;
+
+        window.document.body.style.overflow = 'hidden';
+        if (Number.isNaN(parseInt(prevPaddingRight, 10))) {
```
bedrich-schindler, line 16: Should be this condition documented in `[Prevent scrolling underneath the Modal](http://localhost:3000/components/modal#prevent-scrolling-underneath-the-modal)` section? What does @adamkudrna think?

bedrich-schindler (changes requested):
I would approve it as it is. But there are some opened topics to be resolved. When they are resolved, I will approve it.

bedrich-schindler (changes requested):
I would approve it as it is. But there are some opened topics to be resolved. When they are resolved, I will approve it.

src/lib/components/Modal/_hooks/useModalScrollPrevention.js
adamkudrna: We could mention the scroll prevention accounts with scrollbar if there is one.

src/lib/components/Modal/Modal.jsx
mbohal: Without it clicking anywhere int he modal closes the modal as the click propagates to the backdrop `<div>`.
I think it must stay as is.

src/lib/components/Modal/Modal.jsx
mbohal: We generally prefer alphabetic sorting in places where order does not matter (e.g. imports, object attributes, etc.). In function parameters however the order must reflect other things such as omitting parameters, backword compatibility, etc.

Therefore I think that sticking to alphabetic order here brings no benefit and I agree with @adamkudrna .

src/lib/components/Modal/README.mdx
```diff
@@ -825,21 +826,157 @@ To enable it, you just need to pass a reference to the buttons using
...
+In case there are none, Modal itself is focused.
+
+For most use cases, to make Modal accessible to all users, it should not be possible
+to move focus on elements outside of it.
 
-Autofocus is enabled by default, so if you want to control the focus of
-elements manually, set the `autoFocus` prop on Modal to `false`.
+This behavior is enabled by default and works only if `autoFocus` is enabled.
+We strongly recommend you keep it this way. If you need to move focus outside
+of Modal, set `focusTrap` to `false`.
```
mbohal, line 845: So is there a valid case for setting `focusTrap` to `false`?
If not, then why do we support it?

src/lib/components/Modal/_hooks/useModalFocus.js
mbohal: I would leave it as is.

If we come across something generally useful that we forgot we can add it to the list.

If we come across a use case when we will need to focus something special that is not worth adding to the list I would prefer passing to the `Modal` in a React `Ref` pointing to the element to be focused on `Modal` open (something like `initialFocusRef`). That is in my view more flexible.

I would not worry about it now though.

src/lib/components/Modal/_hooks/useModalFocus.js
mbohal: I can not think of anything else. In the original implementation that @hubacekj is building upon we only checked for `button`, so this is an improvement already.

I think this is OK once the indentation is fixed.

src/lib/components/Modal/_hooks/useModalScrollPrevention.js
mbohal: Is there a use case when the user would prefer other behavior? If not then in my view there is no need to mention it.

It is expected that these things are handled correctly and there is no need to point it out. The shorter the documentation the greater chance that someone will read it.

mbohal (approved):

src/lib/components/Modal/README.mdx
```diff
@@ -825,21 +826,157 @@ To enable it, you just need to pass a reference to the buttons using
...
+          onClick={() => setModalOpen(1)}
+        />
+        <Button
+          label="Launch modal with autofocus and focusTrap disabled"
+          onClick={() => setModalOpen(2)}
+        />
+        <Button
+          label="Launch modal with both autoFocus and focusTrap disabled"
+          onClick={() => setModalOpen(3)}
+        />
```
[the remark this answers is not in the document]
hubacekj: I intended it to behave this way. If you have `autoFocus` and `focusTrap` disabled, which means you can focus elements outside of `Modal`, you don't want your `Escape` and `Enter` keys to have overridden the default behavior. I will add the information to the docs.

src/lib/components/Modal/_hooks/useModalScrollPrevention.js
hubacekj: Well, for the other use cases there are options to turn it off or provide custom functions. Since I already mentioned each mode in the docs, I can add few words to point out that we account for scrollbar width. If someone wants to know how (and if it is the right behaviour for his use case), he can look into it in the code.

hubacekj:
@bedrich-schindler should be ready for approval

src/lib/components/Modal/README.mdx
mbohal: > of a chat window or some other widget type elements

I'm not sure if our `Modal` with automatic positioning in the center of the screen and with an opaque backdrop is the correct component. 

I think I would like to hear the arguments supporting this before merge if possible. Could we please postpone merge until out Monday call?

src/lib/components/Modal/_hooks/useModalScrollPrevention.js
mbohal: I think it not necessary to add anything to the docs, but do as you want. I have no strong feelings about it.

src/lib/components/Modal/README.mdx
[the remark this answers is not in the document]
mbohal: I quite dislike this. This is unexpected behavior that can lead to hard to find errors in the UI.

src/lib/components/Modal/README.mdx
hubacekj: Ok, i will add it as a topic on FEK board.

src/lib/components/Modal/README.mdx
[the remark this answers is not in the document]
adamkudrna: This is going to change with the `<dialog>` element anyway: the `<dialog>` (the first interactive element in it) [always gets focus](https://web.dev/learn/html/dialog/).

src/lib/components/Modal/README.mdx
adamkudrna: Good point @mbohal. While dialogs (and the `<dialog>` element, once we implement it) can be opened in both modal and non-modal way, our use case is limited to the former. Given the component name is `Modal`, we are providing a _modal_ dialog which means the rest of the page is expected to be inert and non-interactive. Then the `focusTrap` option shouldn't be necessary.

src/lib/components/Modal/README.mdx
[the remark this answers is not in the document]
hubacekj: Resolved (see final comment)

hubacekj:
Modal now always traps focus. Based on the recent discussion, we leave the option to disable autoFocus for cases in which user doesn't want the long content inside Modal to automatically scroll to the autoFocused element.

bedrich-schindler:
I tested it and it works as expected. I checked code just slightly as I have already seen it. This PR was discussed primarily with @adamkudrna and @mbohal , so I expect them to check it detaily

adamkudrna (approved):
Cool! I only tweaked a last little detail regarding scrolling and autofocus, and it's ✅.

PR #442, opened 2023-01-09
Fix unwanted white space in the upper part of horizontal `ScrollView`

adamkudrna:


mbohal (approved):
I did not test it extensively, but  it seems OK.

bedrich-schindler (approved):

PR #443, opened 2023-01-10
Create `InputGroup` component (#430)

### LINKED ISSUE #430: create inputgroup component
We need a new layout component `InputGroup`.

It will accept as props:
* `children`: an array  of elements
  * one of (defined in docs only, we can not validate their type):
    * `Button`
    * `SelectField`
    * `TextField`
  * their label must be hidden, but present for accessibility reasons
  * some of thier props will be overloaded
    *  `size` according to the `size` prop of the `InputGroup` element
    * for `Button` the prop `priority` will either be set to `filled` as there is no one to pay for the other options or it must work out of the box with **no extra work**
    * for `SelectField` and `TextField` the prop `variant` will either be set to `outline` as there is no one to pay for the other options or it must work out of the box with **no extra work**
* `validationTexts`: An array of validation messages. These are not semantically tied to the `children` elements, the connection should be expressed in textual form in the actual message. The individual `children` elements must not show any `validationText`, they only show their respective `validationState`.
* `size`: Defines the size of the `children` elements. If the element specifies a size it will be overloaded by the `InputGroup` prop value.
* `layout`: Indicates if it is `horizontal` or `vertical`

Design:

[Inputgroup.pdf](https://github.com/react-ui-org/react-ui/files/9909041/Inputgroup.pdf)

dacerondrej:
Any suggestions, especially regarding CSS and documentation, are warmly welcome. @adamkudrna 

Build failed due to asset size limit. I will not change the limit for now as it is solved in simultaneous pull requests which will be probably merged before this one.

Closes #430

### DIFF
--- a/src/lib/components/Button/Button.jsx
+++ b/src/lib/components/Button/Button.jsx
@@ -7,6 +7,7 @@ import { getRootSizeClassName } from '../_helpers/getRootSizeClassName';
 import { resolveContextOrProp } from '../_helpers/resolveContextOrProp';
 import { transferProps } from '../_helpers/transferProps';
 import { ButtonGroupContext } from '../ButtonGroup';
+import { InputGroupContext } from '../InputGroup/InputGroupContext';
 import getRootLabelVisibilityClassName from './helpers/getRootLabelVisibilityClassName';
 import getRootPriorityClassName from './helpers/getRootPriorityClassName';
 import styles from './Button.scss';
@@ -28,8 +29,14 @@ export const Button = React.forwardRef((props, ref) => {
     color,
     ...restProps
   } = props;
+  const buttonGroupContext = useContext(ButtonGroupContext);
+  const inputGroupContext = useContext(InputGroupContext);
 
-  const context = useContext(ButtonGroupContext);
+  if (buttonGroupContext && inputGroupContext) {
+    throw new Error('Button cannot be placed both in `ButtonGroup` and `InputGroup`.');
+  }
+
+  const primaryContext = buttonGroupContext ?? inputGroupContext;
 
   return (
     /* No worries, `type` is always assigned correctly through props. */
@@ -39,20 +46,21 @@ export const Button = React.forwardRef((props, ref) => {
       className={classNames(
         styles.root,
         getRootPriorityClassName(
-          resolveContextOrProp(context && context.priority, priority),
+          resolveContextOrProp(buttonGroupContext && buttonGroupContext.priority, priority),
           styles,
         ),
         getRootColorClassName(color, styles),
         getRootSizeClassName(
-          resolveContextOrProp(context && context.size, size),
+          resolveContextOrProp(primaryContext && primaryContext.size, size),
           styles,
         ),
         getRootLabelVisibilityClassName(labelVisibility, styles),
-        resolveContextOrProp(context && context.block, block) && styles.isRootBlock,
-        context && styles.isRootGrouped,
+        resolveContextOrProp(buttonGroupContext && buttonGroupContext.block, block) && styles.isRootBlock,
+        buttonGroupContext && styles.isRootInButtonGroup,
+        inputGroupContext && styles.isRootInInputGroup,
         feedbackIcon && styles.hasRootFeedback,
       )}
-      disabled={resolveContextOrProp(context && context.disabled, disabled) || !!feedbackIcon}
+      disabled={resolveContextOrProp(buttonGroupContext && buttonGroupContext.disabled, disabled) || !!feedbackIcon}
       id={id}
       ref={ref}
     >
@@ -171,8 +179,8 @@ Button.propTypes = {
   /**
    * Size of the button.
    *
-   * Ignored if the component is rendered within `ButtonGroup` component
-   * as the value is inherited in such case.
+   * Ignored if the component is rendered within `ButtonGroup` or `InputGroup` component as the value is inherited in
+   * such case.
    */
   size: PropTypes.oneOf(['small', 'medium', 'large']),
   /**
--- a/src/lib/components/InputGroup/InputGroup.jsx
+++ b/src/lib/components/InputGroup/InputGroup.jsx
@@ -1,0 +1,152 @@
+import PropTypes from 'prop-types';
+import React, { useContext } from 'react';
+import { Text } from '../Text';
+import { withGlobalProps } from '../../provider';
+import { classNames } from '../../utils/classNames';
+import { getRootSizeClassName } from '../_helpers/getRootSizeClassName';
+import { getRootValidationStateClassName } from '../_helpers/getRootValidationStateClassName';
+import { isChildrenEmpty } from '../_helpers/isChildrenEmpty';
+import { resolveContextOrProp } from '../_helpers/resolveContextOrProp';
+import { transferProps } from '../_helpers/transferProps';
+import { FormLayoutContext } from '../FormLayout';
+import { InputGroupContext } from './InputGroupContext';
+import styles from './InputGroup.scss';
+
+export const InputGroup = ({
+  children,
+  id,
+  isLabelVisible,
+  label,
+  layout,
+  size,
+  validationTexts,
+  ...restProps
+}) => {
+  const formLayoutContext = useContext(FormLayoutContext);
+
+  if (isChildrenEmpty(children)) {
+    return null;
+  }
+
+  const validationState = children.reduce(
+    (state, child) => {
+      if (state === 'invalid' || (state === 'warning' && child.props.validationState === 'valid')) {
+        return state;
+      }
+      return child.props.validationState ?? state;
+    },
+    null,
+  );
+
+  return (
+    <div
+      {...transferProps(restProps)}
+      id={id}
+      className={classNames(
+        styles.root,
+        formLayoutContext && styles.isRootInFormLayout,
+        resolveContextOrProp(formLayoutContext && formLayoutContext.layout, layout) === 'horizontal'
+          ? styles.isRootLayoutHorizontal
+          : styles.isRootLayoutVertical,
+        getRootSizeClassName(size, styles),
+        getRootValidationStateClassName(validationState, styles),
+      )}
+    >
+      <div
+        className={classNames(
+          styles.label,
+          !isLabelVisible && styles.isLabelHidden,
+        )}
+        id={id && `${id}__label`}
+      >
+        {label}
+      </div>
+      <div className={styles.field}>
+        <div
+          className={styles.inputGroup}
+          id={id && `${id}__group`}
+          role="group"
+        >
+          <InputGroupContext.Provider
+            value={{
+              layout,
+              size,
+            }}
+          >
+            {children}
+          </InputGroupContext.Provider>
+        </div>
+        {validationTexts && (
+          <div
+            className={styles.validationText}
+            id={id && `${id}__validationTexts`}
+          >
+            {validationTexts.map((validationText) => (
+              <Text blockLevel key={validationText}>
+                {validationText}
+              </Text>
+            ))}
+          </div>
+        )}
+      </div>
+    </div>
+  );
+};
+
+InputGroup.defaultProps = {
+  children: null,
+  id: undefined,
+  isLabelVisible: true,
+  layout: 'vertical',
+  size: 'medium',
+  validationTexts: null,
+};
+
+InputGroup.propTypes = {
+  /**
+   * Supported elements to be grouped:
+   * * `Button`
+   * * `SelectField`
+   * * `TextField`
+   *
+   * If none are provided nothing is rendered.
+   */
+  children: PropTypes.node,
+  /**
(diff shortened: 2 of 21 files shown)


src/lib/components/InputGroup/__tests__/InputGroup.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+  label: 'Input group label',
+};
+
+describe('rendering', () => {
+  it.each([
+    ...childrenEmptyPropTest,
+    [
+      {},
+      (rootElement) => {
+        expect(within(rootElement).getByText('Input group label'));
```
mbohal, line 34: I would prefer a dedicated test for `isLabelVisible: true`. Here we test two props in one test, while elsewhere we always test one prop per one test case.

src/lib/components/InputGroup/__tests__/InputGroup.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+    ],
+    [
+      { size: 'large' },
+      (rootElement) => {
+        expect(within(rootElement).getByText('Text field').closest('label')).toHaveClass('isRootSizeLarge');
+        expect(within(rootElement).getByText('Select field').closest('label')).toHaveClass('isRootSizeLarge');
+        expect(within(rootElement).getByRole('button')).toHaveClass('isRootSizeLarge');
+      },
+    ],
+  ])('renders with props: "%s"', (testedProps, assert) => {
```
mbohal, line 76: Test for the `validationTexts` is missing

src/lib/components/InputGroup/README.mdx
```diff
@@ -0,0 +1,228 @@
...
+### Validation States
+
+Validation states visually present the result of validation of the grouped
+inputs. Input group's validation state is taken from its child inputs. You
+should always **provide validation messages for states other than valid**
+directly through `validationTexts` prop so users know what happened and what
+action they should take or what options they have. Validation messages passed
+to input elements' `validationText` prop will be ignored.
+
+<Playground>
```
mbohal, line 164: I would like to provide en example with multiple `validationTexts`.

Also, as specified in the issue, the validation texts are not semantically tied to the input elements, the connection should be expressed in textual form in the actual message.

Could you please update the example to better illustarte this?

src/lib/components/Button/Button.jsx
```diff
@@ -28,8 +29,9 @@ export const Button = React.forwardRef((props, ref) => {
     color,
     ...restProps
   } = props;
-
-  const context = useContext(ButtonGroupContext);
+  const inputGroupContext = useContext(InputGroupContext);
+  const buttonGroupContext = useContext(ButtonGroupContext);
+  const primaryContext = buttonGroupContext ?? inputGroupContext;
```
mbohal, line 34: I think we do not want to support including a button in both `inputGroup` and `buttonGroup` If that happens I would prefer to end with an error. 

I like a nice `throw` better then producing something that might differ from what the user expects.
What do others think?

mbohal (changes requested):
Nicely done!

src/lib/components/TextField/TextField.scss
```diff
@@ -100,3 +100,17 @@
 .isRootSizeLarge {
     @include box-field-sizes.size(large);
 }
+
+// Groups
+.isRootGrouped {
```
bedrich-schindler, line 105: Can @adamkudrna check this? Don't we need to have those values specified by CSS variables? At least the gap?

src/lib/components/InputGroup/InputGroup.jsx
```diff
@@ -0,0 +1,133 @@
...
+import { classNames } from '../../utils/classNames';
+import { getRootValidationStateClassName } from '../_helpers/getRootValidationStateClassName';
+import { isChildrenEmpty } from '../_helpers/isChildrenEmpty';
+import { resolveContextOrProp } from '../_helpers/resolveContextOrProp';
+import { transferProps } from '../_helpers/transferProps';
+import { FormLayoutContext } from '../FormLayout';
+import { InputGroupContext } from './InputGroupContext';
+import styles from './InputGroup.scss';
+
+export const InputGroup = ({
```
bedrich-schindler, line 14: image

Please fix alignment of labels to be horizontally centered.

bedrich-schindler, line 14: image

@adamkudrna @adamkudrna Invalid state is always preferred one. Warning is preffered that success. It is ok to me. But what about colors of validation texts? If one of them is warning and one of them is danger, it always shows red color. Is it ok or do we want to have multiple colors?

bedrich-schindler (changes requested):

src/lib/components/InputGroup/README.mdx
```diff
@@ -0,0 +1,253 @@
...
+  <InputGroup
+    label="Vertical layout"
+    layout="vertical"
+  >
+    <TextField label="Label" />
+    <Button label="Submit" />
+  </InputGroup>
+</Playground>
```
adamkudrna, line 151: Do we actually need this type of layout?

src/lib/components/InputGroup/README.mdx
```diff
@@ -0,0 +1,253 @@
...
+  <InputGroup
+    label="Large size"
+    size="large"
+  >
+    <TextField label="Input" />
+    <Button label="Submit" />
+  </InputGroup>
+</Playground>
+
+## Invisible Label
```
adamkudrna, line 110: I think the labels of children inputs can be forced to be hidden at all times, without any chance to override it on the input level.

adamkudrna:
Please do a rebase against the latest `master`, thank you!

src/lib/components/Button/Button.jsx
dacerondrej: `buttonGroupContext && inputGroupContext` should be sufficient. E.g., if `Button` is used inside the `InputGroup` component, the value of `inputGroupContext` is `{layout: 'vertical', size: 'medium'}`, while `inputGroupContext?.current` is `undefined`.

src/lib/components/InputGroup/README.mdx
dacerondrej: Thanks to the `inputGroupContext` the labels of children inputs are always hidden. This "Invisible Label" part refers to the group label. Or did I misunderstand your comment?

src/lib/components/InputGroup/README.mdx
dacerondrej: I just followed the pdf file attached to the [issue](https://github.com/orgs/react-ui-org/projects/2?pane=issue&itemId=13446248) where both layouts are present. Should I change it?

src/lib/components/SelectField/SelectField.scss
```diff
@@ -102,3 +102,17 @@
 .isRootSizeLarge {
     @include box-field-sizes.size(large);
 }
+
+// Groups
+.isRootGrouped {
```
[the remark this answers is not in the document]
dacerondrej: I moved the code to `_box-field-layout.scss` file. Please check if it is ok.

src/lib/components/Button/Button.jsx
```diff
@@ -39,20 +46,20 @@ export const Button = React.forwardRef((props, ref) => {
...
         getRootSizeClassName(
-          resolveContextOrProp(context && context.size, size),
+          resolveContextOrProp(primaryContext && primaryContext.size, size),
           styles,
         ),
         getRootLabelVisibilityClassName(labelVisibility, styles),
-        resolveContextOrProp(context && context.block, block) && styles.isRootBlock,
-        context && styles.isRootGrouped,
+        resolveContextOrProp(buttonGroupContext && buttonGroupContext.block, block) && styles.isRootBlock,
+        primaryContext && styles.isRootGrouped,
```
adamkudrna, line 59: Here we need to distinguish between `ButtonGroup` context and `InputGroup` context. Current styling looks like it doesn't matter, but things would break with custom `ButtonGroup` theming, e.g. with a gap between buttons or with a different styling of the separator. So there will be two classes:

- `.isRootGrouped` becomes something like `.isRootInButtonGroup`,
- new class `.isRootInInputGroup` will be used for the `InputGroup` context.

adamkudrna (changes requested):
When I was implementing the support of custom inner border radius, I found a few more details that need to be changed. Will you please have a look at it? 🙏🏻

src/lib/components/InputGroup/README.mdx
adamkudrna: OK, thanks for the explanation.

src/lib/components/InputGroup/README.mdx
adamkudrna: Now it looks alright! If I'm not mistaken, the inputs were stacked on each other which led me to the question.

adamkudrna:
Theming of inner border radius: #454.

src/lib/components/InputGroup/InputGroup.jsx
adamkudrna: I think this can be closed. The `InputGroup` can have its own validation state other than the fields inside and, at the same time, the individual fields _cannot_. So it's correct if the color of the validation texts follows the state of the group.

src/lib/components/TextField/TextField.scss
adamkudrna: This is being implemented in #454 (and #455).

src/lib/components/InputGroup/InputGroup.jsx
```diff
@@ -0,0 +1,133 @@
...
+   */
+  isLabelVisible: PropTypes.bool,
+  /**
+   * Input group label.
+   */
+  label: PropTypes.string.isRequired,
+  /**
+   * Layout of the group.
+   *
+   * Ignored if the component is rendered within `FormLayout` component
```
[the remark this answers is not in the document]
adamkudrna: Horizontal form layout is still broken:

Snímek obrazovky 2023-04-26 v 15 21 44

Adding a CSS class `.isRootInFormLayout` (see other form fields) should fix it.

src/lib/components/SelectField/SelectField.jsx
```diff
@@ -51,7 +57,7 @@ export const SelectField = React.forwardRef((props, ref) => {
       <div
         className={classNames(
           styles.label,
-          !isLabelVisible && styles.isLabelHidden,
+          (!isLabelVisible || inputGroupContext) && styles.isLabelHidden,
```
adamkudrna, line 60: Please document these conditions in the component API, e.g. for label visibility:

`Automatically set to 'false' when the component is rendered within InputGroup.` (⬅️ please use backticks instead of single quotes)

And for validation text:

`Validation text is never rendered when the component is placed into an InputGroup. Instead, the InputGroup itself renders all validation texts of its nested components.`

Or something like that…

Please, document it in **all affected components**, not only `SelectField`. Thank you!

adamkudrna (approved):

src/lib/components/InputGroup/__tests__/InputGroup.test.jsx
```diff
@@ -0,0 +1,108 @@
...
+    />,
+    <Button key="button" label="button" />,
+  ],
+  label: 'label',
+};
+
+describe('rendering', () => {
+  it.each([
+    ...childrenEmptyPropTest,
+    [
```
mbohal, line 33: Do we need this test case. I think it is covered by the test case bellow. If we confirm, that the element has a class, we also know it exists.

src/lib/components/InputGroup/README.mdx
mbohal, line 1: @adamkudrna I'm confused about the relationship between `ButtonGroup` and `InputGroup`.

Does `ButtonGroup` allow us to do anything that `InputGroup` does not? Is there a reason why not to remove the `ButtonGroup` component?

If such reason exists, we should put it in docs.

image
Screenshot from 2023-05-02 10-53-25

bedrich-schindler (approved):

src/lib/components/InputGroup/README.mdx
adamkudrna: Well, the needs are different, and I'm not only talking about design. `ButtonGroup` is optimized for controlling `Button`s while `InputGroup` is good for controlling form fields. And you don't want a group of `Button`s to have a validation message, do you 🙂.

package-lock.json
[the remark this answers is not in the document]
adamkudrna: It's the version number:

obrazek

Looks like there may be a new place to update the version number during release. I'll check it with the next release.

src/lib/components/InputGroup/README.mdx
adamkudrna: obrazek

adamkudrna (approved):
Now I believe it's all done 🙂.

package-lock.json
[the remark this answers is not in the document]
mbohal: > 'll check it with the next release.
Or @adamkudrna can just use https://docs.npmjs.com/cli/v8/commands/npm-version

PR #450, opened 2023-02-17
Improve the default theming of `secondary` and `selected` colors of `Button`

adamkudrna:
obrazek
obrazek
obrazek
obrazek

### DIFF
--- a/src/lib/components/ButtonGroup/README.mdx
+++ b/src/lib/components/ButtonGroup/README.mdx
@@ -43,6 +43,13 @@ See [API](#api) for all available options.
 - Use button group to group **related actions** that a user can take. Buttons
   should not be grouped just to save space on the screen.
 
+- **Mixing multiple button priorities** and some other properties within a
+  button group is [not allowed](#shared-properties). The priority can only be
+  set for all buttons in the group at once.
+
+- In most use cases, **secondary action color** is probably the best option for
+  buttons in a group as it works good with the colors of the selected state.
+
 - Use **short labels or icons** so the buttons can fit small screens.
 
 - For toggling between on/off states, use rather the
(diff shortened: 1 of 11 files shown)


src/lib/components/ButtonGroup/README.mdx
```diff
@@ -43,6 +43,12 @@ See [API](#api) for all available options.
 - Use button group to group **related actions** that a user can take. Buttons
   should not be grouped just to save space on the screen.
 
+- **Mixing multiple button priorities** within a button group is **not
```
mbohal, line 46: Priorities can not be mixed, they come from context:

[react-ui/src/lib/components/Button/Button.jsx](https://github.com/react-ui-org/react-ui/blob/296e73fd8d8525eaffeac9bc245a3125986625e2/src/lib/components/Button/Button.jsx#L42)

There is no need to warn users about it.

src/lib/components/Modal/README.mdx
```diff
@@ -478,15 +435,38 @@ There are two ways to position buttons within the ModalFooter:
...
-                  <p>
-                    Do you really want to delete the user <code>admin</code>?
-                    This cannot be undone.
-                  </p>
+                  <Radio
+                    label="Footer justification"
+                    onChange={(e) => setModalJustify(e.target.value)}
+                    options={[
+                      {
+                        label: 'Start',
```
mbohal, line 448: I would consider adding to the  labels the actual value passed to `justify` prop of the `<ModalFooter>` component. Like this as a user I have to view source what to use.

mbohal (approved):

src/lib/components/ButtonGroup/README.mdx
adamkudrna: I'm pointing it out because I tried to do that, it didn't work and I wasn't sure if it was a bug or feature.

mbohal: OK, but then we should also mention the other enforced props like `size` and `block`.

src/lib/components/Modal/README.mdx
adamkudrna: obrazek

PR #455, opened 2023-03-01
Make inner border radius of `Button` in `ButtonGroup` configurable

adamkudrna:


adamkudrna:
Everything will be done in #454.

PR #466, opened 2023-05-17
Render `InputGroup`, `ButtonGroup`, and `Radio` as `<fieldset>` to improve controlling and accessibility

adamkudrna:
All input groups now use the [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) element.

`InputGroup` can be `disabled` now.

**⚠️ Pontential BC:** The suffix of the ID of `Radio` label element changed from `__labelText` to `__label`
(so all input groups use the same approach).


src/lib/components/Radio/README.mdx
```diff
@@ -77,6 +77,11 @@ See [API](#api) for all available options.
 - Use **clear, calm error messages** when there's a problem with what they
   entered.
 
+- In the background, InputGroup uses the [`fieldset`][fieldset] element. Not
```
mbohal, line 80: Do we want to communicate this to the user? 

Isn't it an implementation detail? Are there any cases when using div as before would be preferable from the users perspective? If not, then I don't think we should bother the user with this information.

src/lib/components/Radio/Radio.jsx
```diff
@@ -36,53 +37,59 @@ export const Radio = ({
...
           styles.label,
           !isLabelVisible && styles.isLabelHidden,
         )}
-        id={id && `${id}__labelText`}
+        id={id && `${id}__displayLabel`}
       >
         {label}
       </div>
       <div className={styles.field}>
-        <ul className={styles.list}>
```
mbohal, line 51: Why are we dropping the list based implementation?
Isn't list more semantic? 

Let me know, I'm eager for knowledge :books:

tests/propTests/isLabelVisible.js
```diff
@@ -1,12 +1,32 @@
 import { within } from '@testing-library/react';
 
-export const isLabelVisible = [
+export const isLabelVisible = (type = 'label') => (
```
mbohal, line 3: I think that `HtmlTag` or such would be more descriptive then just `type`.

tests/propTests/labelPropTest.js
```diff
@@ -1,8 +1,20 @@
 import { within } from '@testing-library/react';
 
-export const labelPropTest = [
+export const labelPropTest = (type = 'label') => (
```
mbohal, line 3: argument name, see above

mbohal (approved):

bedrich-schindler (approved):

src/lib/components/Radio/Radio.jsx
adamkudrna: [My idea is that] Once there is the `<fieldset>` element, the inputs are already tied together which makes the `<ul>` redundant.

adamkudrna: (Also, Heydon [does it](https://www.sitepoint.com/replacing-radio-buttons-without-replacing-radio-buttons/) the same way.)

src/lib/components/Radio/README.mdx
adamkudrna: > … it also allows you to make use of its built-in features like disabling all nested inputs or pairing the group with a form outside.

I don't think you want to use a `<div>` instead, but this definitely is a benefit.

PR #475, opened 2023-08-04
Require Node 18 (current LTS) (#374)

### LINKED ISSUE #374: Require Node 18 (latest LTS)
Version 18 is currently not supported because of error during `npm build` on `Error: error:0308010C:digital envelope routines::unsupported`.

However, commonly suggested workaround (passing the `--legacy-ssl-options`) is not supported by GH actions.

adamkudrna:
Closes #374.

### DIFF
--- a/docker/mkdocs/Dockerfile
+++ b/docker/mkdocs/Dockerfile
@@ -1,7 +1,5 @@
-FROM            node:16
+FROM            squidfunk/mkdocs-material:9
 
-RUN             apt-get update && apt-get install -y python3-pip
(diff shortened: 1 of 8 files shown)


docker/mkdocs/Dockerfile
```diff
@@ -1,7 +1,5 @@
-FROM            node:16
```
adamkudrna, line 1: Using a ready-to-use Docker image is way easier than trying to install MkDocs ourselves, especially on the new Ubuntu with more strict security measures for `apt-get`.

bedrich-schindler (approved):

dacerondrej (approved):

PR #476, opened 2023-08-04
Narrow down supported browsers in Browserslist config according to the docs

adamkudrna:
```sh
npx browserslist

and_chr 116
and_ff 116
chrome 116
chrome 115
edge 116
edge 115
firefox 117
firefox 116
firefox 115
firefox 102
ios_saf 16.5
ios_saf 16.4
ios_saf 16.3
ios_saf 16.2
ios_saf 16.1
ios_saf 16.0
safari 16.6
safari 16.5
safari 16.4
safari 16.3
safari 16.2
safari 16.1
safari 16.0
```

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,5 +1,15 @@
 # Browsers that we support
 
-defaults
-not ie > 1
-not ie_mob > 1
+# Blink/Chromium based
+last 1 ChromeAndroid versions
+last 2 Chrome versions
+last 2 Edge versions
+
+# Gecko based
+last 1 FirefoxAndroid versions
+last 2 Firefox versions
+Firefox ESR
+
+# WebKit based
+last 1 iOS major versions
+last 1 Safari major versions
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,9 +60,9 @@
         "npm": ">=9"
       },
       "peerDependencies": {
-        "prop-types": "^15.8.1",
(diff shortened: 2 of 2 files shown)


bedrich-schindler:
I am just curious. What is behind this change? Why don't we support two latest browsers as before? Be aware, that we must/should support the latest major versions in some of our clients ...

adamkudrna:
> I am just curious. What is behind this change? Why don't we support two latest browsers as before? Be aware, that we must/should support the latest major versions in some of our clients ...

@bedrich-schindler I did two things:

1. Changed the vague `defaults` config to an explicit white-list (allow-list 👀 ) of browser versions.
2. Desktop browsers still support last 2 major versions. I only limited mobile browsers to the most recent major version. I believe it is based on our discussion about pre-15.4 versions of mobile Safari which are still "2 major versions back", but much less capable. However, I can change the config to be more open and include at least the last 2 versions of other mobile browsers.

What do you think?

bedrich-schindler (approved):

mbohal (approved):
I would add Firefox ESR support.

adamkudrna:
> I would add Firefox ESR support.

That effectively means supporting up to one-year-old version of Firefox (version 102 as of today). I think we may be open to fixing bugs reported in ESR, but I doubt we can actively test in it…

Anyway, I'm adding ESR to the list.

PR #477, opened 2023-08-07
Check all Markdowns for broken links

adamkudrna:
There are two link checks now:

- This check is performed during every PR and only checks **absolute** URLs in all **Markdown files**.
- There is still the existing broken links check that checks **all** links at **react-ui.io**.

### DIFF
--- a/.github/markdown-link-check.json
+++ b/.github/markdown-link-check.json
@@ -1,0 +1,10 @@
+{
+  "ignorePatterns": [
+    {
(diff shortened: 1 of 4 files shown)


bedrich-schindler:
Tbh, I am unable to review this. It seems OK, but I believe you tested it.

adamkudrna:
> Tbh, I am unable to review this. It seems OK, but I believe you tested it.

I did and it seemed to be working to me 🙂.

PR #482, opened 2023-08-28
Allow the `Modal` component to block scrolling on any HTML element instead of just the `<body>` element (#472)

### LINKED ISSUE #472: `Modal` component `preventScrollUnderneath` prop not working
When migrating to MkDocs/Docoff based docs (https://github.com/react-ui-org/react-ui/pull/440) an issue with the prop `preventScrollUnderneath` in the `Modal` component appeared.

It seems that the default scroll prevention does not work with the MkDocs platform:

1. Scrolling is possible when `preventScrollUnderneath="default"`
2. In Chrome (FF is OK) there is extra padding in the top right corner when `Modal` is open:
      image

mbohal:
Closes #472.

### DIFF
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -26,7 +26,13 @@ React.createElement(() => {
   const modalPrimaryButtonRef = React.useRef();
   const modalCloseButtonRef = React.useRef();
   return (
-    <>
+    {/*
+      The `preventScrollUnderneath` feature is necessary for Modals to work in
+      React UI docs. You may not need it in your application.
+    */}
+    <RUIProvider globalProps={{
+      Modal: { preventScrollUnderneath: window.document.documentElement }
+    }}>
       <Button
         label="Launch modal"
         onClick={() => setModalOpen(true)}
@@ -66,7 +72,7 @@ React.createElement(() => {
           </Modal>
         )}
       </div>
-    </>
+    </RUIProvider>
   );
 });
 ```
@@ -116,7 +122,13 @@ React.createElement(() => {
   const modalPrimaryButtonRef = React.useRef();
   const modalCloseButtonRef = React.useRef();
   return (
-    <>
+    {/*
+      The `preventScrollUnderneath` feature is necessary for Modals to work in
+      React UI docs. You may not need it in your application.
+    */}
+    <RUIProvider globalProps={{
+      Modal: { preventScrollUnderneath: window.document.documentElement }
+    }}>
       <Button
         label="Launch blocking modal without title"
         onClick={() => {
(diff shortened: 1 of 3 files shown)


dacerondrej (approved):

hubacekj (approved):

src/components/Modal/README.md
```diff
@@ -26,7 +26,9 @@ React.createElement(() => {
   const modalPrimaryButtonRef = React.useRef();
   const modalCloseButtonRef = React.useRef();
   return (
-    <>
+    <RUIProvider globalProps={{
+      Modal: { preventScrollUnderneath: window.document.documentElement }
```
bedrich-schindler, line 30: I might miss something, but why is everywhere `window.document.documentElement`? Why not to use default?

mbohal: Because default only works in cases where the `<body>` element is scrollable. Our documentation platform (MkDocs material) however scrolls the `<html>` element.

From what @adamkudrna was saying I understand that scrolling `<body>` is more common, but it is not the only valid way. The API tweaks in this PR reflect that I think.

bedrich-schindler (approved):

src/components/Modal/README.md
```diff
@@ -26,7 +26,9 @@ React.createElement(() => {
...
+    }}>
```
adamkudrna, line 31: Can we have a comment explaining the reason in all examples? Example code snippets should be copy-paste ready (that's why I'd like to get rid of this workaround in the future, even if we have to modify the docs theme).

mbohal: > that's why I'd like to get rid of this workaround in the future, even if we have to modify the docs theme

That somehow implies that the docs theme is wrong. Is it wrong? If not then I see two options:

1.  Make the `<html>` element default. However changing defaults based on the docs platform is  is strange…
2.  Or we could make default the `off` value. That would be agnostic to the context of usage. I actaally kind of like this, because it makes no assumptions and it is expected that RUI should be tweaked before usage anyway.

adamkudrna: > That somehow implies that the docs theme is wrong. Is it wrong?

No. We just want to show examples **as simple as possible**, but with the current docs theme we are forced to showcase the more complicated version. And I still believe we chose the right default for `Modal`, i.e. to expect scrolling on `<body>`.

> it is expected that RUI should be tweaked before usage anyway

Oh, it doesn't need to be! 🙂 But that's another discussion…

According to our call today, I'm adding dropping a decent comment on it so we can move on.

adamkudrna (approved):

Issue #478, opened 2023-08-07
Upgrade React to v18 and test against v17 and v18


adamkudrna:

bedrich-schindler:
Update performed in #358

PR #483, opened 2023-09-01
Update React from version 17 to version 18 and update all development dependencies excluding `webpack-cli`, `stylelint` and `sinon` (#358)

### LINKED ISSUE #358: Upgrade dependencies
Blocks upgrade to:

- `sass-loader` 11+
- `style-loader` 3+
- `postcss-loader` 5+
- `css-loader` 6+
- `gatsby-plugin-sass` 4+

Originally pointed out in https://github.com/react-ui-org/react-ui/pull/336#pullrequestreview-806994446.

bedrich-schindler:
## BREAKING CHANGE

* Update `react` from version 17 to version 18

## Migration

Upgrade to React 18.

## Other Changes

*  Update all development dependencies excluding `webpack-cli`, `stylelint` and `sinon`, fix failing tests caused by breaking changes in testing libraries (#358, #478)
* Remove `sinon` and replace `sinon.spy()` with `jest.fn()` (#358)
* Update `webpack-cli` (#358)
* [Update documentation to fix markdownlint errors (#358)

Closes #358

### DIFF
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -98,14 +98,14 @@ See [API](#api) for all available options.
 
 Modal is decomposed into the following components:
 
-- [Modal](#api)
-  - [ModalHeader](#modalheader_1)
-    - [ModalTitle](#modaltitle)
-    - [ModalCloseButton](#modalclosebutton)
-  - [ModalBody](#modalbody_1)
-    - [ModalContent](#modalcontent)
+- Modal
+  - [ModalHeader](#modalheader)
+    - ModalTitle
+    - ModalCloseButton
+  - [ModalBody](#modalbody)
+    - ModalContent
       (may be wrapped with [ScrollView](/components/ScrollView))
-  - [ModalFooter](#modalfooter_1)
+  - [ModalFooter](#modalfooter)
 
 Using different combinations, you can compose different kinds of modals,
 e.g. dialog modal, blocking modal, scrollable modal, etc.
--- a/src/components/Card/README.md
+++ b/src/components/Card/README.md
@@ -269,7 +269,7 @@ its accessibility.
 
 ### CardBody
 
-Space your content with CardBody. See [Card Composition](#card-composition) for
+Space your content with CardBody. See [Composition](#composition) for
 all details.
 
 <docoff-react-props src="/components/Card/CardBody.jsx"></docoff-react-props>
@@ -277,7 +277,7 @@ all details.
 ### CardFooter
 
 Separate your card actions with CardFooter. See
-[Card Composition](#card-composition) for all details.
+[Composition](#composition) for all details.
 
 <docoff-react-props src="/components/Card/CardFooter.jsx"></docoff-react-props>
 
--- a/src/components/FormLayout/README.md
+++ b/src/components/FormLayout/README.md
@@ -187,7 +187,7 @@ work together yet 🎩.
 ## Custom Fields
 
 You can even place any content you need into the FormLayout — just wrap it with
-the [FormLayoutCustomField](#formlayoutcustomfield) component. This layout
+the FormLayoutCustomField component. This layout
 helper ensures your content is properly spaced and aligned with to other
 FormLayout elements. FormLayoutCustomFields are designed to work solely inside
 the FormLayout component.
--- a/src/components/Toolbar/README.md
+++ b/src/components/Toolbar/README.md
@@ -353,10 +353,7 @@ A wrapper for individual toolbar items.
 | `--rui-Toolbar__gap`                                 | Gap between toolbar items                                    |
 | `--rui-Toolbar__gap--dense`                          | Dense gap between toolbar items                              |
 
-[inline-elements]: https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements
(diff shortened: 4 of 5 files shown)


src/components/Modal/README.md
bedrich-schindler, line 1: @adamkudrna @mbohal There is a problem with headings I wasn't able to solve. Even if I change headings under API section from `ModalHeader -> ModalHeader API`, so that `#modalheader_1` can be changed `#modalheader`, it still does not work. So I left it as it was and just ignored those specific case.

If you know how to fix it, do it, please.

mbohal: > Even if I change headings under API section from ModalHeader -> ModalHeader API, so that #modalheader_1 can be changed #modalheader, it still does not work.

It seems to me that changing `ModalHeader -> ModalHeader API` and `#modalheader_1 -> #modalheader-api` works.

However I'm not sure if linking from here to the API section is all that useful. When I encounter this section, I would prefer to see links to the `Composition -> <component>` section rather then to the `API  -> <component>` section. For components that do not have a textual description I think it is fine to have them as text (not link). The reason is, that when I'm reading about composition I'm not  that interested in API, that comes later.

I leave it up to @adamkudrna , but I would prefer solution that does not disable lint rules.

mbohal (approved):

src/components/Modal/README.md
mbohal: >  Even if I change headings under API section from ModalHeader -> ModalHeader API, so that #modalheader_1 can be changed #modalheader, it still does not work.

It seems to me that changing `ModalHeader` -> `ModalHeader API` and `#modalheader_1` -> `#modalheader-api` works.

However I'm not sure if linking from here to the API section is all that useful. When I encounter this section, I would prefer to see links to the `Composition -> <component>` section rather then to the `API -> <component>` section. For components that do not have a textual description I think it is fine to have them as text (not link). The reason is, that when I'm reading about composition I'm not that interested in API, that comes later.

I leave it up to @adamkudrna , but I would prefer solution that does not disable lint rules.

bedrich-schindler: Nonono, there are duplicate headings. So I tried to change `ModalHeader` (under `API` section`) to `ModalHeader API`, so I can change `#modalheader_1` to `#modalheader`, but it does not solve the linter error. And I do not know why ...

bedrich-schindler: We agreed with @mbohal to remove links completely. It solves both problems

src/components/Modal/README.md
```diff
@@ -98,14 +98,14 @@ See [API](#api) for all available options.
...
-    - [ModalContent](#modalcontent)
+- Modal
+  - ModalHeader
+    - ModalTitle
+    - ModalCloseButton
+  - ModalBody
+    - ModalContent
       (may be wrapped with [ScrollView](/components/ScrollView))
-  - [ModalFooter](#modalfooter_1)
+  - ModalFooter
```
adamkudrna, line 108: Looking at it, I'd like to bring the links back — at least `ModalHeader`, `ModalBody` and `ModalFooter` make sense as they have their own sections.

obrazek

(Broken list indentation is another issue…)

adamkudrna (changes requested):
Cool! Just a few details in the docs, after all…

src/components/Modal/README.md
dacerondrej: I came across another two occurrences of broken list indentation:

- On the `Card` component page, in the 'Composition' section
image
- On the `Button` component page, in the 'Theming Priorities and Color Variants' section
image

package.json
```diff
@@ -50,53 +50,52 @@
     "normalize.css": "^8.0.1"
   },
   "peerDependencies": {
-    "prop-types": "^15.7.2",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "prop-types": "^15.8.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
```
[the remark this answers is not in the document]
mbohal: Major versions are probably fine technically, but I prefer to see here the last version actually included and presumably tested by the developer.

It is easier to find it here then to look through git history of the `package-lock.json` file.

src/components/Modal/README.md
adamkudrna: I created #486 to fix it.

package.json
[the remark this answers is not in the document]
bedrich-schindler: @adamkudrna has a point. We should look at the peer deps from different perspective. We do no have any reason to strictly require latest minor version while major version is important for us. Unless there are incompatible changes, I will required ^18 and so on.

src/components/Modal/README.md
bedrich-schindler: It does make sense to us and @mbohal to have links to subsections that are just below the defined links. What more, I have problems with linter which was the primary reason to those changes. I will try it again.

package.json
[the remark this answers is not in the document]
mbohal: I agree with @bedrich-schindler I did not realize we are talking about peer deps here.

bedrich-schindler:
I merged this, @adamkudrna is gonna check failing coverage check.

PR #490, opened 2023-09-18
Rename CSS and JS endpoints to use hyphenation

adamkudrna:
- Use hyphenation in `react-ui` because we use it everywhere.
- Rename `cssCustomProperties` endpoint to `docsCustomProperties` so it's clear what is it for.


mbohal (approved):
I did not test it, but it looks OK.

bedrich-schindler (approved):

dacerondrej (approved):

PR #502, opened 2023-10-17
Move `transferProps` back from `fieldset` to `input` in `RadioField` (#481)

### LINKED ISSUE #481: Move the `onChange` handler back to `Radio` inputs
Currently, the `onChange` event handler is tied onto the wrapping `<fieldset>` which causes React to throw the following warning:

```md
Warning: You provided a `checked` prop to a form field without an `onChange` handler.
This will render a read-only field. If the field should be mutable use `defaultChecked`.
Otherwise, set either `onChange` or `readOnly`.
```

The `onChange` handler needs to be transferred onto individual radio inputs as it was before https://github.com/react-ui-org/react-ui/pull/466.

Don't forget to make the docs more obvious: https://react-ui.io/components/radio#forwarding-html-attributes

bedrich-schindler:
Closes #481

### DIFF
--- a/src/components/Radio/README.md
+++ b/src/components/Radio/README.md
@@ -281,8 +281,8 @@ It's possible to disable just some options or the whole set.
 In addition to the options below in the [component's API](#api) section, you
 can specify [React synthetic events] or you can **add whatever HTML attribute
 you like.** All attributes that don't interfere with the API are forwarded to
-the native HTML `<input>`. This enables making the component interactive and helps
-to improve its accessibility.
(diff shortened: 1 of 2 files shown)


mbohal (approved):
I would prefer to also update the docs. Change:

> All attributes that don't interfere with the API are forwarded to the native HTML <input>

to:

> All attributes that don't interfere with the API are forwarded to the native HTML <input> elements

To make it clear, that props are transfered to all of the nested `<inputs>`s.

adamkudrna (approved):

PR #512, opened 2023-11-29
Update markdownlint rules

mbohal:
* rules were sorted
* `MD013` is used instead of `line-length` because we use codes everywhere else
* `MD024` was titghtened to allow duplicate headings only under different parents


bedrich-schindler (approved):

mbohal:
I fixed both typos in the commit message and I also rebased this PR on top of master.

adamkudrna (approved):
Thanks! 👍🏻

PR #519, opened 2024-01-10
Omit link to legacy list of React synthetic events in component documentation

mbohal:
The page listing react synthetic events became legacy and was moved to https://legacy.reactjs.org/docs/events.html

I could not find a direct replacement, the closes thing I found is https://react.dev/reference/react-dom/components/common#common-props. This page however is not HTML element specific and thus I think it better to link do MDN even though there some differences (e.g. `class` vs `className`)

Also I have updated transferProps based on the new doc page.

### DIFF
--- a/src/components/Alert/README.md
+++ b/src/components/Alert/README.md
@@ -171,13 +171,15 @@ React.createElement(() => {
 ## Forwarding HTML Attributes
 
 In addition to the options below in the [component's API](#api) section, you
-can specify [React synthetic events] or **any HTML attribute you like.** All
-attributes that don't interfere with the API are forwarded to the root `<div>`
-HTML element. This enables making the component interactive and helps to improve
+can specify **any HTML attribute you like.** All attributes that don't interfere
+with the API of the React component are forwarded to the root `<div>` HTML
+element. This enables making the component interactive and helps to improve
 its accessibility.
 
-👉 Refer to the MDN reference for the full list of supported attributes of the
-[div] element.
+👉 For the full list of supported attributes refer to:
+
+- [`<div>` HTML element attributes][div-attributes]{:target="_blank"}
+- [React common props]{:target="_blank"}
 
 ## API
 
@@ -211,5 +213,5 @@ Where:
 - `<PROPERTY>` is one of `color` (color of text), `foreground-color` (color of
   border, icon, links, and emphasis), or `background-color`.
 
-[React synthetic events]: https://reactjs.org/docs/events.html
-[div]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#attributes
+[div-attributes]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#attributes
+[React common props]: https://react.dev/reference/react-dom/components/common#common-props
--- a/src/components/Tabs/README.md
+++ b/src/components/Tabs/README.md
@@ -157,14 +157,20 @@ React.createElement(() => {
 ## Forwarding HTML Attributes
 
 In addition to the options below in the [component's API](#api) section, you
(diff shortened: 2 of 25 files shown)


src/components/Alert/README.md
mbohal, line 1: This is just an example. If we agree on this change I will copy it over to all the other components.

adamkudrna:
> I could not find a direct replacement, the closes thing I found is https://react.dev/reference/react-dom/components/common#common-props. This page however is not HTML element specific and thus I think it better to link do MDN even though there some differences (e.g. class vs className)

It's not just the `className`. There are dozens of attributes (or props and events) that are React specific and cannot be found in HTML spec. I think this page ⬆️ is actually a very good replacement.

Also I'm unsure if we can omit such a basic thing as the `onClick` prop. So, I'd prefer a replacement rather than just deleting it without saying.

What does @bedrich-schindler think?

> Also I have updated transferProps based on the new doc page.

I could not find this in the PR…

mbohal:
@adamkudrna I chnaged the wording to include a reference to both the MDN `div` attribute list and React common prop page as agreed on our video call.

src/components/Alert/README.md
```diff
@@ -171,13 +171,13 @@ React.createElement(() => {
...
-attributes that don't interfere with the API are forwarded to the root `<div>`
-HTML element. This enables making the component interactive and helps to improve
+can specify **any HTML attribute you like.** All attributes that don't interfere
+with the API of the React component are forwarded to the root `<div>` HTML
+element. This enables making the component interactive and helps to improve
 its accessibility.
 
-👉 Refer to the MDN reference for the full list of supported attributes of the
-[div] element.
+👉 For the full list of supported attributes refer to [div HTML element attributes]
```
adamkudrna, line 179: I am aware it wasn't there before, but I wish we could polish the formatting like this:

```suggestion
👉 For the full list of supported attributes refer to the [`div` HTML element attributes]
```

Does it work?

adamkudrna (approved):
Perfect! This is much more helpful. 👏🏻

src/components/Alert/README.md
mbohal: > Does it work?

No, I have tested it bedore.

At least I made the links open in new tab.
@bedrich-schindler @adamkudrna can I copy this change to other README files?

bedrich-schindler (approved):

src/components/Alert/README.md
mbohal: On fact, there is a way. I did change it as suggested.

mbohal:
In the end I arranged links to resources as a list because:

1. It improves readability in both `*.md` and HTML versions
2. It allows for uniform link labels without sacrificing readability

adamkudrna (approved):
Beautiful job, thank you! 👏🏻

src/components/Tabs/README.md
```diff
@@ -157,14 +157,20 @@ React.createElement(() => {
...
-element in case of `Tabs` component and to the `<a>` HTML element in
-case of `TabsItem`. This enables making the component interactive and helps
-to improve its accessibility.
+can specify **any HTML attribute you like.** All attributes that don't interfere
+with the API of the React component are forwarded to:
 
-👉 Refer to the MDN reference for the full list of supported attributes of the
-[nav] and [li] element.
+- the root `<nav>` HTML element in case of `Tabs` component
+- the `<a>` HTML element in case of `TabsItem`.
```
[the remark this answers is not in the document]
mbohal: I went the no-punctuation way. Mixing sentences and bullet points is to wild for the old school self that I am.

src/components/Alert/README.md
mbohal: done

PR #520, opened 2024-01-10
Update blacklisted props that can be passed to native HTML elements using the transferProps principle

mbohal:


src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,10 @@
 export const transferProps = ({
   children,
   className,
+  dangerouslySetInnerHTML,
   ref,
+  suppressContentEditableWarning,
```
adamkudrna, line 6: Why this one? The others seem reasonable to me (i.e. we use them ourselves so passing them would break things).

Maybe the items here deserve a comment (if not a MD documentation) so it doesn't fell like a random list.

mbohal: > Why this one? 

Because I think we do not want to allow users to make a HTML element within a component to have `contenteditable` attribute set if it has child nodes . There would be to many edge cases. Honestly I do not know what is there to document. By listing it is obvious that we do not allow this.

I added some docs, but I'm not sure that is there to write about the individual blacklisted props.

I think we do not want to blacklist `contenteditable` completely, because I can imagine allowing it for some components (e.g. `Badge`). Not that it is good UX, but technically there is no problem with it and UX decisions should take place on the concrete UI library based on RUI.

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,44 @@
...
+    className,
+    dangerouslySetInnerHTML,
+    ref,
+    staticContext,
+    style,
+    suppressContentEditableWarning,
+    ...restProps
+  } = props;
+
+  if (process.env.NODE_ENV !== 'production') {
```
mbohal, line 25: This trick is describe here: https://overreacted.io/how-does-the-development-mode-work/.

I tested it and it works as expected.

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,44 @@
...
-  className,
-  ref,
-  staticContext,
-  ...restProps
-}) => restProps;
+/**
+ * Controls passing of props from the React component to the HTML element
+ *
+ * Sometimes it is useful to have a mechanism to pass props from the React component to a rendered HTML element.
+ * It enables making the component interactive and it helps to improve its accessibility. However some props should
```
adamkudrna, line 5:
```suggestion
 * It enables making the component interactive and helps improve its accessibility. However some props should
```

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,44 @@
...
+ * never be passed to the HTML element as it would break things. This function is used to filter out such props.
+ *
+ * When run in development mode the function will log error to the console if any invalid props are passed.
```
adamkudrna, line 8:
```suggestion
 * When run in development mode, the function will log the error to the console if any invalid props are passed.
```

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,44 @@
...
+ *
+ * @param props The props that were passed tot he React component
```
adamkudrna, line 10:
```suggestion
 * @param props The props that were passed to the React component
```

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,44 @@
...
+      ref,
+      staticContext,
+      style,
+      suppressContentEditableWarning,
+    })
+      .filter((key) => props[key] !== undefined);
+
+    if (invalidProps.length > 0) {
+      // eslint-disable-next-line no-console
+      console.error(`Invalid prop(s) supplied to the transferProps method: ${invalidProps.join(', ')}`);
```
adamkudrna, line 39: > supplied to the transferProps method

Just a DX thing. This is technically correct, but will developers know what the `transferProps()` method is? (Isn't it actually a _function_ BTW?) Can we guide the developers to the place where they can fix the error?

```suggestion
      console.error(`Invalid prop(s) supplied to component: ${invalidProps.join(', ')}`);
```

adamkudrna (commented):
Thank you for the docs! 👍🏻

src/components/_helpers/transferProps.js
mbohal: I do not know how to get the caller function. the best I could find was https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/caller and I do not think we want that.

I improved readability by adding some quotes though.

src/components/_helpers/transferProps.js
mbohal: I also blacklisted the `contentEditable` attribute

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,46 @@
...
+    dangerouslySetInnerHTML,
+    ref,
+    staticContext,
+    style,
+    suppressContentEditableWarning,
+    ...restProps
+  } = props;
+
+  if (process.env.NODE_ENV !== 'production') {
+    const invalidProps = Object.keys({
```
bedrich-schindler, line 27: I would rather keep this as array of strings due to performance. And this array should be as constant before the function to prevent inicialization on every function call.

bedrich-schindler (changes requested):

src/components/_helpers/transferProps.js
mbohal: I changed the implementation.

Since the code only runs in dev mode I don't think it makes much of a difference, but it for sure does not hurt.

adamkudrna (commented):
I'd still appreciate a comment—or documentation—for the list of the ~~black~~deny-listed props. Otherwise seems OK to me.

adamkudrna:
One more thing: please clean up the commits a bit, not all of them are fixups. Thank you!

obrazek

mbohal:
@adamkudrna wrote:

> I'd still appreciate a comment—or documentation

I would prefer to deal with documentation once we decide if we want to make the `transferProps` function public. I hope we will make it public and also publicly documented.

As per comment I dont know what to write. The concept is describe already. the individual blacklisted props are what they are. I can not come up with anything other then banal sentences like `We blacklist "className" prop to prevent setting custom classes.`.

Could you please either write the comments yourself or at least make an example? I'm obviously not really  able to do it myself.

src/components/_helpers/transferProps.js
```diff
@@ -1,7 +1,46 @@
...
+  if (process.env.NODE_ENV !== 'production') {
+    const invalidProps = [
+      'children',
+      'className',
+      'contentEditable',
+      'dangerouslySetInnerHTML',
+      'ref',
+      'staticContext',
+      'style',
+      'suppressContentEditableWarning',
```
adamkudrna, line 35: OK, here we go:

```suggestion
      'children', // We only read children from between the component tags, e.g. `<Component>children</Component>`.
      'className', // We want to control the styling ourselves and not allow anyone else to change it.
      'contentEditable', // ???
      'dangerouslySetInnerHTML', // We don't want the HTML of our components to be changed from outside.
      'ref', // ???
      'staticContext', // ???
      'style', // We don't want to allow manual styling of our components. Also, we use it to set custom properties in some places.
      'suppressContentEditableWarning', // ???
```

I'm just unable to put down the reasons for all of the items. This is all I'm aware of.

mbohal: done

mbohal:
It is rebased and ready for CR.

adamkudrna (approved):
👍🏻

dacerondrej (approved):

bedrich-schindler (approved):

PR #529, opened 2024-03-06
Leverage CSS cascade layers to fix and user-proof styles order

### LINKED ISSUE #518: Leverage CSS cascade layers to fix and user-proof styles order
This is a follow-up issue of #517.

As a result of #440, component styles are no longer part of the JS bundle. They reside in the extracted `react-ui[.development].css` file. The order of styles in this stylesheet is incorrect and **may cause styling issues**.

The reason of the incorrect order is in the call order in `src/index.js`: first, there are imports (theme, foundation,  ⚠️ **helpers** ⚠️), then exports (components).

 Cascade layers will help keep the styles in correct order, no matter how they are called.

adamkudrna:
As long as `layers.scss` (layers order definition) is called first, our Sass styles can be imported in any order.

It solves the incorrect (but unavoidable, due to the way we compile CSS) call order of styles in our `index.js`.

Also, developers can now extend the CSS cascade instead of just prepending or appending their styles to ours.

For example:

```css
/* My custom app styles */
@layer foundation.elements {
  code {
    background-color: silver;
  }
}
```

Closes #518.


bedrich-schindler:
Just wow!

src/layers.scss
```diff
@@ -0,0 +1,4 @@
+// Establish CSS cascade layers.
+// ⚠️ WARNING: This file must be called before other React UI styles for the cascade layers to work as intended.
+
+@layer theme, foundation, helpers, components, utilities;
```
adamkudrna, line 4: I realized the component styles are currently unlayered, thus ordered as last (after the `utilities` layer). It seems this is what I need to do with all component classes:

```css
.root {
    @layer components.checkboxField {
        /* root styles */
    }
}
```

mbohal (approved):

dacerondrej (approved):

bedrich-schindler:
For some reason, Table` is not working in this branch.`master` is OK.

image

bedrich-schindler (changes requested):
For some reason, Table is not working in this branch. `master` is OK.

313993940-02e712c7-4a77-4e3b-9039-2f0858c252bd

adamkudrna:
> For some reason, Table is not working in this branch. `master` is OK.

Fixed, I forgot to rename nested TableCell styles.

bedrich-schindler (approved):

PR #539, opened 2024-04-24
Minor release workflow improvements

adamkudrna:
- Label `fix/bug/bugfix` branches as `fix`
- Include PR authors in release notes
- Remove permissions settings from the release workflow


adamkudrna:
After merging this PR, `template` and `footer` options were force-pushed because it seemed impossible to debug the output on a branch other than `master`. 🙁

PR #542, opened 2024-05-30
Introduce usage of nested `RUIProvider` (#541)

### LINKED ISSUE #541: GlobalProps inheritance
Presently the `GlobalProps` mechanism does not allow overloading of only specific props. The entire object gets overloaded and thus it is near to impossible to use it to modify `GlobalProps` only for certain parts of the app.

**Minimal Example: **

```js
<RUIProvider
  globalProps={{
    TextField: { layout: 'horizontal' },
  }}
>
  <RUIProvider
    globalProps={{
      TextField: { size: 'small' },
    }}
  >
    <TextField label="username" />
  </RUIProvider>
</RUIProvider>
```

**Expected `GlobalProps`:** layout=horizontal, size=small
**Received `GlobalProps`:** size=small

bedrich-schindler:
Closes #541

### DIFF
--- a/src/docs/customize/global-props.md
+++ b/src/docs/customize/global-props.md
@@ -83,3 +83,53 @@ React.createElement(() => {
   );
 });
 ```
+
+## Nesting
+
+Global props can be nested. This is useful e.g. when you want to configure
+props across whole application and then override some of them in a specific
+part of the application.
+
+When nested `RUIProvider` is used, the props are merged deeply together. This
+means that you can extend specific object with new props or override existing
+ones. If you need to remove some prop, you can set it to `undefined`.
+
+```docoff-react-preview
+React.createElement(() => {
+  const [variant, setVariant] = React.useState('filled');
+  return (
+    <RUIProvider globalProps={{
+      Grid: {
+        columns: {
+          xs: '1fr',
+          md: '1fr 1fr',
+        },
+        justifyItems: 'center',
+        rows: {
+          xs: '50px',
+          md: '100px',
+        },
+      },
+    }}>
+      <RUIProvider globalProps={{
+        Grid: {
+          columns: {
+            sm: '1fr 1fr 1fr',
+          },
+          justifyItems: 'undefined',
+          rows: undefined,
+        },
+      }}>
+        <Grid>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+          <docoff-placeholder bordered>Grid item</docoff-placeholder>
+        </Grid>
+      </RUIProvider>
+    </RUIProvider>
+  );
+});
+```
--- a/src/provider/__tests__/RUIProvider.test.jsx
+++ b/src/provider/__tests__/RUIProvider.test.jsx
@@ -3,8 +3,9 @@ import {
   render,
   within,
 } from '@testing-library/react';
-import { Badge } from '../../components/Badge';
 import { Alert } from '../../components/Alert';
+import { Badge } from '../../components/Badge';
+import { Grid } from '../../components/Grid';
 import RUIProvider from '../RUIProvider';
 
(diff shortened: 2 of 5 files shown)


src/utils/mergeDeep.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const newObject = { ...prev };
+
+  Object.keys(obj).forEach((key) => {
+    const pVal = prev[key];
+    const oVal = obj[key];
+
+    if (Array.isArray(pVal) && Array.isArray(oVal)) {
+      newObject[key] = pVal.concat(...oVal);
```
bedrich-schindler, line 21: Not sure whether we want to concat arrays or just replace them.  Props are typically plain values or objects so it can be omitted. What do you think @adamkudrna @mbohal @dacerondrej?

src/docs/customize/global-props.md
```diff
@@ -83,3 +83,13 @@ React.createElement(() => {
   );
 });
 ```
+
+## Nested Usage
```
bedrich-schindler, line 87: I was thinking about the code example. But in the end, I decided to omit it as it seems quite straighforward to me.

mbohal: I would prefer to include it to illustrate scalar, object and array handling.

src/utils/mergeDeep.js
mbohal: I think we want replace rather the `concat`. 
Replace is safer, because otherwise there is no way to remove something from an array prop for a subtree of the app. I think this needs to be changed here.

In case we need more flexibility, we could in the future allow dynamic prop value definition. Maybe the value of a prop could be defined by a function accepting the `oldProp` as an argument.

mbohal (changes requested):

src/docs/customize/global-props.md
bedrich-schindler: Added example that contains adding new prop, overriding and removing. Grid is only component where objects can be logically used. We do not have any array as a prop so it cannot be documented.

mbohal (approved):

dacerondrej (approved):

src/docs/customize/global-props.md
```diff
@@ -83,3 +83,53 @@ React.createElement(() => {
...
```
adamkudrna, line 87:
```suggestion
## Nesting
```

I think plain "Nesting" is fine.

src/utils/mergeDeep.js
```diff
@@ -0,0 +1,28 @@
...
+export const mergeDeep = (...objects) => objects.reduce((prev, obj) => {
+  if (obj == null) {
+    return prev;
+  }
+
+  const newObject = { ...prev };
+
+  Object.keys(obj).forEach((key) => {
+    const pVal = prev[key];
+    const oVal = obj[key];
```
adamkudrna, line 18: While `obj` could be acceptable, I have no idea what `pVal` or `oVal` means. Can we come up with better names please? 🙂

src/utils/__tests__/mergeDeep.js
adamkudrna, line 1: I suppose the tests have been already reviewed by @mbohal so I'm skipping it now.

src/provider/__tests__/RUIProvider.test.jsx
```diff
@@ -36,4 +37,77 @@ describe('rendering', () => {
...
+                Content text
+              </div>
+            </Grid>
+          </RUIProvider>
+        </RUIProvider>
+      </RUIProvider>
+    ));
+
+    // Assert alignContent
+    expect(dom.container.firstChild.style.cssText.includes('--rui-local-align-content')).toBeFalsy();
```
adamkudrna, line 95: Is truthy/falsy sufficient? Can we be sure that e.g. `column` is correctly overwritten to `row dense`?

adamkudrna (commented):
Wow, nice work! 👍🏻

src/utils/mergeDeep.js
mbohal: +1

src/provider/__tests__/RUIProvider.test.jsx
bedrich-schindler: Is is sufficient to compare it to false as `alignContent` is reset via `undefined`.

adamkudrna (approved):

Issue #489, opened 2023-09-18
Button label not shown on hover


mbohal:
When the `<Button>` component is rendered with `labelVisbility` set to `none` or to value smaller the current viewport the `label` value is not accessible.

I think it should appear on mouse hover.

adamkudrna:
The [invisible label](https://react-ui.io/components/button#icon-buttons) does remain accessible:

<img width="802" alt="obrazek" src="https://github.com/react-ui-org/react-ui/assets/5614085/f28fde9c-af2c-4698-843e-32b1d9e69260">

FF dev tools:
<img width="452" alt="obrazek" src="https://github.com/react-ui-org/react-ui/assets/5614085/57548afd-f2aa-42c9-9f2e-dcb4977fd580">

What you probably mean is the button does not provide any textual alternative to sighted users. In such case this is what you would do:

1. add the `title` attribute (which may be annoying for AT users as both texts are announced: _"My action, Title description, button"_, or something like that), or
2. combine the button with custom Tooltip; unfortunately, RUI currently does not have such component (Popover is close, but you need to click to make it show up).

To sum it up, current behavior is correct. But I'll be happy to implement a `Tooltip` component which could be paired with `Button`s to improve the UX while not putting accessibility at stake.

mbohal:
> What you probably mean is the button does not provide any textual alternative to sighted users.

Yes, that is what I had in mind.

Lets go with option 1). When we will have the component `<Tooltip>` we can use it here. We do not have the capacity to implement `<Tooltip>` at this point.

adamkudrna:
Then I suggest putting something like this into the docs:

_For icon-only buttons, consider using the `title` attribute to provide a textual alternative._

OK?

mbohal:
@adamkudrna 

I think there are two ways to solve this:

1.  Update the docs as you suggest and make user pass in both `label` and `title` props
2.  Use user specified `title` if provided or use the text passed in via the `label`  prop if not provided

I think I prefer 2) developers (e.g. myself) will keep forgetting to pass in the `title` . Are there any arguments why 1) would be a better idea?

mbohal:
The docs will be updated in the `General guidelines` section.

When mentioning the option to add `title` it should be mentioned that it will lead to aria reader a duplicit description.

adamkudrna:
[Adobe Spectrum](https://spectrum.adobe.com/page/button/#Tooltip-when-the-label-is-hidden):

[...]

PR #543, opened 2024-05-30
Allow form fields' `label` to accept any node (#526)

### LINKED ISSUE #526: Allow to pass label as `node` into form elements
We need to add the possibility to pass either `string` or `node` into the `label` prop of form field components.

This is due to a requirement to use the label as a trigger for showing a `Popover` component.

Change needs to take place on:
* https://react-ui.io/components/CheckboxField/
* https://react-ui.io/components/FileInputField/
* https://react-ui.io/components/Radio/
* https://react-ui.io/components/SelectField/
* https://react-ui.io/components/TextArea/
* https://react-ui.io/components/TextField/
* https://react-ui.io/components/Toggle/

bedrich-schindler:
Closes #526

### DIFF
--- a/src/components/CheckboxField/CheckboxField.jsx
+++ b/src/components/CheckboxField/CheckboxField.jsx
@@ -114,7 +114,7 @@ CheckboxField.propTypes = {
   /**
    * Checkbox field label.
    */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.node.isRequired,
   /**
    * Placement of the label relative to the input.
    */
--- a/src/components/FileInputField/FileInputField.jsx
+++ b/src/components/FileInputField/FileInputField.jsx
@@ -123,9 +123,9 @@ FileInputField.propTypes = {
    */
   isLabelVisible: PropTypes.bool,
   /**
-   * Text field label.
+   * File input field label.
    */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.node.isRequired,
   /**
    * Layout of the field.
    *
--- a/src/components/CheckboxField/README.md
+++ b/src/components/CheckboxField/README.md
@@ -45,6 +45,9 @@ See [API](#api) for all available options.
   send me any emails” which would mean that the user needs to turn the
   checkbox **on** in order for something **not** to happen.
 
+- **Use text labels** unless it is necessary to wrap text label into
+  Popover-like to component to provide additional info about the field.
+
 - Only make the CheckboxField's label invisible when there is **another visual
   clue** to guide users through toggling the input.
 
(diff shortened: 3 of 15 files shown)


bedrich-schindler:
Documentation does not specifically mention data type of labels, only in the props table.  @adamkudrna @mbohal @dacerondrej, do you think we need to extend documentation of those components to mention this possibility or not? Personally, I do not see any additional value to this.

I expect to extend documentation of Popover to mention possibility to extend label with Popover (https://github.com/react-ui-org/react-ui/issues/489), but do not know what to say about it in documentation of inputs ...

src/components/CheckboxField/CheckboxField.jsx
mbohal, line 1: I think we also need to add support for the `InputGroup` component.

mbohal, line 1: I think we should document this with code examples.

mbohal (changes requested):

src/components/CheckboxField/CheckboxField.jsx
mbohal: I checked and I think the `InputGroup` is the only one, where we want to include it.

Other components mostly render `label` "within" something (e.g. `Badge`, `Button`). Then there is `TextLink` where the name kind of implies text prop.

bedrich-schindler: Please check my comment: https://github.com/react-ui-org/react-ui/pull/543#issuecomment-2139112900

Personally, I don't think so we need to show examples of everything we support. End user must know React, so he/she must know what React.Node means and therefore I do not see any reason to have it as example. What more, I think user should use always except situations when he/she needs to provide additional info about the field. I can imagine to extend prop documentation or general guidelines with something like "Use text labels unless it is necessary to wrap text label into Popover-like to  component to provide additional info about the field". 

There was separate issue for that I guess but I think it disappeared from the board after Monday's meeting.

What do you think @mbohal @adamkudrna @dacerondrej

dacerondrej (approved):

src/components/CheckboxField/CheckboxField.jsx
mbohal: Ok, lets put it in general guidelines

src/components/FileInputField/FileInputField.jsx
```diff
@@ -123,9 +123,9 @@ FileInputField.propTypes = {
    */
   isLabelVisible: PropTypes.bool,
   /**
-   * Text field label.
+   * File input field label.
```
adamkudrna, line 126: Thanks!

adamkudrna (approved):

Issue #468, opened 2023-06-19
Add support of feedback colors to `Modal`


adamkudrna:
Support all relevant Component Colors (feedback + neutral): https://react-ui.io/docs/foundation/colors/#component-colors for these properties:

- border,
- footer background,
- anything else?

PR #544, opened 2024-05-30
Re-implement `Modal` component using HTMLDialogElement (#461)

### LINKED ISSUE #461: Use native `<dialog>` element for `Modal`
We want to migrate `Modal` currently using div to native `<dialog>` element.

Relevant links:
https://github.com/react-ui-org/react-ui/pull/432#discussion_r1033535677
https://github.com/react-ui-org/react-ui/issues/453

### LINKED ISSUE #537: In docs Modal is not dismissed by pressing `Esc`
Likely it is related to the documentation platform and not the component itself as it seems to work in app.

Also the fix might need to be ported to RADES.

bedrich-schindler:
Closes #461, closes #537.


src/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -14,7 +14,10 @@ import { ModalContent } from '../ModalContent';
 import { ModalFooter } from '../ModalFooter';
 import { ModalHeader } from '../ModalHeader';
 
-describe('rendering', () => {
+// Test suites skipped due tu missing implementation of HTMLDialogElement in jsdom
```
bedrich-schindler, line 17: As it requires change of test environment, I'm preparing separate pull request with those changes.

bedrich-schindler: https://github.com/react-ui-org/react-ui/pull/545

mbohal (approved):

src/theme.scss
mbohal, line 1: We need to manually set the border to mimick the old default behavior. As it is the default is a black thick border on Chromium.

mbohal (changes requested):

dacerondrej (changes requested):

src/theme.scss
mbohal: Maybe we could make the border color be configurable as required in https://github.com/react-ui-org/react-ui/issues/468 in this PR?

bedrich-schindler:
Update CSS properties in README.md.

src/theme.scss
bedrich-schindler: We decided to do it separately in #468 as it requires more that just setting one attribute. I will reset it to `0` to obtain same appearance as before change.

mbohal (approved):

src/components/Modal/Modal.jsx
```diff
@@ -18,32 +21,27 @@ const preRender = (
...
-    className={styles.backdrop}
+  <dialog
+    {...transferProps(restProps)}
+    className={classNames(
+      styles.root,
+      getSizeClassName(size, styles),
+      getPositionClassName(position, styles),
+    )}
     onClick={(e) => {
+      e.stopPropagation();
```
adamkudrna, line 32: The effect of this is way too big, don't you think…? Why do we need it?

src/components/Modal/Modal.jsx
```diff
@@ -18,32 +21,27 @@ const preRender = (
...
+    }}
+    onClose={(e) => {
       e.preventDefault();
       if (closeButtonRef?.current != null) {
         closeButtonRef.current.click();
       }
     }}
+    ref={childrenWrapperRef}
     role="presentation"
```
adamkudrna, line 41: After moving to the `<dialog>` element, the `role` attribute is now incorrect.

```suggestion
```

src/components/Modal/Modal.jsx
```diff
@@ -18,32 +21,27 @@ const preRender = (
...
+    {...transferProps(restProps)}
+    className={classNames(
+      styles.root,
+      getSizeClassName(size, styles),
+      getPositionClassName(position, styles),
+    )}
     onClick={(e) => {
+      e.stopPropagation();
+    }}
+    onClose={(e) => {
```
adamkudrna, line 34: This seems to fix #537.

src/components/Modal/Modal.jsx
```diff
@@ -18,32 +21,27 @@ const preRender = (
...
       e.preventDefault();
       if (closeButtonRef?.current != null) {
         closeButtonRef.current.click();
       }
     }}
+    ref={childrenWrapperRef}
```
adamkudrna, line 40: A reference to the top-level HTML element can be simply passed through `restProps`, is it correct? I therefore think the `childrenWrapperRef` prop can be removed.

src/components/Modal/Modal.jsx
```diff
@@ -121,7 +122,7 @@ Modal.propTypes = {
    */
   children: PropTypes.node,
   /**
-   * Reference to close button element. It is used to close modal when Escape key is pressed or the backdrop is clicked.
+   * Reference to close button element. It is used to close modal when Escape key is pressed.
```
adamkudrna, line 125: ⚠️  This means we are losing functionality of closing the Modal by clicking on backdrop. That has a big impact on UX.

IIRC, we added this functionality to Spirit Modal later, so maybe it's not that easy.

I'd like to ask you to:

- either bring back/re-implement the functionality of closing the dialog on backdrop click (watch out for "nested" dialogs),
- or file an issue that we get on rather sooner than later.

https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/src/hooks/useClickOutside.ts

https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web-react/src/components/Modal#stacking-modals

https://spirit-design-system.netlify.app/packages/web-react/src/components/modal/

src/components/Modal/Modal.jsx
adamkudrna: Oh, I see we use the reference in a hook… I'm eager to learn new things then 🙂.

src/components/Modal/Modal.module.scss
```diff
@@ -13,33 +20,34 @@
...
     }
 
-    .backdrop {
-        position: fixed;
-        top: 0;
-        left: 0;
-        z-index: settings.$backdrop-z-index;
-        width: 100vw;
-        height: 100vh;
+    .root:has(.root)::backdrop {
```
adamkudrna, line 39: ⚠️ `<dialog>`s rather should not be nested like that. It is OK to stack `<dialog>`s visually, but in practice, it should mean they are siblings in the DOM. While the HTML spec does not explicitly discourage from nesting one `<dialog>` into another in the DOM (and while it seems to "work" somehow), we are not quite sure of the consequences.

OK:

```html
<dialog></dialog>
<dialog></dialog>
```
⚠️ Not OK:

```html
<dialog>
  <dialog></dialog>
</dialog>
```

https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web-react/src/components/Modal#stacking-modals

src/components/Modal/Modal.module.scss
```diff
@@ -1,9 +1,16 @@
+// 1. Modal uses <dialog> element that uses the browser's built-in dialog functionality, so that:
+//    * visibility of the .root element and its backdrop is managed by the browser
+//    * positioning of the .root element and its backdrop is managed by the browser
+//    * z-index of the .root element and its backdrop is not needed as dialog is rendered in browser's Top layer
+// 2. When dialogs are stacked, only the topmost dialog should have a backdrop.
```
adamkudrna, line 5: I don't think so. Another `Dialog` means another layer that prevents interaction with content underneath it. That implies the backdrop should always apply.

src/components/Modal/Modal.module.scss
```diff
@@ -69,12 +77,8 @@
...
     }
 
-    .isRootPositionCenter {
-        top: 50%;
-        transform: translate(-50%, -50%);
-    }
-
     .isRootPositionTop {
+        position: sticky;
         top: var(--rui-local-outer-spacing);
```
adamkudrna, line 82: I think this should do the job:

```suggestion
        top: var(--rui-local-outer-spacing);
        bottom: auto;
```

src/components/Modal/Modal.module.scss
```diff
@@ -13,33 +20,34 @@
...
+    }
+
+    .root[open] {
+        display: flex;
+        animation: fade-in theme.$animation-duration ease-out;
+    }
+
+    .root[open]::backdrop {
         background: theme.$backdrop-background;
+        animation: fade-in theme.$animation-duration ease-out;
```
adamkudrna, line 50: This is DRYer:

```suggestion
        animation: inherit;
```

src/components/Modal/_settings.scss
adamkudrna, line 1: +1 for cleanup

src/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -14,7 +14,10 @@ import { ModalContent } from '../ModalContent';
 import { ModalFooter } from '../ModalFooter';
 import { ModalHeader } from '../ModalHeader';
 
-describe('rendering', () => {
+// Test suites skipped due to missing implementation of HTMLDialogElement in jsdom
+// See https://github.com/jsdom/jsdom/issues/3294
```
adamkudrna, line 18: Fixed in #545.

src/components/Modal/README.md
```diff
@@ -1065,27 +1065,29 @@ accessibility.
...
+| `--rui-Modal__footer__gap`                           | Modal footer gap between children                           |
+| `--rui-Modal__backdrop__background`                  | Modal backdrop background (including `url()` or gradient)   |
+| `--rui-Modal--auto__min-width`                       | Min width of auto-sized modal (when enough screen estate)   |
+| `--rui-Modal--auto__max-width`                       | Max width of auto-sized modal (when enough screen estate)   |
+| `--rui-Modal--small__width`                          | Width of small modal                                        |
+| `--rui-Modal--medium__width`                         | Width of medium modal                                       |
+| `--rui-Modal--large__width`                          | Width of large modal                                        |
+| `--rui-Modal--fullscreen__width`                     | Width of fullscreen modal                                   |
+| `--rui-Modal--fullscreen__height`                    | Height of fullscreen modal                                  |
+| `--rui-Modal__animation__duration`                   | Duration of animation used (when opening modal)             |
```
adamkudrna, line 1089: > when opening modal

Currently, closing of `Modal` is not animated because the component is simply removed from the DOM.

@mbohal Do we also want animated closing? That would mean delayed removal from the DOM.

src/components/Modal/Modal.jsx
adamkudrna, line 1: obrazek

I'd expect more deletions than additions. 😞  Can we get rid of some more features in our focus management? Or anywhere else?

adamkudrna (changes requested):
Nice job! 👏🏻  Even with the decision of not having the closed Modal in the DOM. 👍

I just think we cannot remove the click-on-backdrop-to-close functionality.

src/components/Modal/Modal.jsx
mbohal: On call we agreed that for now we do not need this functionality.

Not having it has the added benefit of not loosing unsaved work in modal on accidental click.
If we need it in the future it will be added via a new issue.

src/components/Modal/README.md
bedrich-schindler: I would like to add one thing. Closing animation is hard to impelement. We do not have close handler, we have `closeButtonRef`. I tried to implement it somehow today but without success. So it is up to our decision whether we want to spend some time on it or not.

src/components/Modal/Modal.jsx
bedrich-schindler: `childrenWrapperRef` is part of `preRender` function used by `Modal` component. As it is passed as `ref`, it cannot be send to the `preRender` function just via rest props.. `childrenWrapperRef` must be there to call `showModal()` to show modal dialog (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog).

src/components/Modal/Modal.jsx
bedrich-schindler: This is part of removing functionality commented in https://github.com/react-ui-org/react-ui/pull/544#discussion_r1643152681

src/components/Modal/Modal.module.scss
bedrich-schindler: Let's discuss it on Monday.

src/components/Modal/Modal.jsx
bedrich-schindler: Yes. As mentioned in PR description.

src/components/Modal/Modal.jsx
bedrich-schindler: Removed.

src/components/Modal/Modal.jsx
bedrich-schindler: I introduced it back with possibility to disable it as decided on Monday's call.

bedrich-schindler:
PR updated:

* Introduced `allowCloseOnBackdropClick` and `allowCloseOnEscapeKey` prop, both set to true by default. It allows end-user to configure modal in such way that he/she can prevent closing by clicking on backdrop or pressing Escape key.
* Improved blocking of firing click events on primary and close button as there were missing conditions checking whether the button is disabled (it was there for a long time)
* The biggest problem was with propagation of events, Enter was mistakenly closing dialog in unwanted cases. I can talk about it on Monday if you wish.

While I introduces `allowCloseOnBackdropClick` and `allowCloseOnEscapeKey`, I am thinking about `allowProceedOnEnterKey` (or something like that) to disable Enter functionality which can be unwanted in some scenarios.

src/components/Modal/Modal.module.scss
bedrich-schindler: I reverted the change, so every Modal will open its backdrop. So the background gets darker and darker every time new layer is opened. It is standard behavior.

bedrich-schindler:
Just for info, tests need to be updated in https://github.com/react-ui-org/react-ui/pull/545 as new test environment is necessary.

src/components/Modal/README.md
mbohal: After a fruitless afternoon I concur with @bedrich-schindler that:

> Closing animation is hard to implement.

However, this change is not BC in terms of usage, so I think we should merge it as is.

We can add hide animation int he future. This will very likely change the API though…

mbohal: On video call agreed that for the time being we will only support opening animation.

src/components/Modal/README.md
adamkudrna, line 1: I checked the "Launch modal as form" example and found out there is no `<form>` element in it. Is it purposely? Maybe we didn't want the `<form>` there before switching to `<dialog>` but it would add more semantic meaning (and maybe functionality) now.

There is the `method="dialog"` form attribute (or `formmethod="dialog"` for buttons) if we find it useful.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#usage_notes

adamkudrna (commented):
I have a note regarding forms inside dialog, otherwise OK for me.

bedrich-schindler:
Can @adamkudrna or anybody else help me with this?

image

I returned to this to complete the pull request but suddenly, I have all the texts inside of Modal in white even though there is no change in the code. `master` is OK. Does anybody know why? I do not want to rebased it until this is resolved.

adamkudrna:
> Does anybody know why? I do not want to rebased it until this is resolved.

I added a fix. It seems to happen only when the docs is switched to dark mode.

bedrich-schindler:
I have rebased and squashed it. I'm gonna perform self-review first as I have seen that documentation is not up-to-date with latest changes and then I will request review from you-

bedrich-schindler:
Since rebase:

I fixed submitting dialog when form fields are present. It is not working either on master. Formerly, it was part of auto focus hook which I have modified. It was also containing focus trap that it can be removed now. Now, this functionality is part of dialogOnKeyDownHandler.js. 

Then I have improved documentation and did minor fixes.

bedrich-schindler:
I made separate commit (https://github.com/react-ui-org/react-ui/pull/544/commits/2514fd74af700250c12546aa4f92952c56256c52) to demonstrate example with native form. Personally, I would not encourage this approach as every application differs in the way how it handles forms, state, validations and so. But it is true that it can be useful for example in that way, that native form will not be submitted until HTML validations are passing.

We had 4 examples of different kind, so I added native form as fifth and atypically explained it in the example. Even though it is atypical, I would leave it there as it is only place how we can describe example with modal with creating explicit section.  And we do not want to handle all possibilities in the documentation, so it compromise I think.

@adamkudrna @mbohal, you can review the pull request and tell whether to include this native form example or drop it.

bedrich-schindler:
> I fixed submitting dialog when form fields are present. It is not working either on master. Formerly, it was part of auto focus hook which I have modified. It was also containing focus trap that it can be removed now. Now, this functionality is part of dialogOnKeyDownHandler.js.

I just want to mention (it will be part of commit message) that even though HTML Dialog states that it automatically focuses first focusable element in dialog, it does not work as our solution. The problem is that it autofocuses only some elements and for example states when there are only buttons and no input fields do not work. So we have to keep this functionality and we are not able to remove it!

**The only question is: Do we want to have autoFocus always enabled? Personally, I would suggest removing autoFocus prop and make it mandatory as native solution would automatically focuses some elements even when our autoFocus is off.**

mbohal:
> The only question is: Do we want to have autoFocus always enabled? Personally, I would suggest removing autoFocus prop and make it mandatory as native solution would automatically focuses some elements even when our autoFocus is off.

Agreed we want to make `autoFocus` always on.

src/components/Modal/Modal.jsx
```diff
@@ -47,29 +48,32 @@ export const Modal = ({
...
-  const dialogRef = useRef();
+  const internalDialogRef = useRef();
 
   useEffect(() => {
-    dialogRef.current.showModal();
+    internalDialogRef.current.showModal();
   }, []);
 
-  useModalFocus(autoFocus, dialogRef, primaryButtonRef);
+  useImperativeHandle(dialogRef, () => internalDialogRef.current);
```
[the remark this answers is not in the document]
mbohal: Explained on call.
Could you please add a comment.

mbohal (changes requested):

src/components/Modal/_hooks/useModalFocus.js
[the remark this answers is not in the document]
bedrich-schindler: I checked this and you are right that it is working as you say on our deployed demo, but it does behave the same way in real application where old RUI is used and fields there are not focused on initial load either. I just weird and I am not sure what to do with it. 

I will try to find whether it works or not on current master and I will let you know.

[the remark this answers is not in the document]
bedrich-schindler: So, I tried it and it works on master (It might have been caching, so I disregard last message).

I have a theory. As I wrote few days ago, [<dialog> element implements autofocus functionality by design.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility). If I remove `useModalFocus` from `Modal.jsx`, I can see how native autofocus works: it focuses first focusable element which is close button and focus ring is not visible. This is standard scenario when browser automatically focuses some element but it has not focus ring visible.

If I return `useModalFocus` hook, my theory is that it interferes with native autofocus. For dialog with input fields, it does focuses input but without focus ring. For dialog with button only, it focuses button with focus ring.

The problem points to src/styles/tools/_accessibility.scss where mixin for focus-ring is defined:

```
@mixin focus-ring() {
    &:focus-visible {
        outline: theme.$focus-outline;
        outline-offset: theme.$focus-outline-offset;
        box-shadow: theme.$focus-box-shadow;
    }
}
```

If I change it from `:focus-visible` to `:focus`, it works.

bedrich-schindler:
I have merged PR with happy-dom into this PR.

bedrich-schindler:
After discussion with @mbohal, we decided not to change our `autoFocus` functionality. Even though native autofocus functionality was introduced into <dialog>, it does not work as good as out `autoFocus` functionality because it selects first item which is typically close button in our case, so that's the reason we want to leave it as it is.

I dug deep into in again @mbohal and I found out, that with autoFocus disabled (through props, not through deleting whole functionality), our autoFocus internally forces focus on <dialog> element, so it still can prevent native behavior. I had been testing it with completely removed functionality and probably cached page, so I had been obtaining wrong results before.

src/components/Modal/README.md
```diff
@@ -92,11 +92,19 @@ See [API](#api) for all available options.
...
 
-- Modal **automatically focuses the first non-disabled form field** by default
-  which allows users to confirm the modal by hitting the enter key. When no
-  field is found then the primary button (in the footer) is focused. To turn
+- Modal **automatically focuses the first non-disabled form field** by default.
+  When no field is found then the primary button (in the footer) is focused. To turn
   this feature off, set the `autofocus` prop to `false`.
 
+- Modal **submits the form when the user presses the `Enter` key** . The primary
+  button is clicked in this case. To turn this feature off, set the
```
adamkudrna, line 100: I think the click is just _emulated_, isn't it?

src/components/Modal/README.md
```diff
@@ -92,11 +92,19 @@ See [API](#api) for all available options.
...
+  `allowPrimaryActionOnEnterKey` prop to `false`.
+
+- Modal **closes when the user presses the `Escape` key**. The close button is
+  clicked in this case. To turn this feature off, set the `allowCloseOnEscapeKey`
```
adamkudrna, line 104: The same about "clicking" here.

src/components/Modal/README.md
```diff
@@ -225,9 +237,33 @@ React.createElement(() => {
             </ModalHeader>
             <ModalBody>
               <ModalContent>
-                <FormLayout fieldLayout="horizontal">
-                  <TextField label="Username" />
+                <FormLayout fieldLayout="horizontal" labelWidth="limited">
```
adamkudrna, line 240: Thanks for extending the form with new fields!

src/components/Modal/README.md
```diff
@@ -246,6 +282,91 @@ React.createElement(() => {
...
+                </p>
+                <p>
+                  First, you need to set <code>allowPrimaryActionOnEnterKey</code>
+                  to <code>false</code> and remove <code>onClick</code> from the
+                  primary button. Then, you need to set <code>form</code> attribute
+                  on the primary button to the <code>id</code> of the form to
+                  connect it with the form.
+                </p>
+                <p>
+                  Although we do not encourage using this approach, it is still
```
adamkudrna, line 318: Why don't we encourage the native solution? Is there something wrong when combining it with React? I'd like to see the reason mentioned here.

src/components/Modal/README.md
```diff
@@ -151,6 +159,10 @@ React.createElement(() => {
         label="Launch modal as form"
         onClick={() => setModalOpen(4)}
       />
+      <Button
+        label="Launch modal as native form"
+        onClick={() => setModalOpen(5)}
+      />
```
adamkudrna, line 165: I think this section has become too long and deserves splitting into more blocks. From my experience, using the modal dialogs with forms is a frequent use case and it deserves a prominent place in the design system docs. This is how I see it:

1. Create a new section called "Forms" right before the "Mouse and Keyboard Control" section.
2. Explain shortly the Modal can be used with or without the native `<form>` element. If you can also tell what are the use cases, even better.
3. Extract the text from the last example (native form) so it's right in the document and not hidden in the demo.
4. In the original section, please move the "Launch modal as dialog" example to the first place. Ever since we have the Modal component, I always click on the first button and immendiately hate myself for having to wait for the blocking modal to close 🙈.

src/components/Modal/_helpers/dialogOnCancelHandler.js
```diff
@@ -0,0 +1,19 @@
+// Disable coverage for the following function
+/* istanbul ignore next line */
```
adamkudrna, line 2: 🕌

src/components/Modal/_helpers/dialogOnClickHandler.js
```diff
@@ -0,0 +1,34 @@
...
+  // Detection of the click on the backdrop is based on the following conditions:
+  // 1. The click target is the dialog itself. This prevents detection of clicks on the dialog's children.
+  // 2. The click is outside the dialog's boundaries.
+  const dialogRect = dialogRef.current.getBoundingClientRect();
+  const isClickedOnBackdrop = dialogRef.current === e.target && (
+    e.clientX < dialogRect.left
+    || e.clientX > dialogRect.right
+    || e.clientY < dialogRect.top
+    || e.clientY > dialogRect.bottom
+  );
```
adamkudrna, line 23: Do we still need this with native dialog? The `<dialog>` triggers the `close` event so I was wondering if we could listen on it?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#handling_the_return_value_from_the_dialog

webpack.config.babel.js
```diff
@@ -4,7 +4,7 @@ const StyleLintPlugin = require('stylelint-webpack-plugin');
 const TerserPlugin = require('terser-webpack-plugin');
 const VisualizerPlugin = require('webpack-visualizer-plugin2');
 
-const MAX_DEVELOPMENT_OUTPUT_SIZE = 3300000;
+const MAX_DEVELOPMENT_OUTPUT_SIZE = 3400000;
```
adamkudrna, line 7: How come? 🤔 

obrazek

adamkudrna (changes requested):
Well done with beautiful demos! I only have two major requests/comments: 

- Make a new "Forms" section. Forms in dialogs are a common use case. This is why I click "Request changes" here.
- Consider if we can get more out of the native functionality of the `<dialog>`. But maybe you have done an in-depth reserach and I see it wrong.

Thank you!

src/components/Modal/_helpers/dialogOnClickHandler.js
bedrich-schindler: If it wasn't needed, it wouldn't be here. Neither close nor cancel event is called on backdrop click. If you want to close dialog by clicking on backdrop, you have to listen to click event and detect user clicked outside of dialog element or use different method of detection. However, this is most common approach.

It is quite hard to decide what to comment. This is default behavior/knowledge and developer is supposed to know this if he/she want to dig into it. Now, each line of each event handler is commented. But I would add docblock for each function to make it crystal clear.

src/components/Modal/_helpers/dialogOnCancelHandler.js
bedrich-schindler: Yeah. I was also surprised ...

src/components/Modal/README.md
bedrich-schindler: It is not. `click()` function is called directly on HTML button, it is not emulated, it's real click that also triggers its own click event. So I would not call it emulated.

@mbohal can correct me, but emulated event would be dispatchEvent(event), but not click().

webpack.config.babel.js
mbohal: There is no direct relationship between lines changed and bundle size. For example bundle size is not effected by comments and such.

src/components/Modal/README.md
adamkudrna: > A click is programmatically triggered on the primary  button in this case.

This is great. Thanks!

bedrich-schindler:
Mention that we are using native <dialog> and that we do not want to have hidden dialog element, we want remove it when closed ...

src/components/Modal/_helpers/dialogOnClickHandler.js
bedrich-schindler: I added docblock for those handlers. It still require knowledge, but it is easier to understand now. Most problematic part is Escape click that can end-up in 3 handles, but it is described in native documanation.

bedrich-schindler:
Everything should be done as agreed on Monday.

mbohal (approved):

src/components/Modal/Modal.module.scss
adamkudrna, line 1: ⚠️  The auto-sized Modal is broken now. Weirdly enough, removing `width: auto` from `.isRootSizeAuto` (line 77) will fix it 🤷🏻‍♂️.

src/components/Modal/README.md
```diff
@@ -839,7 +723,241 @@ is focused.
 Autofocus is enabled by default, so if you want to control the focus of
 elements manually, set the `autoFocus` prop on Modal to `false`.
 
-## Scrolling Long Content
+## Use Cases
```
adamkudrna, line 726: OK, looks good. What do you think about grouping the features so we have it like:

- composition
- features
- use cases

?

obrazek

src/components/Modal/README.md
```diff
@@ -839,7 +723,241 @@ is focused.
...
+
+### Interaction blocking
+
+Modal can be used to block user interaction while some action is being
```
adamkudrna, line 730:
```suggestion
Modal can be used to block user interaction while an action is being
```

Maybe?

src/components/Modal/README.md
```diff
@@ -839,7 +723,241 @@ is focused.
...
+  );
+});
+```
+
+### Forms
+
+Modal can be used to display forms. It is recommended to use
+[FormLayout](/components/FormLayout) component to layout form fields.
+
+While we support only controlled components, and we encourage you to use them,
```
adamkudrna, line 801: > and we encourage you to use them

Can we point our users anywhere? Do we have it in the docs?

src/components/Modal/README.md
```diff
@@ -92,11 +92,24 @@ See [API](#api) for all available options.
 - **Modal actions** should correspond to the modal purpose, too. E.g. “Delete”
   tells better what happens rather than “OK”.
 
-- Modal **automatically focuses the first non-disabled form field** by default
-  which allows users to confirm the modal by hitting the enter key. When no
-  field is found then the primary button (in the footer) is focused. To turn
+- While native `<dialog>` (that is used under the hood) can be present in DOM,
+  modal is a more feature-rich component that provides more control over the
```
adamkudrna, line 96: I'd consider using capital `M` whan speaking about RUI component:

```suggestion
  Modal is a more feature-rich component that provides more control over the
```

This way we can distinguish our component from a generic design pattern.

adamkudrna (approved):
Nice! Please fix the auto-size CSS, the rest of my comments is just suggestions. Thank you!

src/components/Modal/README.md
bedrich-schindler: I checked the rest of the documentation, I must say that this is the style we use among the documentation. So, if you want to have it as you purpose, I would solve it in separate issue as this is something not related to this change.

Personally, I have not opinion on this. If we talk about the inner component, we typically  emphasize it by using HTML tags rather than using Modal vs modal, Button vs button. I would not say intentionally, but I can see the parent there.

What do you think @mbohal and @adamkudrna ?

adamkudrna: No problem, feel free to dismiss it.

src/components/Modal/README.md
adamkudrna: > have it for every component the same way

I think the "Use cases" section we have now for Modal is different. We don't have anything like that anywhere else (IIRC), so I was thinking about improving the structure just here.

mbohal: I think the `Modal` component is complicate so I do not see a problem in having a more complicated documentation file structure.

I agree with @bedrich-schindler that if the docs should be reworked or unified, it deserves a separate issue. There are no guidelines and each person has a more or less different opinion. about the ideal state. We could argue for long time over this.

I think the docs are good enough to be merged as they are generally and definitely so for RACOM. 

I would like to close this PR rather sooner then later, because it is blocking other work.

bedrich-schindler: If @adamkudrna thinks it would be good to improve docs in such way, please create issue for that. I would approve such change if it is changed across whole docs.

PR #545, opened 2024-05-31
Replace `jsdom` with `happy-dom` testing environment to enable `Modal` tests (#461)

bedrich-schindler:
**This is work-in-progress pull request!**

~~I've replaced `jsdom` with `happy-dom` testing environment and with little changes _(typically because of use of incorrect jest functions )_ and `Modal` tests with `<dialog>` works almost as expected. But for some reason, there are 4 tests in `Grid.test.jsx` that does not work. TBH, I don't know why.~~

~~I debbuged test functions, respectively `toHaveStyle` which is part of `node_modules/@testing-library/jest-dom/dist/matchers-342a062d.js`. `toHaveStyle` internally uses `isSubset` function`. I edited function to be able to debug it step-by-step:

```
function isSubset(styles, computedStyle) {
  return (
    !!Object.keys(styles).length &&
    Object.entries(styles).every(([prop, value]) => {
      const isCustomProperty = prop.startsWith('--');
      const spellingVariants = [prop];
      if (!isCustomProperty) spellingVariants.push(prop.toLowerCase());

      return spellingVariants.some(
        name => {
          const a = computedStyle[name];
          const b = computedStyle.getPropertyValue(name);
          const v = [];
          for (let i = 0; i < computedStyle.length; i++) {
            v.push(computedStyle.item(i));
          }

          return a === value ||
            b === value;
        },
      )
    })
  )
}
```

~~And for some reason, not all CSS properties are part of computed styles (variable `v`)  and that's the reason why the tests fails.~~

~~Can anybody look at it? Can we discuss if this is the right way? Because without `happy-dom`, we won't be able to test our `Modal` component which is problem guys from LMC's Spirit Design System encountered. (cc @adamkudrna @crishpeen @literat).~~

~~If we want to have Modal tested, we either skip those tests for rowGap and columnGap or we can test it different way, for example using `.outerHTML.contains(<css>)`. It is not perfect nor good, but it works at least.~~


bedrich-schindler:
The problem is in `var()` values. For exact values it works as expected.

bedrich-schindler:
Because of missing support for `var()` in testing environment, I have updated `Grid.test.jsx` to compare styles using outerHTMT - the only way how to test it now.

It ready for review.

A mentioned guys from Spirit team for inspiration as their Modal is missing tests.

src/components/Grid/__tests__/Grid.test.jsx
```diff
@@ -116,13 +122,19 @@ describe('rendering', () => {
...
       (rootElement) => expect(rootElement).toHaveStyle({ '--rui-local-justify-items-xs': 'placeholder' }),
     ],
+    // Two following tests must used rootElement.outerHTML.includes() to test presence of CSS variables in the DOM,
+    // because the `toHaveStyle` matcher does not support CSS variables with var() function.
     [
       { rowGap: responsiveSpacingBreakpoints },
-      (rootElement) => expect(rootElement).toHaveStyle(responsiveSpacingStyles('row-gap')),
+      (rootElement) => {
+        Object.entries(responsiveSpacingStyles('row-gap')).forEach(([ccsAttribute, cssValue]) => {
+          expect(rootElement.outerHTML.includes(`${ccsAttribute}: ${cssValue}`)).toBeTruthy();
```
adamkudrna, line 131:
```suggestion
        Object.entries(responsiveSpacingStyles('row-gap')).forEach(([cssAttribute, cssValue]) => {
          expect(rootElement.outerHTML.includes(`${cssAttribute}: ${cssValue}`)).toBeTruthy();
```

adamkudrna (approved):
As far as I can understand what's going on, a ✅ for me.

mbohal (approved):

src/components/Modal/__tests__/Modal.test.jsx
```diff
@@ -368,30 +349,6 @@ describe.skip('functionality', () => {
     assertFocus(el, true);
   });
 
-  it('traps focus', async () => {
```
bedrich-schindler, line 371: HTML dialog handles focus trap manually, so there is no need to test it.  What more, it does not work in testing environment as it is not implemented.

src/components/Modal/_helpers/dialogOnKeyDownHandler.js
```diff
@@ -5,19 +5,24 @@ export const dialogOnKeyDownHandler = (
...
+    if (
       closeButtonRef?.current == null
       || closeButtonRef?.current?.disabled === true
       || !allowCloseOnEscapeKey
-    )
-  ) {
-    e.preventDefault();
+    ) {
+      e.preventDefault();
+    } else if (process?.env?.NODE_ENV === 'test') {
```
bedrich-schindler, line 23: I choose this approach as it is only way to test it. If we do not want to have this here, all tests using close button would stop working. It is consequence of using dialog whose support in testing environment struggles.

mbohal (approved):

src/components/Modal/_helpers/dialogOnKeyDownHandler.js
bedrich-schindler: There are 3 lines describing it, there is even ⚠️ to make it visible ...

PR #546, opened 2024-06-03
Improve docs in relation to using icon only buttons (#489)

### LINKED ISSUE #489: Button label not shown on hover

mbohal:
Closes #489

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -23,10 +23,13 @@ See [API](#api) for all available options.
 - Use **short yet comprehensible labels** to make buttons fit small screens.
 
 - Since buttons are there to take actions, **use a verb** to make it obvious
-  what your buttons do.
+    what your buttons do.
 
 - **Don't overwhelm your UI** with too many high-emphasis actions. There should
-  always be one but chances are that having more of them is not necessary.
+    always be one but chances are that having more of them is not necessary.
+
+- Ensure the **button action is well recognazible** across your target audience.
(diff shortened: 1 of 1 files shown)


bedrich-schindler (approved):

dacerondrej (approved):

src/components/Button/README.md
```diff
@@ -181,16 +192,6 @@ of your choice and display label once you know there is enough room for it.
...
-  <Button
-    label="Label visible from x2l up"
-    labelVisibility="x2l"
-    beforeLabel={<rui-icon icon="pencil" />}
-  />
-  <Button
-    label="Label visible from x3l up"
-    labelVisibility="x3l"
-    beforeLabel={<rui-icon icon="pencil" />}
-  />
```
adamkudrna, line 193: Why removing these? You think it's too much to show everything? (I'm not against your change, after all.)

adamkudrna (approved):
Nicely done! 👍

src/components/Button/README.md
mbohal: > You think it's too much to show everything?

Yes, that is the reason. I think the functionality is sufficiently shown without them and too much code makes for worse readability.

I can put it back, I have no strong feelings about it.

PR #548, opened 2024-06-27
Make `transferProps` available to users (#524)

### LINKED ISSUE #524: make transferProps available to users
In general the `transferProps` function is useful for better accesibility while safely keeping the component code short and manageable.

It would be useful to export it so it can be used inside component libraries based on RUI.

I propose:

1. Move the `transferProps.js` to `src/utils` and export it (like `classNames`)
2. Create a new documentation page with examples in the `Helpers > JS Helpers` section (like `classNames`)
3. Change the `Forwarding HTML Attributes` section in each components `README.md` to link to the new page. This would allow us to also remove examples as they would be provided on the dedicated page.

dacerondrej:
I don't feel strong in wording and writing documentation, so ideas for changes are definitely welcome.

Closes #524.


bedrich-schindler (approved):

adamkudrna (approved):
@dacerondrej Thank you! 🙏🏻

PR #550, opened 2024-07-29
Upgrade Stylelint to v16

### LINKED ISSUE #484: Update Stylelint to version 16
Currently, stylelint@15 cannot be installed because stylelint configuration requires stylelint@14.

adamkudrna:
- Upgrade Stylelint to v16 and `@visionappscz/stylelint-config` to v4.0.0-beta1 

  For now, the order plugin is being repeatedly disabled in places where the new config requires line rules to precede block rules. This may be addressed in the new version of the config.

- Introduce `@stylistic/stylelint-plugin` to watch over stylistic rules removed in Stylelint 16
- Restrict React version to 18.2.x to prevent tests from failing due to deprecation warnings

  Also, remove npm from "engines" as it is not an engine.

Closes #484


package.json
```diff
@@ -54,8 +53,8 @@
   },
   "peerDependencies": {
     "prop-types": "^15.8.0",
```
bedrich-schindler, line 55: Please, add `prop-types` to `devDependencies` as well.

bedrich-schindler (approved):

adamkudrna:
Chaning label as this is a dependency upgrade in the first place.

PR #559, opened 2024-08-23
Fix rendering glitch of `Popover` arrow on low-dpi displays

atmelmicro:
Because of rendering quirks sometimes a line can appear between the popover and arrow
Snímek obrazovky 2024-08-23 v 9 08 11
(more visible on low dpi displays)

This pr fixes that by moving the arrow 1px down
Snímek obrazovky 2024-08-23 v 9 44 18

### DIFF
--- a/src/components/Popover/Popover.module.scss
+++ b/src/components/Popover/Popover.module.scss
@@ -51,52 +51,52 @@
 
     // Sides
     .isRootAtTop {
-        bottom: 100%;
+        bottom: calc(100% + #{theme.$arrow-gap} - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtBottom {
-        top: 100%;
+        top: calc(100% + #{theme.$arrow-gap} - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtLeft {
-        right: 100%;
+        right: calc(100% + #{theme.$arrow-gap} - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtRight {
-        left: 100%;
+        left: calc(100% + #{theme.$arrow-gap} - #{theme.$arrow-safe-rendering-overlap});
     }
 
     // Arrows
     .isRootAtTop > .arrow {
-        top: 100%;
+        top: calc(100% - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtBottom > .arrow {
-        bottom: 100%;
+        bottom: calc(100% - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtLeft > .arrow {
-        left: 100%;
+        left: calc(100% - #{theme.$arrow-safe-rendering-overlap});
     }
 
     .isRootAtRight > .arrow {
-        right: 100%;
+        right: calc(100% - #{theme.$arrow-safe-rendering-overlap});
     }
 
     // Side alignments: top
     .isRootAtTop.isRootAtCenter {
         left: 50%;
-        transform: translate(-50%, #{-1 * theme.$arrow-height});
+        transform: translate(-50%, calc(-1 * #{theme.$arrow-height}));
     }
 
     .isRootAtTop.isRootAtStart {
         left: 0;
-        transform: translate(0, #{-1 * theme.$arrow-height});
+        transform: translate(0, calc(-1 * #{theme.$arrow-height}));
     }
 
     .isRootAtTop.isRootAtEnd {
         right: 0;
-        transform: translate(0, #{-1 * theme.$arrow-height});
+        transform: translate(0, calc(-1 * #{theme.$arrow-height}));
     }
 
     .isRootAtTop.isRootAtCenter > .arrow {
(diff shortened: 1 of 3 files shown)


src/components/Popover/Popover.module.scss
```diff
@@ -68,19 +68,19 @@
 
     // Arrows
     .isRootAtTop > .arrow {
-        top: 100%;
+        top: calc(100% - 1px);
```
adamkudrna, line 71: Creating a variable (or a constant) makes the code easier to understand:

```suggestion
        top: calc(100% - #{$arrow-safe-overlap});
```

And in the `_theme.scss`:

```scss
$arrow-safe-overlap: 1px;
```

src/components/Popover/_theme.scss
```diff
@@ -11,6 +11,7 @@ $color: var(--rui-Popover__color);
 $background-color: var(--rui-Popover__background-color);
 $box-shadow: var(--rui-Popover__box-shadow);
 
-$arrow-width: 1rem;
+// 1 rem (16px) + 2px (for insetting the arrow)
+$arrow-width: 18px;
```
adamkudrna, line 15: You can actually use the calculation (and keep the rem units) and delete the comment:

```suggestion
$arrow-width: calc(1rem + #{$arrow-safe-overlap + $arrow-gap});
```

Plus:

```scss
$arrow-safe-overlap: 1px;
$arrow-gap: 1px;
```

adamkudrna: BTW, the one extra pixel creates this gap between the arrow and the parent element which is something I also wanted to add. Thank you! 👍🏻

image

adamkudrna (commented):
Thank you for the fix! I'd like to ask for a few minor adjustments to improve code readability and keep the existing flexibility (rem units).

src/components/Popover/Popover.module.scss
adamkudrna: Maybe the name could be `$arrow-safe-rendering-overlap` so it's even clearer. Something like that.

src/components/Popover/Popover.module.scss
```diff
@@ -68,35 +68,35 @@
 
     // Arrows
     .isRootAtTop > .arrow {
-        top: 100%;
+        top: calc(100% - #{theme.$arrow-safe-rendering-overlap + theme.$arrow-gap});
```
adamkudrna, line 71: We only want to shift the arrow by `$arrow-gap`. Otherwise the resulting gap is 2 px (with current theme values):

obrazek
⬆️ 2px gap here

```suggestion
        top: calc(100% - #{theme.$arrow-gap});
```

src/components/Popover/_theme.scss
```diff
@@ -11,6 +11,8 @@ $color: var(--rui-Popover__color);
 $background-color: var(--rui-Popover__background-color);
 $box-shadow: var(--rui-Popover__box-shadow);
 
-$arrow-width: 1rem;
-$arrow-height: math.div($arrow-width, 2); // 1.
+$arrow-safe-rendering-overlap: 1px;
+$arrow-gap: 1px;
+$arrow-width: calc(1rem + #{$arrow-safe-rendering-overlap + $arrow-gap});
```
adamkudrna, line 16: On the other hand, only `$arrow-safe-rendering-overlap` needs to be here:

```suggestion
$arrow-width: calc(1rem + #{$arrow-safe-rendering-overlap});
```

src/components/Popover/_theme.scss
```diff
@@ -11,6 +11,8 @@ $color: var(--rui-Popover__color);
...
+$arrow-height: calc(math.div($arrow-width, 2)); // 1.
```
adamkudrna, line 17: The CSS division syntax does not work here?

```suggestion
$arrow-height: calc(#{$arrow-width} / 2); // 1.
```

adamkudrna (changes requested):
OK, this took me some time to figure out, but I think I found the right calculations for what we need. You can test it with various values, for example 2 px gap and 8 px overlap or whatsoever.

src/components/Popover/_theme.scss
atmelmicro: If i just do 
```css
$arrow-width: calc(1rem + #{$arrow-safe-rendering-overlap});
```
Then the arrow shrinks as I increase the arrow-gap
Snímek obrazovky 2024-09-02 v 10 28 27
Snímek obrazovky 2024-09-02 v 10 28 19

What works is 
```css
$arrow-width: calc(1rem + #{$arrow-safe-rendering-overlap + ($arrow-gap * 2)});
```

adamkudrna: OK, let's try it this way.

src/components/Popover/Popover.module.scss
```diff
@@ -68,35 +68,35 @@
 
     // Arrows
     .isRootAtTop > .arrow {
-        top: 100%;
+        top: calc(100% - #{theme.$arrow-gap});
```
adamkudrna, line 71: I found out the advanced positioning is now broken. It is not related to this PR and it was already broken before, but if you could fix it here, it would be great. Otherwise let's file a separate issue for it.

obrazek

atmelmicro: This is because `transferProps` doesn't allow passing of `style`.  Wrapping the popover in a `div` like 

```jsx
{isPopoverOpen && (
  <div
    style={{
      position: strategy,
      top: y ? y : '',
      left: x ? x : '',
    }}
    ref={floating}
  >
    <Popover
      id="my-advanced-popover"
      placement={finalPlacement}
    >
      Auto-flipping Popover
    </Popover>
  </div>
)}
```
doesn't work because the wrapper `div` has dimensions of 0x0 so the popover doesn't switch sides as it should.
Snímek obrazovky 2024-09-04 v 15 13 39

Not sure what would be the best solution for this.

adamkudrna: OK, I'll find an issue for this. Thanks!

adamkudrna: Created #566.

adamkudrna (approved):
It works! 🎉

atmelmicro:
Snímek obrazovky 2024-09-12 v 14 58 05
Hey, the build failing because of the bundle being too big. Is it safe to merge?

mbohal:
@atmelmicro 

Please do increase the allowed size:
image

Here, as well as in our other codebases, we do not have a strict size increase process. We use the limit as a general guideline to prevent the size from exploding beyond reason as a result of bad development decision or a an error.

So if you, as the change author, consider the size increase justified, then feel free to increase the limit. The limit change will be reviewed during the CR same as any other change.

The size increase in this PR is minimal and is fine by me.

mbohal (approved):

PR #561, opened 2024-08-31
Make `Badge` priority class always present like in all other components

adamkudrna:
⚠️  This change affects output HTML and may break downstream snapshot tests.

### DIFF
--- a/src/components/Badge/Badge.jsx
+++ b/src/components/Badge/Badge.jsx
@@ -4,6 +4,7 @@ import { withGlobalProps } from '../../provider';
 import { classNames } from '../../utils/classNames';
 import { transferProps } from '../../utils/transferProps';
 import { getRootColorClassName } from '../_helpers/getRootColorClassName';
+import { getRootPriorityClassName } from '../_helpers/getRootPriorityClassName';
 import styles from './Badge.module.scss';
 
 export const Badge = ({
@@ -16,7 +17,7 @@ export const Badge = ({
     {...transferProps(restProps)}
     className={classNames(
       styles.root,
-      priority === 'outline' && styles.isRootPriorityOutline,
+      getRootPriorityClassName(priority, styles),
       getRootColorClassName(color, styles),
     )}
   >
--- a/src/components/Badge/Badge.module.scss
+++ b/src/components/Badge/Badge.module.scss
@@ -20,6 +20,9 @@ $_badge-size: 1.25rem;
         vertical-align: baseline;
         color: var(--rui-local-color);
         border-radius: math.div($_badge-size, 2);
+    }
+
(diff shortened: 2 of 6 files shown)


atmelmicro (dismissed):
`src/components/Button/helpers/getRootPriorityClassName.js` is not used anymore, afaik it is safe to delete

---

Snímek obrazovky 2024-09-03 v 13 15 58
Filled badges do not have the priority class

bedrich-schindler (approved):

src/components/Badge/Badge.jsx
```diff
@@ -16,7 +17,7 @@ export const Badge = ({
     {...transferProps(restProps)}
     className={classNames(
       styles.root,
-      priority === 'outline' && styles.isRootPriorityOutline,
+      getRootPriorityClassName(priority, styles),
```
bedrich-schindler, line 20: CSS classes for priority=filled are missing in Badge.module.scss.

bedrich-schindler (changes requested):

src/components/Badge/Badge.jsx
bedrich-schindler: I got confused as @atmelmicro. While CSS classes are missing in this pull request, their are added in following pull request: https://github.com/react-ui-org/react-ui/pull/562

@adamkudrna, if you can, leave the information about it somewhere (probably in description). To be honest, it should be one PR as this leave unfinished feature in the code.

Closing comment and approving PR.

bedrich-schindler (approved):

src/components/Badge/Badge.jsx
adamkudrna: Of course you are right, sorry, I didn't notice it. I tried to split my work into as many PRs as possible and this thing leaked 🙂.

BTW, moments like this make me feel we'd appreciate visual tests here in RUI.

adamkudrna: **UPDATE:** OK, I checked the rendered output and guess what, it looks correct 🎉. So what's happening here is that calling of a non-existing class name in CSS modules is not reported. When I think of it, I should have added the priority class check to tests and then I would know. 🤔 

obrazek

Anyway, thanks to both of you for catching this. I'll fix it and add a test (and then probably use my admin power to merge it so the PR cascade gets smaller).

adamkudrna: Didn't find anything, maybe the classes are just resolved differently in the test environment (which sounds terribly), I don't know 🤷🏻‍♂️.

Anyway, I added the missing class to CSS.

bedrich-schindler: If tests does not fail when they should, consider creating issue to look at it.

PR #563, opened 2024-08-31
Generate styles of `Alert`, `Button` and `Card` color variants using collections

adamkudrna:


atmelmicro (approved):

bedrich-schindler (approved):
This looks terrific! Hard to review, but I hope it is all right.

adamkudrna:
> Hard to review

From my experience, adding unit tests for Sass mixins and functions helps. We use the [`sass-true`](https://github.com/oddbird/true) library for unit tests in Spirit. If we agree, I'll be happy to add it here (and in the downstream projects).

PR #565, opened 2024-09-02
Inherit link color in components using the Feedback color collection (#492)

### LINKED ISSUE #492: Inherit link color in components with feedback/validation states
When we use the `<TextLink>` component inside an `<Alert>` element, it uses the context color (from `<Alert>`) instead of the default link color (from `<TextLink>`).

image

However in other contexts this is not the case for other elements:

image

We need to somehow unify this behavior across all our components.

adamkudrna:
⚠️ BREAKING CHANGES:

- For `Alert` and `Card`, link color changes by component variant.
- For all input components, link color changes by validation state.

See #492.

### DIFF
--- a/src/components/TextLink/README.md
+++ b/src/components/TextLink/README.md
@@ -64,13 +64,20 @@ helps to improve its accessibility.
 
 ## Theming
 
+ℹ️ The TextLink component is context-aware and can inherit text color from its
+parent component. This applies for components using
+[Feedback color collection](/docs/foundation/collections#colors) and for
+components in any of the supported
+[validation states](/docs/foundation/colors#validation-states).
+In such cases, the custom properties marked with an asterisk (\*) are ignored.
+
 | Custom Property                           | Description                         |
 |-------------------------------------------|-------------------------------------|
-| `--rui-TextLink__color`                   | Text color                          |
+| `--rui-TextLink__color` \*                | Text color                          |
 | `--rui-TextLink__text-decoration`         | Text decoration, e.g. underline     |
-| `--rui-TextLink--hover__color`            | Text color on hover                 |
+| `--rui-TextLink--hover__color` \*         | Text color on hover                 |
 | `--rui-TextLink--hover__text-decoration`  | Text decoration on hover            |
-| `--rui-TextLink--active__color`           | Text color in the active state      |
+| `--rui-TextLink--active__color` \*        | Text color in the active state      |
 | `--rui-TextLink--active__text-decoration` | Text decoration in the active state |
 
 [a-attributes]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes
--- a/src/styles/tools/_collections.scss
+++ b/src/styles/tools/_collections.scss
@@ -1,6 +1,47 @@
+@use "sass:list";
+@use "sass:map";
+@use "../settings/collections";
 @use "string" as rui-string;
 
-// Mixin to generate CSS custom properties for a set of visual properties.
+// Function to get the parent collection category by value.
+//
+// 1. Returns **only the first** matching collection category.
+//
+// @param {String} $value - The value to get the category for.
+// @param {Map} $collections - The collections map to search in.
+
+@function _get-category-by-value($value, $collections) {
+    @each $category, $values in $collections {
+        @if list.index($values, $value) {
+            @return $category; // 1.
+        }
+    }
+
+    @error
+        "Supplied value \""
+        + $value
+        + "\" not found in any category ("
+        + map.keys($collections)
+        + ")";
+}
+
+// Function to get the matching link color for a component variant.
+//
+// @param {String} $value - The value to get the link color for.
+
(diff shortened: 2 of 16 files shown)


src/styles/tools/_collections.scss
```diff
@@ -72,6 +97,38 @@
...
+//
+// --rui-local-link-color: var(--rui-color-feedback-success);
+// --rui-local-link-color-hover: var(--rui-color-feedback-success-hover);
+// --rui-local-link-color-active: var(--rui-color-feedback-success-active);
+
+@mixin generate-link-properties($prefix, $variant-value) {
+    $color-category: _get-category-by-value($value: $variant-value, $collections: collections.$colors);
+
+    // 1.
+    // TODO: Figure out what to do with the `dark` color category where blue links do not work.
```
adamkudrna, line 124: This seems to be an issue. As stated in #492, we want to keep the basic appearance for links inside components in the Neutral Light color. However, Neutral Dark is unreadable then.

I have to think about it yet. 🤔 

obrazek

bedrich-schindler: Personal opinion: Inherit color for neutral dark color (and keep only neutral light color) OR split link color property/ies between light and dark variant/s. This is beyond this discussion, but if we generally split colors between light and dark, couldn't we use it later for dark mode variant (which we haven't talk about it yet)?

adamkudrna: Light and dark colors (the neutral colors collection) and light and dark modes are something different. But if you mean we would face the same problem during dark mode design, you are probably right. Just with the difference that light and dark neutral colors will be both somewhat dark, just using a different shade of gray.

When I'm thinking about it from the design perspective, it makes sense to use the default link appearance in light color variants. Even if it means an exception in our code.

(Other way would be adding even more design tokens (= custom properties) which sounds like an overkill…)

adamkudrna: Maybe we can have another exception for `Alert`s where dark text would work well 😄 

obrazek

adamkudrna: I decided to keep the exception only for links inside `Card`.

> ℹ️ To meet users' expectation, links in the light neutral variant of Card use the [default link color](http://localhost:8000/docs/foundation/colors/#text-colors). In other variants, links are colored according to the variant.

image
image

mbohal: I think that for  `Alert` and `Card` we should either use default link color in both in `light` and `dark` versions or we should use default link color for `light` and text color for `dark` versions.

Unifying the behavior across components would make it much easier to grasp and it is a better DX. Tweaking the colors and boldness to make it work can then be performed by each RUI based project.

Maybe, if possible, we can tweak the link color to make it work in both color contexts.

---


Presently the solution with different visualisation of `Alert` and `Card` works because `Alert` shows link with emphasis. I think this is something we want to be able to turn of, that is `Alert` should accept along with `--rui-Alert__emphasis__font-weight` also something like `--rui-Alert__link__font-weight`. Maybe someone does not like link-as-emphasis approach.

This would allow us to render `Alert` with normal weight links. Then there would be no reason to have differently link color between `Alert` and `Card`.

src/components/TextLink/README.md
```diff
@@ -64,13 +64,20 @@ helps to improve its accessibility.
 
 ## Theming
 
+ℹ️ The TextLink component is context-aware and can inherit text color from its
+parent component. This applies for components using
+[Feedback color collection](/docs/foundation/collections#colors) and for
+components in any of the supported
+[validation states](/docs/foundation/colors#validation-states).
```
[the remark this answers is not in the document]
bedrich-schindler: You are right. It does not work for Input components. What more, we should consider adding support for color inheritance for `label` props as we potentially can pass link as label. It is not what we encourage, but we should still support that. I can imagine we would need this to show popover.

src/styles/tools/_collections.scss
adamkudrna: > Maybe, if possible, we can tweak the link color to make it work in both color contexts.

I tried and I don't see a way here…

> Then there would be no reason to have differently link color between Alert and Card.

I changed it and there is currently no exception. It looks like this which we know is not ideal, so I filed a followup #577 for it. But the approach is the same everywhere now which is what we want.

image
image
image

src/components/TextLink/README.md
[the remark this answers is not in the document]
adamkudrna: OK, now it's there. It works for both label and validation text. I added links into labels and validation text in a few components.
 
Link underline will be reviewed in #577 (the underline you see on the screenshots below is on hover).

image
image
image

mbohal:
I'm going to give approve , but honestly I'm unsure if we should merge this without resolving https://github.com/react-ui-org/react-ui/issues/577 .

mbohal (approved):

adamkudrna:
> I'm going to give approve , but honestly I'm unsure if we should merge this without resolving #577 .

Just merged #577 into this branch, so I'm merging this one as well.

PR #568, opened 2024-09-17
Fix layout context prop resolving for boolean types (#567)

### LINKED ISSUE #567: Layout context not working for bool props
**Usecase:**

Im trying to get an `InputGroup` with one child element disabled and one not:

```jsx
<InputGroup>
  <TextField
    label="First name"
    disabled
  />
  <TextField
    label="Last name"
  />
</InputGroup>
```

**Problem:**

In the above example:

Current behaviour:  none of the `TextField`s is disabled
Expected behaviour: The `First name`  `TextField` is disabled

Likely cause is https://github.com/react-ui-org/react-ui/blob/d532592001380c4c88a0f92d637e1b6bfbead5ce/src/components/_helpers/resolveContextOrProp.js#L2 where the `contextProp` is checked for being `null` / `undefined` but not against being `false`.

**Solution:**

We need to figure out how to take into account the default prop value in https://github.com/react-ui-org/react-ui/blob/d532592001380c4c88a0f92d637e1b6bfbead5ce/src/components/_helpers/resolveContextOrProp.js#L2. Assuming the default is `null` / `undefined` does not work for bool typed props.

atmelmicro:
Allows boolean props to propagate when layout context is false

Closes #567

### DIFF
--- a/src/components/_helpers/resolveContextOrProp.js
+++ b/src/components/_helpers/resolveContextOrProp.js
@@ -1,7 +1,10 @@
 export const resolveContextOrProp = (contextValue, propValue) => {
-  if (contextValue != null) {
-    return contextValue;
+  // We need to test:
+  //  * `false` - for when the `contextValue` is boolean
+  //  * `null` - for when the `contextValue` is non-boolean
+  if (contextValue === false || contextValue === null) {
+    return propValue;
   }
 
-  return propValue;
+  return contextValue;
 };


src/components/_helpers/resolveContextOrProp.js
```diff
@@ -1,4 +1,8 @@
 export const resolveContextOrProp = (contextValue, propValue) => {
+  if (contextValue === false) {
```
mbohal, line 2: Could you please rewrite it so there are only two returns?

```
if (…) {
  return a;
}

return b;
```

Having three returns makes it hard to follow the flow of the code. I know this function is short, but still…

mbohal (changes requested):

src/components/_helpers/resolveContextOrProp.js
adamkudrna, line 1: Does it make any sense to have a test for it? Possibly with an example similar to that in the original issue #567.

mbohal: I think Ill leave it up to @atmelmicro to decide.

It is one simple condition, so I think that maybe just a comment would work:

```js
// We need to test:
//  * `false` - for when the `contextValue` is boolean
//  * `null` - for when the `contextValue` is non-boolean
if (contextValue === false || contextValue === null) {
```

However test would communicate it just as good and it would test it on top.

I have no preference, really, but I agree with @adamkudrna that one or the other should be added.

atmelmicro: In my opinion the comment is more descriptive, it describes the why, not just how it should work.

mbohal (approved):

bedrich-schindler (approved):

PR #573, opened 2024-10-21
Use Node 22 (current LTS) for development

adamkudrna:
- Drop Node 18 for testing builds.
- Add Node 22.x to dev engines.


mbohal (approved):

bedrich-schindler:
I would update package.json in such way, that I would remove `engines` section and introduce `devEngines`:

```
  "devEngines": {
    "node": "20.x || 22.x"
  },
  ```
  
  This is how React does it. RUI does not required any specific version to run, only dev environment does.
  
  What do you think @adamkudrna @mbohal ?

**Comment by mbohal** [comment]
I agree with using `devEngines`. However the code example provided by @bedrich-schindler is wrong.

See: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#devengines for usage. I wonder if we should fail or warn on error. I think I would prefer to fail…

**Comment by bedrich-schindler** [comment]
@mbohal good point. It is weird that library like React have it wrong, but you are right unless there is some undocumented approach.

File: package.json [hunk 1]
```diff
@@ -31,6 +31,17 @@
...
+    "runtime": {
+      "name": "node",
+      "version": "22.x",
+      "onFail": "error"
+    },
+    "packageManager": {
+      "name": "npm",
+      "onFail": "error"
+    }
+  },
```
adamkudrna, line 44: @bedrich-schindler @mbohal What do you think?

.github/workflows/lint.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
```
adamkudrna, line 16: Update commit message from "upcoming" to "current".

package.json
adamkudrna: Approved by @bedrich-schindler.

PR #576, opened 2024-11-19
Fix misplaced Popover when using Floating UI (#566)

### LINKED ISSUE #566: Misplaced `Popover` when used with Floating UI
> I found out the advanced positioning is now broken. It is not related to this PR and it was already broken before, but if you could fix it here, it would be great. Otherwise let's file a separate issue for it.

Image

https://github.com/react-ui-org/react-ui/pull/559#discussion_r1743774576

atmelmicro:
Closes #566

### DIFF
--- a/src/components/Popover/README.md
+++ b/src/components/Popover/README.md
@@ -162,6 +162,29 @@ automatically, including smart position updates to ensure Popover visibility,
 we recommend to involve an external library designed specifically for this
 purpose.
 
+To position the popover, you need to provide the `placementStyle` prop with the
+style you want to apply to the popover. This prop should only be used to
+position the popover. The allowed props are:
+
+- `position`
+- `inset`
+- `inset-inline-start`
+- `inset-inline-end`
+- `inset-block-start`
+- `inset-block-end`
+- `top`
+- `right`
+- `bottom`
+- `left`
+- `translate`
+- `transform-origin`
+
+⚠️ [`inset`][mdn-inset] is a shorthand for `top right bottom left`, not for
+`inset-*` properties.
+
+As opposed to `top right bottom left` and the `inset` shorthand, `inset-*`
+properties are writing-direction aware.
+
 ℹ️ The following example is using external library [Floating UI]. To use
 Floating UI, install it first:
 
@@ -267,10 +290,10 @@ React.createElement(() => {
             <Popover
               id="my-advanced-popover"
               placement={finalPlacement}
-              style={{
+              placementStyle={{
                 position: strategy,
-                top: y ? y : '',
-                left: x ? x : '',
+                top: `${y}px`,
+                left: `${x}px`,
               }}
               ref={floating}
             >
@@ -359,5 +382,6 @@ which enables [Advanced Positioning](#advanced-positioning).
 
 [div-attributes]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#attributes
 [Floating UI]: https://floating-ui.com/docs/react-dom
+[mdn-inset]: https://developer.mozilla.org/en-US/docs/Web/CSS/inset
 [React common props]: https://react.dev/reference/react-dom/components/common#common-props
 [ref]: https://reactjs.org/docs/refs-and-the-dom.html
--- a/src/components/Popover/_helpers/cleanPlacementStyle.js
+++ b/src/components/Popover/_helpers/cleanPlacementStyle.js
@@ -1,0 +1,20 @@
+export default (placementStyle) => {
+  const validProps = [
+    'position',
+    'inset',
+    'inset-inline-start',
+    'inset-inline-end',
+    'inset-block-start',
+    'inset-block-end',
+    'top',
+    'right',
+    'bottom',
+    'left',
+    'translate',
+    'transform-origin',
+  ];
+
+  return Object.fromEntries(
+    Object.entries(placementStyle).filter(([prop]) => validProps.includes(prop)),
+  );
+};
--- a/src/components/Popover/Popover.jsx
+++ b/src/components/Popover/Popover.jsx
@@ -4,6 +4,7 @@ import { createPortal } from 'react-dom';
 import { withGlobalProps } from '../../providers/globalProps';
 import { classNames } from '../../utils/classNames';
 import { transferProps } from '../../utils/transferProps';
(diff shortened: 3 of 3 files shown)


mbohal (approved):
:+1:

bedrich-schindler (changes requested):
In the end, after Monday's discussion, I do not see this solution satisfying. 

Floating UI is just example we use just because Floating UI is currently trending library that solves positioning. But we are open to every implementation user decides to use. This means that we are unable to determine attributes user would need to pass, position, x, y are attributes that are necessary for our example but might not be for production use.

I want @adamkudrna to decide this together with us. I would personally suggest to implement something like `style`. I would call it e.g. `placementStyle`.

`placementStyle` could be either just the way how to bypass styles (`style={placementStyle}`) or it can have black or white list or style attributes to pass.

I know that React 19 will get rid of `forwardRef`, but I see this concept of forward-ref (respectively forward-styles) useful as we can control it for the component we know it use useful. 

This is my proposal.

I can also imagine that we would require to pass either `placement` or `placementStyle`, so that only one can be set at the time. But this is just question arising in my head.

mbohal:
> But we are open to every implementation user decides to use.

I agree with @bedrich-schindler that we do not want to embrace Floating UI to closely.

I think the question is if some other placement implementation is likely to express position differently then `x`, `y` and `position`. I don't feel versed enough in CSS to decide this. I would think it less likely, but I'm don't really know.

> I can also imagine that we would require to pass either placement or placementStyle, so that only one can be set at the time. But this is just question arising in my head.

So you mean to have one new prop:

```
placement: PropTypes.shape({
    /**
   * This sets the CSS property `position` of the popover. This is reserved for use with Floating UI.
   */
  position: PropTypes.string,
  /**
   * This sets the CSS property `top` of the popover. This is reserved for use with Floating UI.
   */
  x: PropTypes.string,
  /**
   * This sets the CSS property `left` of the popover. This is reserved for use with Floating UI.
   */
  y: PropTypes.string,
})
```

so we can then easily switch between placement formats?

adamkudrna:
- I can imagine positioning solved via CSS transforms (`translateX`, `translateY` properties). I think the new drag-and-drop library we use does it this way as it is more performant and fluent.
- Can we have something like adapters for supported positioning libraries? We would provide the Floating UI adapter and anyone could plug-in their own.

mbohal:
> Can we have something like adapters for supported positioning libraries? We would provide the Floating UI adapter and anyone could plug-in their own.

So something like:

```
getPlacement: PropTypes.func,
```

where the signature of Floating UI implementation would be:
```
getPlacement(x: number, y: number, position: string): string
```

with the return value being the style HTML attribute value?
I like that.

mbohal:
On call we agreed that the best solution is from @bedrich-schindler

> placementStyle could be either just the way how to bypass styles (style={placementStyle}) or it can have black or white list or style attributes to pass.

We would like to filter the style attributes to allow all properties that are neede for:

* absolute positioning
* CSS transform

We need to reflect this in the docs.

adamkudrna:
In case a general positioning white-list of relevant CSS properties is needed, this is what I can think of:

```
position
inset
inset-inline-start
inset-inline-end
inset-block-start
inset-block-end
top
right
bottom
left
translate
transform-origin
```

Notes (could be mentioned in the docs):

- ⚠️ `inset` is a shorthand for `top right bottom left`, not for `inset-*` properties.
- as opposed to `top right bottom left` and the `inset` shorthand, `inset-*` properties are writing-direction aware.

https://developer.mozilla.org/en-US/docs/Web/CSS/inset

I am leaving out these:

- `inset-inline`, `inset-block`: if a shorthand is needed, I would go with `inset`.
- `transform`: it allows to set any transforms, not only x/y translation. That seems to be too powerful to me.

src/components/Popover/README.md
```diff
@@ -162,6 +162,28 @@ automatically, including smart position updates to ensure Popover visibility,
 we recommend to involve an external library designed specifically for this
 purpose.
 
+To position the popover, you need to provide the `placementStyle` prop with the
+style you want to apply to the popover. This prop is filtered and should only
```
mbohal, line 166: I think we should reword:

> This prop is filtered and should only …

to:

> This prop should only …

The code does not filter anything. It merely shows warning and error when in dev mode.  I find this sufficient, but the docs should match the code.

src/components/Popover/_helpers/cleanPlacementStyle.js
```diff
@@ -0,0 +1,29 @@
...
+    'inset-block-end',
+    'top',
+    'right',
+    'bottom',
+    'left',
+    'translate',
+    'transform-origin',
+  ];
+
+  if (process.env.NODE_ENV !== 'production') {
```
mbohal, line 17: In dev mode the allowed styles are already handled by propType validation. I dont think we need this explicit check at all.

If we do, the errors in console caused by `bject.keys(placementStyle)` when `placementStyle === null` need to be fixed,

mbohal (changes requested):

src/components/Popover/README.md
```diff
@@ -162,6 +162,23 @@ automatically, including smart position updates to ensure Popover visibility,
...
+be used to position the popover. The allowed props are
```
adamkudrna, line 167:
```suggestion
be used to position the popover. The allowed props are:
```

adamkudrna (commented):
I tried to test it in the browser and the Floating UI example works, but the others above are all broken.

src/components/Popover/_helpers/cleanPlacementStyle.js
atmelmicro: The null fix is still needed because Object.entries(null) fails. I have implemented it like this. Not sure if this this the correct way

```jsx
style={cleanPlacementStyle(placementStyle ?? {})}
```

adamkudrna (approved):

mbohal (approved):

src/components/Popover/_helpers/cleanPlacementStyle.js
mbohal: I would rather that in such we do not call the `cleanPlacementStyle` function at all. Maybe something like:

`{placementStyle ? (...cleanPlacementStyle(placementStyle)) : null}`

bedrich-schindler (approved):

PR #580, opened 2024-12-02
Make text links underlined by default #577

### LINKED ISSUE #577: Review link styles
Ideas:

1. use regular font weight for links in `Alert`s
2. underline links by default so they are distinguishable from regular text in neutral `Card`s

adamkudrna:
## 1. ~~Allow configuration~~ Remove configurable font weight of links inside `Alert` #577

~~Default font weight of links is now set to "regular" so it is consistent with other links in React UI. However, it remains configurable because of downstream project requirements.~~

Links in `Alert` now inherit the font weight which is consistent with other links in React UI.

## 2. Make text links underlined by default #577

Since the color inheritance for links has been introduced in #492, links in e.g. validation messages cannot be distinguished from regular text until hovered. We chose to underline all text links by default to make those links discoverable by users.

Closes #577

### DIFF
--- a/src/components/Alert/README.md
+++ b/src/components/Alert/README.md
@@ -201,14 +201,14 @@ helps to improve its accessibility.
 
 ### Common Theming Options
 
-| Custom Property                                      | Description                                                  |
-|------------------------------------------------------|--------------------------------------------------------------|
-| `--rui-Alert__padding`                               | Padding between border and message                           |
-| `--rui-Alert__font-weight`                           | Message font weight                                          |
-| `--rui-Alert__border-width`                          | Border width                                                 |
-| `--rui-Alert__border-radius`                         | Corner radius                                                |
-| `--rui-Alert__emphasis__font-weight`                 | Font weight of text emphasised with `<strong>`               |
-| `--rui-Alert__stripe__width`                         | Width of the border at the start of the Alert                |
+| Custom Property                      | Description                                    |
+|--------------------------------------|------------------------------------------------|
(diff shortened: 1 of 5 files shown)


src/components/Alert/README.md
```diff
@@ -201,14 +201,15 @@ helps to improve its accessibility.
...
-| `--rui-Alert__emphasis__font-weight`                 | Font weight of text emphasised with `<strong>`               |
-| `--rui-Alert__stripe__width`                         | Width of the border at the start of the Alert                |
+| Custom Property                      | Description                                    |
+|--------------------------------------|------------------------------------------------|
+| `--rui-Alert__padding`               | Padding between border and message             |
+| `--rui-Alert__font-weight`           | Message font weight                            |
+| `--rui-Alert__border-width`          | Border width                                   |
+| `--rui-Alert__border-radius`         | Corner radius                                  |
+| `--rui-Alert__emphasis__font-weight` | Font weight of text emphasised with `<strong>` |
+| `--rui-Alert__link__font-weight`     | Font weight of links inside Alert              |
```
mbohal, line 211: I dont think we need this. See Slack message in `#dev` channel.

adamkudrna: OK, removing the option.

bedrich-schindler: I am leaving this decision up to @adamkudrna. I don't care about this. If we do not need it, I would remove it. But I do not care. Implementation is correct.

bedrich-schindler (approved):

PR #581, opened 2024-12-03
Allow check fields and selectable fields to render as required #487

### LINKED ISSUE #487: implicitly required form fields visual
For some form fields:

* Those representing boolean value (`Toggle`, `CheckboxField`) 
* Those representing  value from list (`SelectField`, `Radio`) 

The semantics of `required` are different then for fields such as `TextField`. When `TextField` is left empty, it means the user did not specify value. When one of the above mentioned special fields is left untouched, it still carries a meaning (e.g. `false`).

The above mentioned special fields are therefore always required as it is impossible not to fill them. Accordingly, they should always be styled as required fields even when they do not have the `required` attribute.

The `required` attribute on boolean fields means "must be set to true" which is different from its meaning when used with for example a `TextField`. When the `required` attribute is used in this sense, it must be communicated by other means, (e.g. by adding (required) to the label or such).

Note that HTML `<select multiple>`  can be left with no selected value, but RUI does not implement this.

adamkudrna:
Users may find themselves in a situation where the input is not required (i.e. making the input checked), but they also don't want to render the field as optional because not choosing an option can be perfectly valid. For this case, there is the `renderAsRequired` prop.

This affects `CheckboxField`, `Radio`, `SelectField`, and `Toggle`.

Closes #487.

### DIFF
--- a/src/components/CheckboxField/README.md
+++ b/src/components/CheckboxField/README.md
@@ -186,6 +186,81 @@ React.createElement(() => {
 });
 ```
 
+### Required State
+
+The required state indicates that the input is mandatory. Required fields
+display an asterisk `*` after the label by default.
+
+```docoff-react-preview
+React.createElement(() => {
+  const [agree, setAgree] = React.useState(true);
+  return (
+    <CheckboxField
+      checked={agree}
+      label="I agree"
+      onChange={() => setAgree(!agree)}
+      required
+    />
+  );
+});
+```
+
+#### Styling the Required State
+
+All form fields in React UI can be
+[styled](/docs/customize/theming/forms/#required-state)
+to indicate the required state.
+
+However, you may find yourself in a situation where a form field is valid in
+both checked and unchecked states, for example to turn on or off a feature.
+If your project uses the label color as the primary means to indicate the
+required state of input fields and the usual asterisk `*` is omitted, you may
+want to keep the label color consistent for both states to avoid confusion.
+
+For this edge case, there is the `renderAsRequired` prop:
+
+```docoff-react-preview
+React.createElement(() => {
+  const [optional, setOptional] = React.useState(false);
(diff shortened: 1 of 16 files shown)


bedrich-schindler:
> I think the feature is good (we need it and we know why), but honestly the docs are very confusing to me.
> 
> I think we should talk about it before merging - I will schedule this for our Monday meeting.

I am approving this as it is fine for me. But I might be influenced by the reason for this change.

But it is true that one more sentence could be helpful. I am missing the detailed explanation why would user might need this and why it is correct approach.

bedrich-schindler (approved):

src/components/CheckboxField/README.md
```diff
@@ -186,6 +186,76 @@ React.createElement(() => {
 });
 ```
 
+### Required State
```
adamkudrna, line 189: > But it is true that one more sentence could be helpful. I am missing the detailed explanation why would user might need this and why it is correct approach.

I rewrote the docs and reworked the example. What do you guys think now?

obrazek

adamkudrna: Suggestions:

- emphasize it's just a visual thing
- when a checkbox can be true or false, not only true
- consider moving it aside as an edge case

mbohal: I like it.

mbohal: I think we need to port this wording to the other Components (`Radio`, `SelectField` and `Toggle`)

adamkudrna: Done. Now I think it's good, finally.

obrazek

Plus in the forms theming docs:

obrazek

mbohal (approved):

atmelmicro (approved):

PR #582, opened 2024-12-03
Add controlled popover

atmelmicro:
Using the Popover API with the help of a helper element this implements a native way to create a popover that closes and opens on button and backdrop click.

### DIFF
--- a/src/components/Popover/Popover.jsx
+++ b/src/components/Popover/Popover.jsx
@@ -12,24 +12,40 @@ export const Popover = React.forwardRef((props, ref) => {
   const {
     placement,
     children,
+    popoverTargetId,
     portalId,
     ...restProps
   } = props;
 
   const PopoverEl = (
-    <div
-      {...transferProps(restProps)}
-      className={classNames(
-        styles.root,
-        ref && styles.isRootControlled,
-        getRootSideClassName(placement, styles),
-        getRootAlignmentClassName(placement, styles),
+    <>
+      {/**
+        * This hack is needed because the default behavior of the Popover API is to place the popover into a
+        * top-layer. It is currently not possible to position an element in the top-layer relative to a normal element.
+        * This will create a hidden browser popover, then with CSS it will open and close the RUI popover.
+        */}
+      {!!popoverTargetId && (
+        <div
+          className={styles.helper}
+          id={popoverTargetId}
+          popover="auto"
+        />
       )}
-      ref={ref}
-    >
-      {children}
-      <span className={styles.arrow} />
-    </div>
+      <div
+        {...transferProps(restProps)}
+        className={classNames(
+          styles.root,
+          ref && styles.isRootControlled,
+          popoverTargetId && styles.controlledPopover,
+          getRootSideClassName(placement, styles),
+          getRootAlignmentClassName(placement, styles),
+        )}
+        ref={ref}
+      >
+        {children}
+        <span className={styles.arrow} />
+      </div>
+    </>
   );
 
   if (portalId === null) {
@@ -41,6 +57,7 @@ export const Popover = React.forwardRef((props, ref) => {
 
 Popover.defaultProps = {
   placement: 'bottom',
+  popoverTargetId: null,
   portalId: null,
 };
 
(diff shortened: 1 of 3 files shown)


src/components/Popover/Popover.jsx
```diff
@@ -67,6 +73,12 @@ Popover.propTypes = {
     'left-start',
     'left-end',
   ]),
+  /**
+   * If set, the popover will become controlled, meaning it will be hidden by default and will need a trigger to open.
+   * This sets the ID of the internal helper element for the popover.
+   * Assign the same ID to `popovertarget` of a trigger to make it open and close.
+   */
+  popoverHelperId: PropTypes.string,
```
mbohal, line 81: I'm unsure if `popoverHelperId` is a good name. I find it too general.

I think I would prefer something like `popoverTargetId`, `popoverAnchorId`, `popoverPlacementHelperId` or such.

I don't feel super strong about it, so I will leave this to others to consider and decide.

src/components/Popover/Popover.jsx
```diff
@@ -12,24 +12,29 @@ export const Popover = React.forwardRef((props, ref) => {
...
-        getRootSideClassName(placement, styles),
-        getRootAlignmentClassName(placement, styles),
-      )}
-      ref={ref}
-    >
-      {children}
-      <span className={styles.arrow} />
-    </div>
+    <>
+      {!!popoverHelperId && <div className={styles.helper} id={popoverHelperId} popover="auto" />}
```
mbohal, line 22: I think we should add a comment explaining why we use this hack and how it works.

src/components/Popover/README.md
```diff
@@ -284,6 +284,39 @@ React.createElement(() => {
 });
 ```
 
+## Controlled Popover
+
+To make it easier to implement the popover in your app, you can set the
```
mbohal, line 289: I would mention, that the clean solution would be `anchor` API, but that it is not available. Maybe a link to MDN.

src/components/Popover/Popover.jsx
bedrich-schindler: Please reformat the code like this:

```jsx
{!!popoverHelperId && (
  <div
    className={styles.helper}
    id={popoverHelperId}
    popover="auto" 
  />
)}
```

src/components/Popover/Popover.module.scss
```diff
@@ -49,6 +49,28 @@
         }
     }
 
+    // Controlled popover
+    .controlledPopover {
+        display: none;
+    }
+
+    .helper {
```
bedrich-schindler, line 57: @atmelmicro, discuss this with @adamkudrna. I think it is necessary to comment this hack at the beginning of the file and add link to it.

src/components/Popover/Popover.jsx
bedrich-schindler: I think we decided to call it `popoverTargetId` so it matches `popoverTarget`.

src/components/Popover/README.md
bedrich-schindler: I would consider enhancing the documentation to something like:

```
Popover API can be used to control visibility of Popover component. You need to set `id` on the trigger element and  matching `popoverHelperId` attribute on the Popover component.  This leverages the browser's Popover API to control the popover, automatically closing it when the trigger or the backdrop is pressed. 
```

I would explicitly mention setting `id` on the target element so it is crystal clear.

bedrich-schindler (changes requested):

src/components/Popover/Popover.module.scss
adamkudrna: I agree.

@atmelmicro Please drop comments in both CSS and JS and explain why we decided to use this hack.

mbohal (approved):

src/components/Popover/Popover.module.scss
bedrich-schindler: @atmelmicro, @adamkudrna uses a bit different approach for commenting SCSS files using references to comments written ahead of the code (at the top of the file). Please follow the convention unless @adamkudrna states anything else.

bedrich-schindler (approved):

bedrich-schindler:
I am approving this PR even though SCSS comments are not done as required. But it is up to @adamkudrna to approve them.

adamkudrna (approved):

PR #585, opened 2025-01-30
Docs/improve class names docs

mbohal:


bedrich-schindler:
1+2. Let's do in in Martin's way.
3. Lets call it `src/helpers`
4. Let's split it in two examples. One of them is simpler than original and the second one demonstrates usage in MDN-like way, e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

mbohal:
@bedrich-schindler it is fixed I think. Please take a look.

atmelmicro (approved):

src/index.js
```diff
@@ -61,6 +61,6 @@ export {
 export { GlobalPropsProvider } from './providers/globalProps';
 export { TranslationsProvider } from './providers/translations';
 
-// Utils
-export { classNames } from './utils/classNames';
-export { transferProps } from './utils/transferProps';
+// JS Helpers
```
bedrich-schindler, line 64: `JS Helpers` -> `Helpers`

src/helpers/classNames/__tests__/classNames.test.js
bedrich-schindler, line 1: Those changes do not make sense to me.

I would like to include all the values in the first test, including `true`,`false`, `undefined` that you omitted just in case the algorithm changes. The rest you added is perfect.

But passing nothing to functions does not make sense at all. It is useless and it does not show real case either. I still want to have test in former way so it can filter out everything and return empty array. If you want to test explicitly 0 parameters, create extra test for that.

src/helpers/classNames/README.md
bedrich-schindler, line 1: There is no change in the documentation.

I personally do not care but I though we all agreed on creating new section documenting JS functions in MDN way. Or have I missed anything?

bedrich-schindler (changes requested):

src/helpers/classNames/README.md
```diff
@@ -0,0 +1,63 @@
...
+    'd-block',
+    new Date('3000-01-01T00:00:00') < new Date() && 'text-danger',
+  )}
+  >
+  {(new Date()).toLocaleDateString()}
+  </div>
+</>
+```
+
+## Parameter filtering
```
adamkudrna, line 40:
```suggestion
## Parameter Filtering
```

adamkudrna (approved):

src/components/Alert/Alert.jsx
```diff
@@ -2,8 +2,8 @@ import PropTypes from 'prop-types';
 import React, { useContext } from 'react';
 import { withGlobalProps } from '../../providers/globalProps';
 import { TranslationsContext } from '../../providers/translations';
-import { classNames } from '../../utils/classNames';
-import { transferProps } from '../../utils/transferProps';
+import { classNames } from '../../helpers/classNames/classNames';
+import { transferProps } from '../../helpers/transferProps/transferProps';
```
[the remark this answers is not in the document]
mbohal: >  I'd prefer to use index.js as we have it everywhere

@adamkudrna  Could you please provide an example? The closes thing I could fine is https://github.com/react-ui-org/react-ui/blob/master/src/components/Modal/index.js which I don't consider a barrel file as the exported components are related.

I think this is the first example in RUI where we are exporting truly unrelated code from one folder.

So I would prefer to leave it as is, because I see no convention that it would break.

src/helpers/classNames/__tests__/classNames.test.js
mbohal: > I would like to include all the values in the first test, including true,false, undefined that you omitted j

They are still there:
Screenshot from 2025-02-18 06-42-17

The second paragraph is super confusing to me. Ill try to answer as best as I can.

> But passing nothing to functions does not make sense at all.

I was thinking of something like `classNames(...classArray)` where `classArray` is empty. It is an edge case, but is there a reason why not to test it?

> I still want to have test in former way so it can filter out everything and return empty array.

I don't understand, this function returns a `string`, not an array. I can change it so that it will return a`''`. I think that would be preferable, but it is a breaking change. @bedrich-schindler is that what you had in mind?

>  If you want to test explicitly 0 parameters, create extra test for that.

There is a separate test for that:
image

Maybe it should just be renamed to `returns undefined if called with no params`?

src/index.js
mbohal: fixed

src/helpers/classNames/README.md
mbohal: I took inspiration from this:
image

(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor)

I find it the clearest description of what a function does.
Do you feel I should redo it? I find it quite self explanatory.

src/helpers/classNames/README.md
mbohal: fixed

src/helpers/classNames/README.md
```diff
@@ -0,0 +1,63 @@
+# classNames
```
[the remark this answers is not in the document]
mbohal: done

src/helpers/classNames/README.md
lukasbriza: > I took inspiration from this: image
> 
> (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor)
> 
> I find it the clearest description of what a function does. Do you feel I should redo it? I find it quite self explanatory.

This is exactly the type of function documentation I was referring to in our last call. I’m completely fine with documenting them the way Martin showed in the picture he posted.

lukasbriza (changes requested):
Add MDN-style documentation for the className function.

src/helpers/classNames/README.md
bedrich-schindler: **OK. I thought we were talking about the section describing parameters. If @lukasbriza is OK with that, I am approving this as well.**

Nevertheless, I would like to improve visual for this in the future, because outputting string into place where component should be placed is a bit weird. This would not looks good for advanced functions that can output objects and arrays. We will have to solve it our private design library where more functions can be exported, so we can then upgrade it here.

I also think that each attribute will have to be described in the future when there is more than one argument. See Parameters section on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#parameters

src/helpers/classNames/__tests__/classNames.test.js
bedrich-schindler: This is my fault. Few lines remained hidden 🙈 
image

mbohal: Agreed ona call:

1. I will rename the test
2. I will add a test that has params, but filters all out
3. Both cases will return an empty string

src/helpers/classNames/README.md
mbohal: On call agreed that it is OK as is

lukasbriza (approved):

src/components/Alert/Alert.jsx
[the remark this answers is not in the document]
adamkudrna: > we are exporting truly unrelated code from one folder.

That brings me to question: why is the code in the same folder? 🙂 So, from my perspective, the code _is_ related since it lives in the same folder, and I'm really not fan of the lengthy imports like `'../../helpers/transferProps/transferProps'`.

However, if you prefer to keep it this way and nobody else objects like I do, let's do it your way. Eventually, it's rather your field than mine.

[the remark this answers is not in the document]
mbohal: > why is the code in the same folder? 🙂

Well, `/src` is nothing more then folder full of vaguely related code… :)

I guess I just dislike to have a folder with one file and `index.js`.


> However, if you prefer to keep it this way and nobody else objects like I do, let's do it your way.

Will do.

src/helpers/classNames/__tests__/classNames.test.js
```diff
@@ -20,9 +20,15 @@ describe('classNames', () => {
...
 
-  it('returns undefined if all class names are filtered out', () => {
+  it('returns empty string if called with no params', () => {
     const result = classNames();
 
-    expect(result).toEqual(undefined);
+    expect(result).toEqual('');
+  });
+
+  it('returns empty string if all params are removed', () => {
```
bedrich-schindler, line 29: `... if all params are removed` ... I am still not sure about the existence of this test. `... if all params are removed` does not make sense. I know what do you want to test but if I am user seeing this for the first time, I would be confused and consider this bug. What params are removed? How are removed? By any way, this is not related to function test.

If you want to still test this, better describe the function like `returns empty string if called with only falsy params`

src/helpers/classNames/classNames.js
```diff
@@ -4,5 +4,5 @@ export const classNames = (...classes) => {
     && className.trim().length > 0,
   );
 
-  return filteredClassNames.length > 0 ? filteredClassNames.join(' ') : undefined;
+  return filteredClassNames.length > 0 ? filteredClassNames.join(' ') : '';
```
bedrich-schindler, line 7: Why empty string and not `undefined`? We want `undefined` to completely omit parameter from the DOM. Empty strings will polute snapshots and DOM, thus this would have been marked as BC.

bedrich-schindler (changes requested):

src/helpers/classNames/classNames.js
mbohal: You are right. I thought we agree o this on our call to always return the same type for use with TS.

I will revert it and a comment.

bedrich-schindler: I am not aware of it. But if we did, I am sorry for that.

bedrich-schindler: And quick note, you changed test names as well.

src/helpers/classNames/__tests__/classNames.test.js
mbohal: done

src/helpers/classNames/classNames.js
mbohal: done

src/index.js
```diff
@@ -61,6 +61,6 @@ export {
 export { GlobalPropsProvider } from './providers/globalProps';
 export { TranslationsProvider } from './providers/translations';
 
-// Utils
-export { classNames } from './utils/classNames';
-export { transferProps } from './utils/transferProps';
+// Helpers
+export { classNames } from './helpers/classNames/classNames';
```
mbohal, line 65: `helpers/classNames` and `helpers/transferProps` should have `index.js`.

mbohal: done

Issue #417, opened 2022-08-21
Switch to system fonts


adamkudrna:
The core idea of RUI is its customizability, so…?

PR #588, opened 2025-02-15
Playwright component testing

bedrich-schindler:
I introduced Playwright Component Testing as replacement for Jest and RTL.
    
    Those tests must be run within `playwright` docker container with pre-installed
    browsers to ensure uniform testing environment. `.env.playwright` is created
    using `postinstall` script to follow zero-config setup. `.env.playwright` can
    be used to tween parameters for local development if needed.
    
    Currently, only tests for `Alert` and `Button` components are migrated to
    Playwright CT as this commit is supposed to be proof-of-concept. Rest should
    be migrated in later pull request.
    
    All visual tests use prop tests concept we already have for Jest to simplify
    testing and to make it DRY. `mixPropTests()` is introduced to mix those
    to create combinations to be tested.


bedrich-schindler:
@literat, this PR might be interesting for you as it shows Playwright Component Testing in action I talked about on Wednesday.

.env.playwright.dist
bedrich-schindler, line 1: I copied this from our internal project (WECL). I expect more configuration options to come in the future (e.g. reporter configuration) so take this as a base.

.github/workflows/test.yml
```diff
@@ -25,8 +25,22 @@ jobs:
...
       - name: Test and generate code coverage info
         run: npm test
 
+      - name: Test Playwright and generate test report
+        run: npm run test:playwright-ct:all
+
       - name: Upload code coverage info to Coveralls
         uses: coverallsapp/github-action@v2
+
+      - name: Upload Playwright test report
```
bedrich-schindler, line 40: You can download artifact and run the report locally as you would on local machine. If we want to generate some report, create issue for that. Currently, there is no official plugin for this. I just wanted to point this out.

src/components/Button/__tests__/Button.spec.jsx
```diff
@@ -0,0 +1,146 @@
...
+import {
+  mixPropTests,
+  propTests,
+} from '../../../../tests/playwright';
+import {
+  ButtonForTest,
+  ButtonForRefTest,
+} from './Button.story';
+
+test.describe('Button', () => {
```
bedrich-schindler, line 15: @mbohal @adamkudrna I would suggest starting the review from here. This is complex example representing visual testing, including `mixPropTests` that creates combinations of multiple prop tests.

src/components/Button/__tests__/Button.spec.jsx
```diff
@@ -0,0 +1,146 @@
...
+          id="test-id"
+        />,
+      );
+
+      const id = await component.getAttribute('id');
+      expect(id).toBe('test-id');
+    });
+
+    test('ref', async ({ mount }) => {
+      const component = await mount(
```
bedrich-schindler, line 74: One of Playwright CT restrictions is that in can pass only primitives as prop. Otherwise story must be used. Check the documentation: https://playwright.dev/docs/test-components#test-stories

 This is reason why I came up with such weird test for ref.

src/components/Button/__tests__/Button.story.jsx
bedrich-schindler, line 1: Story is standard approach for by-passing restriction of Playwright CT. See documentation: https://playwright.dev/docs/test-components#test-stories

For those who use storybook (I think that @adamkudrna and @literat do), there is possibility to use it with storybook as well. See https://storybook.js.org/blog/portable-stories-for-playwright-ct/

tests/playwright/propTests/labelVisibilityPropTest.js
```diff
@@ -0,0 +1,51 @@
...
+      props: {
+        labelVisibility: 'xs',
+      },
+    },
+  ];
+
+  Object.entries(breakpoints).map(([breakpointName, breakpointValue]) => {
+    tests.push({
+      name: `labelVisibility: ${breakpointName} (visible)`,
+      onBeforeTest: async (page) => {
```
bedrich-schindler, line 24: This is quite interesting idea I had to came up with. Some test cases might require to modify browser and things like that. And this is a possible way how to do it.

tests/playwright/components/TestIcon/TestIcon.jsx
mbohal, line 1: Do we need the extra folder `TestIcon`? We could just leave it in `tests/playwright/components/TestIcon.jsx`.

tests/playwright/constants/breakpoints.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable sort-keys */
+export const breakpoints = {
+  // xs: 0,
```
mbohal, line 3: Why is `xs` commented out? 

We should either enable it or add a comment.

package.json
```diff
@@ -48,16 +48,18 @@
     "copy:css": "cp src/docs/_assets/generated/react-ui.css dist && cp src/docs/_assets/generated/react-ui.development.css dist",
     "copy:js": "cp src/docs/_assets/generated/react-ui.js dist && cp src/docs/_assets/generated/react-ui.development.js dist",
     "eslint": "eslint --ext js,jsx src",
-    "jest": "jest src --coverage",
     "lint": "npm run eslint && npm run markdownlint && npm run stylelint",
     "markdownlint": "markdownlint-cli2 \"CONTRIBUTING.md\" \"README.md\" \"RELEASING.md\" \"src/**/*.md\"",
     "postbuild": "npm run copy",
-    "postinstall": "cp -n .env.dist .env || true",
+    "postinstall": "cp -n .env.dist .env || cp -n .env.playwright.dist .env.playwright || true",
```
mbohal, line 54: This does not work, only the `.env` gets created.

mbohal (changes requested):
I really, really like this!

tests/playwright/components/TestIcon/TestIcon.jsx
bedrich-schindler: I follow standard conventions we have. I can imagine we would like to add simple component (I have one in mind) and it would require styles as well.

CONTRIBUTING.md
```diff
@@ -70,6 +70,23 @@ docker compose run --rm node_shell
     docker compose run --rm mkdocs_build_site
     ```
 
+#### Playwright
```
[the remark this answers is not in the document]
bedrich-schindler: Let's discuss this on the meeting. To be honest, I was struggling with the documentation as we do not have as good documentation as on other project. We do not document most of npm scripts we have. We also do not have system in npm scripts I would like to have. Let's open those topics in few hours.

tests/playwright/components/TestIcon/TestIcon.jsx
mbohal: I think it is OK to have this structure:
```text
|- ComplexComponent
|  | - ComplexComponent.jsx
|  | - ComplexComponent.scss
|  | - index.js
|- SimpleComponent.jsx
```

That is to mix it as needed because in both cases the import is the same. Thus changing syntax from simple to complex is no problem.

CONTRIBUTING.md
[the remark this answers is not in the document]
bedrich-schindler: We decided to create TESTING.md where we cover it in detail.

atmelmicro (approved):

tests/playwright/components/TestIcon/TestIcon.jsx
mbohal: Your solution is going to work better in alphabetical sorting when folders are listed on top/bottom.
Moreover, there is one occurrence and it is used only in tests. 

The rest of the team seemed to prefer your solution, so I guess leave it as is.
I'm marking this conversations as resolved.

lukasbriza (approved):

package.json
```diff
@@ -48,16 +48,18 @@
...
     "precopy": "rm -rf dist && mkdir dist",
     "prepublishOnly": "npm run build",
     "start": "webpack --watch --mode=development",
     "stylelint": "stylelint \"src/**/*.scss\" --config stylelint.config.js",
-    "test": "npm run jest"
+    "test": "npm run test:jest",
```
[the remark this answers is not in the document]
mbohal: On a second thought, this is a bigger issue that is beyond the scope of this PR.

I'm marking this discussion as resolved and I will bring it up on another occasion.

bedrich-schindler:
I rebased it as there were conficts in package-lock.json

CONTRIBUTING.md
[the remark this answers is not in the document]
bedrich-schindler: @adamkudrna @mbohal I added TESTING.md, created symlink as it is with README and CONTRIBUTING and covered it as we are do in our company projects.

bedrich-schindler:
@mbohal and @adamkudrna, ready for round 2.

bedrich-schindler:
One more thing. I created follow up for better organization of test runners that way beyond this pull request:

https://github.com/react-ui-org/react-ui/issues/590

Part of the issue is uniform environment to prevent unexpected errors on Apple Silicon where you need to run `npm ci` inside Docker.

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+#### Opening Test Report
+
+After running Playwright tests, test report can be served into a browser
+by using the following command:
+
+```bash
+npx playwright show-report
+```
```
adamkudrna, line 106: Does it work when run from inside of the Docker container?

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+cp .env.playwright.dist .env.playwright
+```
+
+#### Running Tests
+
+Playwright tests can be run using the following commands:
+
+1. [Open][gh-gg-playwright] `playwright` Docker container
```
adamkudrna, line 80: Could add the command? I think it's something like `docker compose up -d playwright` and `docker compose exec -it playwright bash`, but it seems to quit immediately for me.

adamkudrna: Could you also mention the whole `docker compose run` command? I think in some scenarios it's more convenient to run Playwright like this.

I think it goes like `docker compose run --rm playwright bash -c 'npm run test:playwright-ct:XXX'`.

package.json
```diff
@@ -77,6 +79,7 @@
     "@babel/preset-react": "^7.24.7",
     "@babel/register": "^7.24.6",
     "@happy-dom/jest-environment": "^16.6.0",
+    "@playwright/experimental-ct-react": "^1.50.1",
```
adamkudrna, line 82: This must match the Docker image version, I believe.

.github/workflows/test.yml
```diff
@@ -25,8 +25,22 @@ jobs:
...
         run: npm ci
 
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
+
       - name: Test and generate code coverage info
         run: npm test
 
+      - name: Test Playwright and generate test report
+        run: npm run test:playwright-ct:all
```
adamkudrna, line 35: This means Playwright tests run in a different environment than the one on the local machine. Did you consider running it in the same Docker container also on GitHub? I'm afraid of varying test results otherwise.

See https://github.com/lmc-eu/spirit-design-system/blob/main/.github/workflows/e2e.yml#L69

src/components/Alert/__tests__/Alert.spec.jsx-snapshots/Alert-visual-color-danger-1-chromium-linux.png
adamkudrna, line 1: I can see our web font (currently Titillium, before #417) is not used here. Can you inject it before each test?

adamkudrna (changes requested):
The tests themselves look good, however I have some major comments on the dev stack and the docs.

No matter that, having finally the visual tests is so cool! 👏🏻

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+Tools used to test the application:
+
+* **ESLint** (static code analysis of JS files)
+* **Markdownlint** (static analysis of MD files)
+* **Stylelint** (static code analysis of CSS files)
+* **Jest** (unit testing)
+* **Playwright** (visual and functional component testing)
+
+Generally, `npm test` and `npm run test:playwright-ct:all` should be run  within
+their designated Docker containers before pushing changes to the repository.
```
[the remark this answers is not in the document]
adamkudrna: **UPDATE:** In the light of discussion in #590, this comment may become irrelevant.

TESTING.md
adamkudrna: **UPDATE:** All Docker related comments may become irrelevant if we agree the tests MUST be run within Docker, see #590. Other edits of the docs may need to be made, but let's wait until the changes are done.

src/components/Alert/__tests__/Alert.spec.jsx-snapshots/Alert-visual-color-danger-1-chromium-linux.png
adamkudrna: **UPDATE:** Or we can consider it a feature 🙂.

package.json
bedrich-schindler: Yes, it must. And it matches. If not, Playwright would fail with error telling it.

TESTING.md
bedrich-schindler: Unfortunatelly not. I am glad you catch this as I was originally mentioning host system in the docs, but I must have removed in when changing texts to match styling used in the rest of the documentation.

I made slight improvement so you can change port if you need (this is just part of .env.playwright as always). But it is not exposible from the Docker container yet. I have tried several things to expose it but without success. But I do not see problem running report from the host system or opening HTML file directly. 

Even though `npx playwright show-report` is standard approach and recommended by output after Playwrifght finishes, I added `test:playwright-ct:show-report` script that runs it with variable port.

bedrich-schindler: I might probably create issue in Playwright repository when I have time .

.github/workflows/release-management.yml
```diff
@@ -24,12 +24,18 @@ jobs:
...
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
+
       - name: Lint
         run: npm run lint
 
       - name: Test
         run: npm test
 
+      - name: Test Playwright
```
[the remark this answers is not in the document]
bedrich-schindler: I think that it is not as easy as you think. 4 people agreed that this is OK, so I would stick with it for now.

The reason why Playwright is used in the name is that it covers diffreent types of tests and only those tests must be run there, on unified testing environment. If you run some unit tests in Jest or unit tests for Sass, they do no require nothing but node. But this is not true for Playwright.

I agree that this is not as nice as I would like to have, but there is no easy way to solve it. And what more, this PR will be migrated to TypeScript and during temporary phase, every configuration will be postfixed Playwright to distingush it as we agreed. 

So I would discuss it later to not to block the work that needs to be unblocked.

.github/workflows/test.yml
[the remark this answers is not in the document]
bedrich-schindler: I did it as you wish. But I think that more improvement can be discussed later together with https://github.com/react-ui-org/react-ui/pull/588#discussion_r1969354898

I have quite different opinion on this. Until now, Jest / React testing library tests also run everytime. I am afraid that this is more complicated than you think. if there is some automation, fine. But even during maintenance, tests can got broken pretty easy. Easier that you think. And maybe easier that Spirit team thought of. Even update of dev dependencies can broke it.

.github/workflows/test.yml
bedrich-schindler: I am going to try it. I never configured Docker on Github CI, so that`s why i did not do that. And Playwright did not do that in their setup for Github CI as well. Thanks for the link

TESTING.md
bedrich-schindler: I used style of the documentation we have crafter in RACOM where working with Docker is described in Contribution guidelines. But I did it as you wished.  I added docker command to every section so it is easier to call the commands. 

Speaking about running single command. I added just to contributing guidelines where you can see both approaches as i belongs to section about Docker. I hope it is enough.

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+Playwright tests must be run in a Docker container to ensure a uniform
+environment. Otherwise, snapshots would differ between operating systems.
+
+This is the reason why you need to run Playwright tests separately
+from other tools.
+
+#### Configuration
+
+Test parameters can be tweaked by creating and tweaking `.env.playwright` file:
```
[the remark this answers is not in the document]
bedrich-schindler: I would stay with `tweaked` - we already use this and it is not from my head neither.

TESTING.md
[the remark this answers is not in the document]
bedrich-schindler: Yeah, I do not like this even though I created this PR, but as you write, this is irelevant in light of #590. I think that this must be improved at the conceptional level, althogether with Github workflows so it whole make sense. It is just to big and time exhausting to solve it immediately.

src/components/Alert/__tests__/Alert.spec.jsx-snapshots/Alert-visual-color-danger-1-chromium-linux.png
bedrich-schindler: This is good point! It seems that fonts need to be set explicitelly. I can look at it or leave it like it is. But I think it is bug, not feature. 

So I will solve it, but it should not be a big deal in the light of using system fonts and in situation when user can select any font he/she wants. But I will fix it anyway.

.github/workflows/test.yml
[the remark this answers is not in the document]
adamkudrna: It all depends on how long the tests will run. It's more than 10 minutes in case of Spirit and it takes a considerable amount of infrastructure we pay must pay for, so we decided to put them aside. This is my experience, take it or not. As long as the total running time is acceptable for us, we can surely run the visual tests with every push as you suggest.

TESTING.md
adamkudrna: Thanks for the updates. Because it didn't make sense without knowing the internal RACOM guidelines.

.github/workflows/test.yml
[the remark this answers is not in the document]
bedrich-schindler: For now, it is in separate file as you wished and the running in only when necessary is already part of improvement issue: https://github.com/react-ui-org/react-ui/issues/590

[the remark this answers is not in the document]
bedrich-schindler: Just a quick note. I hope we will be faster than Spirit as we use natively supported approach.

For info, on local machine, 100 tests takes approximately 10 seconds when runnning in sequence and approximately  4 seconds when running in paralel. This is time without initialization and building. Let's see the final results.

I think that even if we have 5000 tests, we can be bellow 10 minutes on single core. I hope we can do better. And if we can find a way ho to share artifacts generated by Playwright, we can be even faster and if we do not have to re-build components that did not changed.

TESTING.md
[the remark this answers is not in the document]
bedrich-schindler: Thank you for the comments before. I think we have still room for improvements but lets see how it works in real work.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,40 @@
+name: Playwright
```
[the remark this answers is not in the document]
bedrich-schindler: I created extra point in the improvement issue: https://github.com/react-ui-org/react-ui/issues/590#issuecomment-2683271025

I see a huge diagreement on this. But when this is just about naming already approved with the rest of the team, I think we can solve it in #590 and get back to it ASAP. I expect to jump on that issue on Monday.

bedrich-schindler:
As I mentioned in comments, there is follow-up issue that tracks improvements that need to be discussed and processed on Monday's call: https://github.com/react-ui-org/react-ui/issues/590#issuecomment-2683271025

If there is not blocking problem, we should close this as soon as it gets and work on improvements later not to block further work as none of the comment is related to writing tests but to stack around it.

.github/workflows/playwright.yml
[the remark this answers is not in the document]
bedrich-schindler: I just want to point out that if you check Playwright documentation, I use pretty same naming as Microsoft in their documentation: https://playwright.dev/docs/ci#fail-fast

And about the `Run Playwright tests`. Even Spirit has the same name, see. https://github.com/lmc-eu/spirit-design-system/blob/main/.github/workflows/e2e.yml#L68C15-L68C35

So I just want to say that me, Microsoft and partly your second team uses the convention you disagree with 👀

.github/workflows/release-management.yml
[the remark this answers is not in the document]
bedrich-schindler: The discussion about the similar think is here: https://github.com/react-ui-org/react-ui/pull/588#discussion_r1970577396

TDLR; Spirit and Microsoft uses the same naming as me.

.github/workflows/test.yml
bedrich-schindler: It is now run in Docker container completely as Playwright suggests in the doc. I think that Spirit's way is not correct as it mixes setup and testing in one step. This is not best practice ...

src/components/Alert/__tests__/Alert.spec.jsx-snapshots/Alert-visual-color-danger-1-chromium-linux.png
bedrich-schindler: Thank you again for catching this. Font was missing in index.html template. 🙈

bedrich-schindler:
Pull request should be ready for next round @adamkudrna.

I also told @lukasbriza to check the discussions again to bring next opinion into discussion.

.github/workflows/playwright.yml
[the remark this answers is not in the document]
adamkudrna: I'm afraid you are missing the point, trying to convince me the rest of the world knows better than I do (do we really need that, BTW?).

My point is about the **workflow name**. It is the first thing to see when scanning the checks here on GitHub, and it's called "End-to-end tests" [in case of Spirit](https://github.com/lmc-eu/spirit-design-system/blob/main/.github/workflows/e2e.yml#L1). Since other workflows (checks) have generic names like "Lint" or "Test" instead of "ESlint, Stylelint, Markdownlint" and "Jest", all I bring is a question why "Playwright" and not "End-to-end test" so it is consistent (more or less — CodeQL is a check added with help of GitHub and it can be renamed).

Never mind, I don't need to waste more time on it. There are more important things for us to do.

tests/playwright/templates/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Playwright Component Testing Template</title>
+    <link
+      href="https://fonts.googleapis.com/css2?family=Titillium+Web:wght@300;400;700&display=swap"
```
adamkudrna, line 8: Thanks for adding the font 👍🏻 .

I couldn't figure out where the styles come from, so this may not be the right place for the question, but it's important:

Do you also load the styles from `src/index.scss` / `react-ui.css`? Now it may not make any difference in the snapshots, but we will see it later when we test more complicated components that rely on inheritance from foundation CSS.

bedrich-schindler: `tests/playwright/templates/index.jsx` is the second entry point that imports styles.

.github/workflows/playwright.yml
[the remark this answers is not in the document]
bedrich-schindler: Yeah. We can of course change it in the future and discuss in in follow-up issue. But as it is based on documentation, it should be OK at least for now.

TESTING.md
mbohal: Honestly I just run `cd playwright-report` and then `python -m http.server 8000`.
It then serves the report on port 8000.

No need to use Playwright for that and it works outside of docker. I think virtually every modern *NIX system has got Python installed.

TESTING.md
```diff
@@ -0,0 +1,119 @@
...
+npm run test:playwright-ct:<all|all-with-update> -- <cli_argument>
+```
+
+#### Opening Test Report
+
+After running Playwright tests, test report can be opened **on the host system**
+by using the following command:
+
+```bash
+npm run test:playwright-ct:show-report
```
mbohal, line 112: I have written this to a comment under outdated yet unresolved discussion. I will put it here as well, just to be sure it does not disappear.

To serve the report I just run `cd playwright-report` and then `python -m http.server 8000`.
It then serves the report on port 8000.

I find this better:

* port is set inline, no need for the `PW_REPORT_PORT` const and non-pretty code in `package.json`
* better separation of concerns as playwright creates a report and then another tool serves it
* On *NIX systems Python is typically installed by default, Node.js is not.

I think we should add this method to docs. I would even consider removing the current one, but I have no strong feelings about it.

CONTRIBUTING.md
```diff
@@ -22,14 +22,20 @@ To change them, edit the `.env` file as needed.
...
 To log into the container, run:
 
 ```bash
 docker compose run --rm node_shell
 ```
 
+If you want to run single command, run:
+
+```bash
+docker compose run --rm node_shell bash -c 'npm run <command>'
```
mbohal, line 36: `'npm run <command>'` shoould be `'npm <command>'`

Example: `npm ci`

CONTRIBUTING.md
```diff
@@ -70,6 +76,28 @@ docker compose run --rm node_shell
...
+To log into the container, run:
+
+```bash
+docker compose run --rm playwright
+```
+
+If you want to run single command, run:
+
+```bash
+docker compose run --rm playwright bash -c 'npm run <command>'
```
mbohal, line 93: If we wanrt to have this detail, I would write:

`docker compose run --rm playwright bash -c 'npm run test:playwright-ct:*'`

To illustrate, that it is not correct to run anything else there.

TESTING.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+Test report cannot be server from Docker container yet.
+
+[gh-gg-node-shell]: /docs/contribute/general-guidelines#node-shell
```
mbohal, line 117: These links do not work outside of the context of mkdocs. We need to add `src` to the path.

I dislike having this text twice in the repo, but this also goes for `CONTRIBUTING.md` so I think it it beyond the scope of this issue -> I will raise it elsewhere.

TESTING.md
```diff
@@ -0,0 +1,119 @@
...
+
+**On host:**
+
+[Open][gh-gg-playwright] `playwright` Docker container:
+
+```bash
+docker compose run --rm playwright
+```
+
+**Within `node_shell`:**
```
mbohal, line 86: I think it should say:

> Within `playwright` container

mbohal (approved):

TESTING.md
bedrich-schindler: I reacted to another conversation. TLDR; Check latest TESTING.md. Now it can be run from Docker, reason is described at another comment.

bedrich-schindler: Just to clarify this. This approach is simply the standard approach recommended by Playwright. Reason is that Playwright can generate not only HTML report, but JSON report and much more. This command always serves the report despite the report type. I might be better to run JSON reports on CI as it is more readable than HTML source, but it is up to another discussion.

bedrich-schindler:
**⚠️⚠️⚠️ Do not merge this PR until @adamkudrna release new version ⚠️⚠️⚠️**

I merged TypeScript support for Playwright into this branch/pull request, squashed commits and rebased it. @lukasbriza, make all changes from this branch (not from `master`).

bedrich-schindler:
@adamkudrna, when you have time, you can review this. Without your permission, we would not be able to merge it as you are codeowner.

PR #589, opened 2025-02-15
Update Github Workflows to use Ubuntu 24.04

bedrich-schindler:
This is required by upcoming pull request as Ubuntu 20.04 is old and generates incorrect snapshots as latest Playwright Docker Image uses Ubuntu 24.04.

### DIFF
--- a/docker/node/Dockerfile
+++ b/docker/node/Dockerfile
@@ -1,3 +1,3 @@
-FROM        node:20
+FROM        node:22
 RUN         mkdir /workspace
 WORKDIR     /workspace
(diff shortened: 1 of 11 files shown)


mbohal (approved):

docker/node/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM        node:20
+FROM        node:22
```
adamkudrna, line 1: 👍🏻

adamkudrna (approved):
I hope the release management doesn't break. Apart from that, thanks for the change! 👍🏻

Issue #587, opened 2025-02-11
Implement visual testing


bedrich-schindler:
The task is to create visual snapshot test for a simple component (e.g. `<Badge />` or such) using Playwright component testing tool.

The aim is to have a working demo to serve as bases for discussion before we add tests for other components.

Issue #394, opened 2022-08-01
Transform codebase to Typescript


bedrich-schindler:
We need to transform codebase from standard JavaScript to TypeScript. Inspiration can be found in private RACOM repositories where TypeScript is already used. If you use parts of configurations from RACOM repositories, it would be easier for you as we would not have to much comments on that.

We would like to use strict typings, so that `any` type is prohibited if not necessary.

---

Speaking of convention where to put type definitions. It seems the most practical for us to have structure of component like this:

- Directory `<ComponentName>`
  - File `index.ts` - exporting named export
  - File `<ComponentName>.tsx` - component itself
  - File `<ComponentName>.types.ts` - all types associated with component

If you thing that `<ComponentName>.types.ts` is not necessary, place type definitions into `<ComponentName>.tsx`  before definition of component, but separate file seems more friendly to us (at least for now). You can of course look into other libraries like Material UI, Ant Design, ..., how they work with type definitions.

---

There is lot of props that are defined as enums. Like colors, variants, priorities etc. Those definitions should be placed in some global file (e.g. `src/types.ts`, `src/definitions.d.ts`, ...) , do not define them per components, we do not want to repeat those definitions in each component.

---

Speaking of "API" section in documentation. As @mbohal wrote in the email, do not try to fix "table of props". It will be solved by @mbohal in following issue as we use our custom documentation platform.

bedrich-schindler:
After @mbohal's decision, I assigned @lukasbriza to this issue.

What we want now is to pull branch `transform-typescript` and rebase it onto `master` and resolve all conflicts as try to finish the transformation. As this was not touched in months/years, there might be a lot of changes and some files might be missing at all.

I would personally consider whether you are (@lukasbriza) able to rebase it or whether there is better way - e.g. start from current `master` and do it manually file-by-file by renaming and copying it manually. For this might be useful to open `transform-typescript` branch in Github editor (GitHub.dev) from where you can copy it simply. See: https://github.dev/react-ui-org/react-ui/tree/transform-typescript

I am just giving you suggestions as we are not aware about the state of transformation and might be better for you to start over.

Do not hesitate to ask us before deciding something as it might result in lot of work in later phase.

mbohal:
Tahnk you @bedrich-schindler , I only have two points to add:

1.  I think there are components that did not change in a significant way for years. In those cases I think it would be practical to use what has been done instead of rewriting. Of course there are other parts that are new or have changed a lot and starting anew is a better option.
[...]

PR #599, opened 2025-02-27
Enable TypesScript for Playwright (#587)

bedrich-schindler:
Transform Playwright related files to Typescript and introduce temporary configuration files `.eslintrc-ts`, `jest.config-ts.js` and `tsconfig.json` that must be removed when whole package is transformed into Playwright.

Due to missing TypeScript support in the rest of the package, `*.spec.tsx` and `*.story.tsx` file do not have complete types and might not be compatible with strict type checking that should be introduced in #394.

The reason for such change is parallel work on transformation to TypeScript and on Playwright tests.


lukasbriza (approved):

.eslintrc-ts
```diff
@@ -0,0 +1,36 @@
+{
+  "extends": [
+    "@visionappscz/eslint-config-visionapps",
+    "airbnb-typescript",
+    "plugin:@typescript-eslint/recommended",
```
mbohal, line 5: I'm looking at https://typescript-eslint.io/getting-started/typed-linting/ and I can not find `recommended-requiring-type-checking`. All I see is `recommended-type-checked`  and that already includes `@typescript-eslint/recommended`.

@bedrich-schindler could you please take a look at this?

.eslintrc-ts
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "project": "./tsconfig.json"
+  },
+  "plugins": [
+    "@typescript-eslint"
+  ],
+  "rules": {
+    "import/prefer-default-export": "off",
```
mbohal, line 33: I would consider adding these:

```
'@typescript-eslint/consistent-type-definitions': ['error'],
'@typescript-eslint/consistent-type-exports': ['error'],
'@typescript-eslint/consistent-type-imports': ['error'],
```

jest.config-ts.js
```diff
@@ -0,0 +1,37 @@
...
+    '\\.scss$': 'identity-obj-proxy',
+    '\\.svg$': '<rootDir>/tests/jest/mocks/svgrMock.jsx',
+  },
+  preset: 'ts-jest',
+  setupFiles: [
+    '<rootDir>/tests/jest/setupJest.js',
+  ],
+  setupFilesAfterEnv: [
+  ],
+  testEnvironment: '@happy-dom/jest-environment',
```
mbohal, line 23: I know it is outside scope of this issue, but maybe we could add a comment here why we use happy-dom.

jest.config-ts.js
```diff
@@ -0,0 +1,37 @@
...
+    '**/*.test.{ts,tsx}',
+  ],
+  transform: {
+    '^.+\\.(ts|tsx)$': [
+      'ts-jest',
+      { useESM: true },
+    ],
+  },
+  transformIgnorePatterns: [
+    'node_modules/(?!(@react-ui-org))',
```
mbohal, line 34: I don't think this is needed here

mbohal (approved):

jest.config-ts.js
bedrich-schindler: We do not have comment neither here nor at `jest.config.js`. I can of course add it there, but while this is temporary file and original file does not have it, it does not make much sense to me.

PR #601, opened 2025-03-03
Introduce custom design of `FileInputField` (#244)

### LINKED ISSUE #244: Custom FileInputField
Customize input appearance to match other box field components and make it themeable.

⚠️ Native drag & drop support must be retained.

adamkudrna:
obrazek
obrazek
obrazek

Closes #244.

### DIFF
--- a/src/components/FileInputField/FileInputField.jsx
+++ b/src/components/FileInputField/FileInputField.jsx
@@ -1,10 +1,19 @@
 import PropTypes from 'prop-types';
-import React, { useContext } from 'react';
+import React, {
+  useContext,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
 import { withGlobalProps } from '../../providers/globalProps';
-import { classNames } from '../../helpers/classNames/classNames';
+import { classNames } from '../../helpers/classNames';
 import { transferProps } from '../../helpers/transferProps';
+import { TranslationsContext } from '../../providers/translations';
+import { getRootSizeClassName } from '../_helpers/getRootSizeClassName';
 import { getRootValidationStateClassName } from '../_helpers/getRootValidationStateClassName';
 import { resolveContextOrProp } from '../_helpers/resolveContextOrProp';
+import { InputGroupContext } from '../InputGroup';
+import { Text } from '../Text';
 import { FormLayoutContext } from '../FormLayout';
 import styles from './FileInputField.module.scss';
 
@@ -17,54 +26,156 @@ export const FileInputField = React.forwardRef((props, ref) => {
     isLabelVisible,
     label,
     layout,
+    multiple,
+    onFilesChanged,
     required,
+    size,
     validationState,
     validationText,
     ...restProps
   } = props;
 
-  const context = useContext(FormLayoutContext);
+  const internalInputRef = useRef();
+
+  // We need to have a reference to the input element to be able to call its methods,
+  // but at the same time we want to expose this reference to the parent component for
+  // case someone wants to call input methods from outside the component.
+  useImperativeHandle(ref, () => internalInputRef.current);
+
+  const formLayoutContext = useContext(FormLayoutContext);
+  const inputGroupContext = useContext(InputGroupContext);
+  const translations = useContext(TranslationsContext);
+
+  const [selectedFileNames, setSelectedFileNames] = useState([]);
+  const [isDragging, setIsDragging] = useState(false);
+
+  const handleFileChange = (files, event) => {
+    if (files.length === 0) {
+      setSelectedFileNames([]);
+      return;
+    }
+
+    // Mimic the native behavior of the `input` element: if multiple files are selected and the input
+    // does not accept multiple files, no files are processed.
+    if (files.length > 1 && !multiple) {
+      setSelectedFileNames([]);
+      return;
+    }
+
+    const fileNames = [];
+
+    [...files].forEach((file) => {
+      fileNames.push(file.name);
+    });
+
+    setSelectedFileNames(fileNames);
+    onFilesChanged(files, event);
+  };
+
+  const handleInputChange = (event) => {
+    handleFileChange(event.target.files, event);
+  };
+
+  const handleClick = () => {
+    internalInputRef?.current.click();
+  };
+
+  const handleDrop = (event) => {
+    event.preventDefault();
+    handleFileChange(event.dataTransfer.files, event);
+    setIsDragging(false);
+  };
+
+  const handleDragOver = (event) => {
+    if (!isDragging) {
+      setIsDragging(true);
+    }
+    event.preventDefault();
+  };
+
+  const handleDragLeave = () => {
+    if (isDragging) {
+      setIsDragging(false);
+    }
+  };
 
   return (
-    <label
+    <div
       className={classNames(
         styles.root,
         fullWidth && styles.isRootFullWidth,
-        context && styles.isRootInFormLayout,
-        resolveContextOrProp(context && context.layout, layout) === 'horizontal'
+        formLayoutContext && styles.isRootInFormLayout,
+        resolveContextOrProp(formLayoutContext && formLayoutContext.layout, layout) === 'horizontal'
           ? styles.isRootLayoutHorizontal
           : styles.isRootLayoutVertical,
-        disabled && styles.isRootDisabled,
+        resolveContextOrProp(inputGroupContext && inputGroupContext.disabled, disabled) && styles.isRootDisabled,
+        inputGroupContext && styles.isRootGrouped,
+        isDragging && styles.isRootDragging,
         required && styles.isRootRequired,
+        getRootSizeClassName(
+          resolveContextOrProp(inputGroupContext && inputGroupContext.size, size),
+          styles,
+        ),
         getRootValidationStateClassName(validationState, styles),
       )}
-      htmlFor={id}
-      id={id && `${id}__label`}
+      id={`${id}__root`}
+      onDragLeave={!disabled ? handleDragLeave : undefined}
+      onDragOver={!disabled ? handleDragOver : undefined}
+      onDrop={!disabled ? handleDrop : undefined}
     >
-      <div
+      <label
         className={classNames(
           styles.label,
-          !isLabelVisible && styles.isLabelHidden,
+          (!isLabelVisible || inputGroupContext) && styles.isLabelHidden,
         )}
-        id={id && `${id}__labelText`}
+        htmlFor={id}
+        id={`${id}__labelText`}
       >
         {label}
-      </div>
+      </label>
       <div className={styles.field}>
         <div className={styles.inputContainer}>
           <input
             {...transferProps(restProps)}
-            disabled={disabled}
+            className={styles.input}
+            disabled={resolveContextOrProp(inputGroupContext && inputGroupContext.disabled, disabled)}
             id={id}
-            ref={ref}
+            multiple={multiple}
+            onChange={handleInputChange}
+            ref={internalInputRef}
             required={required}
+            tabIndex={-1}
             type="file"
           />
+          <button
+            className={styles.dropZone}
+            disabled={resolveContextOrProp(inputGroupContext && inputGroupContext.disabled, disabled)}
+            onClick={handleClick}
+            type="button"
+          >
+            <Text lines={1}>
+              {!selectedFileNames.length && (
+                <>
+                  <span className={styles.dropZoneLink}>{translations.FileInputField.browse}</span>
+                  {' '}
+                  {translations.FileInputField.drop}
+                </>
+              )}
+              {selectedFileNames.length === 1 && selectedFileNames[0]}
+              {selectedFileNames.length > 1 && (
+                <>
+                  {selectedFileNames.length}
+                  {' '}
+                  {translations.FileInputField.filesSelected}
+                </>
+              )}
+            </Text>
+          </button>
         </div>
         {helpText && (
           <div
             className={styles.helpText}
-            id={id && `${id}__helpText`}
+            id={`${id}__helpText`}
           >
             {helpText}
           </div>
@@ -72,13 +183,13 @@ export const FileInputField = React.forwardRef((props, ref) => {
         {validationText && (
           <div
             className={styles.validationText}
-            id={id && `${id}__validationText`}
+            id={`${id}__validationText`}
           >
             {validationText}
           </div>
         )}
       </div>
-    </label>
+    </div>
   );
 });
 
@@ -86,10 +197,11 @@ FileInputField.defaultProps = {
   disabled: false,
   fullWidth: false,
   helpText: null,
-  id: undefined,
   isLabelVisible: true,
   layout: 'vertical',
+  multiple: false,
   required: false,
+  size: 'medium',
   validationState: null,
   validationText: null,
 };
@@ -116,7 +228,7 @@ FileInputField.propTypes = {
    * * `<ID>__helpText`
    * * `<ID>__validationText`
    */
-  id: PropTypes.string,
+  id: PropTypes.string.isRequired,
   /**
    * If `false`, the label will be visually hidden (but remains accessible by assistive
    * technologies).
@@ -135,9 +247,23 @@ FileInputField.propTypes = {
    */
   layout: PropTypes.oneOf(['horizontal', 'vertical']),
   /**
+   * If `true`, the input will accept multiple files.
+   */
+  multiple: PropTypes.bool,
+  /**
+   * Callback fired when the value of the input changes.
+   */
+  onFilesChanged: PropTypes.func.isRequired,
+  /**
    * If `true`, the input will be required.
    */
   required: PropTypes.bool,
+  /**
+   * Size of the field.
+   *
+   * Ignored if the component is rendered within `InputGroup` component as the value is inherited in such case.
+   */
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
   /**
    * Alter the field to provide feedback based on validation result.
    */
--- a/src/components/FileInputField/FileInputField.module.scss
+++ b/src/components/FileInputField/FileInputField.module.scss
@@ -1,8 +1,16 @@
+// 1. The drop zone is constructed as a button to support keyboard operation.
+// 2. Prevent pointer events on all children of the root element to not to trigger drag events on children.
+
 @use "../../styles/tools/form-fields/box-field-elements";
 @use "../../styles/tools/form-fields/box-field-layout";
+@use "../../styles/tools/form-fields/box-field-sizes";
 @use "../../styles/tools/form-fields/foundation";
 @use "../../styles/tools/form-fields/variants";
 @use "../../styles/tools/accessibility";
+@use "../../styles/tools/links";
+@use "../../styles/tools/transition";
+@use "../../styles/tools/reset";
+@use "settings";
 
 @layer components.file-input-field {
     // Foundation
@@ -18,6 +26,54 @@
         @include box-field-elements.input-container();
     }
 
+    .input {
+        @include accessibility.hide-text();
+    }
+
+    .dropZone {
+        --rui-local-color: #{settings.$drop-zone-color};
+        --rui-local-border-color: #{settings.$drop-zone-border-color};
+        --rui-local-background: #{settings.$drop-zone-background-color};
+
+        @include reset.button(); // 1.
+        @include box-field-elements.base();
+
+        display: flex;
+        align-items: center;
+        justify-content: start;
+        font-weight: settings.$drop-zone-font-weight;
+        font-size: var(--rui-local-font-size);
+        line-height: settings.$drop-zone-line-height;
+        font-family: settings.$drop-zone-font-family;
+        border-style: dashed;
+    }
+
+    .isRootDragging .dropZone {
+        --rui-local-border-color: #{settings.$drop-zone-dragging-border-color};
+    }
+
+    .isRootDisabled .dropZone {
+        cursor: settings.$drop-zone-disabled-cursor;
+    }
+
+    .root:not(.isRootDisabled, .isRootDragging) .dropZone:hover {
+        --rui-local-border-color: #{settings.$drop-zone-hover-border-color};
+    }
+
+    .root:not(.isRootDisabled, .isRootDragging) .dropZone:active {
+        --rui-local-border-color: #{settings.$drop-zone-active-border-color};
+    }
+
+    .dropZoneLink {
+        @include links.base();
+
+        &::before {
+            content: "";
+            position: absolute;
+            inset: 0;
+        }
+    }
+
     .helpText,
     .validationText {
         @include foundation.help-text();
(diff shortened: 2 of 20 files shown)


src/components/FileInputField/__tests__/FileInputField.test.jsx
```diff
@@ -26,7 +28,10 @@ describe('rendering', () => {
   formLayoutProviderTest(<FileInputField {...mandatoryProps} />);
 
   it.each([
-    ...disabledPropTest,
+    // TODO by adam@adamkundrna.cz on 2025-03-03
+    // With the new drop-zone element, we cannot get the input element by label anymore (`getByLabel()`).
+    // We need to find a new way to test this, possibly by removing the wrapping `<label>` element in all components.
+    // ...disabledPropTest,
```
mbohal, line 34: Generally speaking it is OK not to use the premade test object (`disabledPropTest`) but to write ones own if the component needs it.

I don't really understand why it does not work though. It seems that label <-> field association works fine in the browser.

mbohal (changes requested):

src/components/FileInputField/FileInputField.jsx
```diff
@@ -18,24 +25,39 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
-  const context = useContext(FormLayoutContext);
+  const formLayoutContext = useContext(FormLayoutContext);
+  const inputGroupContext = useContext(InputGroupContext);
+  const translations = useContext(TranslationsContext);
+
+  const [selectedFileName, setSelectedFileName] = useState('');
+
+  const handleFileChange = (event) => {
+    const file = event.target.files[0];
+    setSelectedFileName(file.name);
```
bedrich-schindler, line 42: We have example in `README.md` where `multiple` prop is used. However, if I select multiple files, only the first name is shown. With @mbohal, we agreed that for now, we would suggest:

* For 1 file: show file name
* For >1 files: show something like "<number> files selected"

In future, we can improve it e.g. by Popover that could show list of selected files. But there could be problem with keyboard interaction, so we would stick to simple solution and can talk about further improvements later.

---

Technically, we would change `selectedFileName` to array `selectedFileNames`. From peformance point of view, it would be easier to hold single string that would contain name or string "<number> files selected". But it would not work in case that user switches languages, so that's why we recommend using array of names for it.

bedrich-schindler:
@adamkudrna, this is more like question than comment. If I zoom in, component's margin/padding looks like this:

image

At first sight, I immediately saw that text is not centered, but it is rather few pixels down from the center. Can you check this?

---

What more, if I change styles of `FileInputField-module__dropZone`, specifically `vertical-align` and `line-height`, it does not affect the appearance, which is weird and might be related to my question.

src/styles/theme/_form-fields.scss
bedrich-schindler, line 1: @adamkudrna, this is more like question than comment. If I zoom in, component's margin/padding looks like this:

image

At first sight, I immediately saw that text is not centered, but it is rather few pixels down from the center. Can you check this?

---

What more, if I change styles of `FileInputField-module__dropZone`, specifically `vertical-align` and `line-height`, it does not affect the appearance, which is weird and might be related to my question.

src/translations/en.js
```diff
@@ -2,6 +2,10 @@ export default {
   Alert: {
     close: 'Close',
   },
+  FileInputField: {
```
bedrich-schindler, line 5: This is not related to this PR, but @mbohal, if I see this and my another suggestion to display number of selected files, I think we need more robust approach for translations. For English it is OK, but I guess there can be languages that can have such language rules for which this approach would not work.

src/components/FileInputField/__tests__/FileInputField.test.jsx
bedrich-schindler: See my comment about transformation to Playwright above.

I guess that problem might be that [getByLabelText](https://testing-library.com/docs/queries/bylabeltext/) does not detect item correctly when in combination with `within`. Other tests does not use `within`, so that is the difference. What more, `input` is not visible and those queries does not detect hidden elements. For example, getByRole does have special option `hidden` to override the behavior.

But as I say, I would solve it in playwright instead.

bedrich-schindler (changes requested):

src/components/FileInputField/FileInputField.jsx
```diff
@@ -54,12 +76,26 @@ export const FileInputField = React.forwardRef((props, ref) => {
         <div className={styles.inputContainer}>
           <input
             {...transferProps(restProps)}
-            disabled={disabled}
+            className={styles.input}
+            disabled={resolveContextOrProp(inputGroupContext && inputGroupContext.disabled, disabled)}
             id={id}
+            onChange={handleFileChange}
```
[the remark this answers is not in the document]
adamkudrna: > Also onChange(event); must be called in the handleFileChange(). This will fix the functionality test.

I don't understand this part. How can I put `onChange(event)` inside the `handleFileChange` function? 🤔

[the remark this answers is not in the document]
mbohal: The component accepts the `onChange` prop which is a function. This is just a normal prop, that is just a function that can be called or not.

The problem is that as it is now this function never gets called. I guess it should have been passed via `transferProps`, but it is not because you set `onChange` attribute on the `<input />` explicitely and that takes preference.

So we need to call the `onChange` function originating from prop manually in the handler.

I'm not so sure if I explained it clearly. Feel free to call me if not.

src/components/FileInputField/__tests__/FileInputField.test.jsx
mbohal: > But as I say, I would solve it in playwright instead.

:+1:

src/components/FileInputField/FileInputField.jsx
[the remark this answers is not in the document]
adamkudrna: I think I get what you mean. How about now?

src/components/FileInputField/FileInputField.jsx
```diff
@@ -17,34 +25,118 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+  };
+
+  const handleDragOver = (event) => {
+    event.preventDefault();
+    setIsDragging(true);
+
+    if (props?.onDragOver) {
+      props.onDragOver(event);
+    }
+  };
```
adamkudrna, line 88: My suspicion is this function is evaluated many times per second. Is it OK?

src/components/FileInputField/FileInputField.jsx
```diff
@@ -54,12 +146,32 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+            className={styles.input}
+            disabled={resolveContextOrProp(inputGroupContext && inputGroupContext.disabled, disabled)}
             id={id}
+            onChange={handleInputChange}
             ref={ref}
             required={required}
             type="file"
           />
+          <div className={styles.dropZone}>
+            <Text lines={1}>
```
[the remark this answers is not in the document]
adamkudrna: Spirit – vanilla JS implementation:
- [live demo](https://spirit-design-system.netlify.app/packages/web/src/scss/components/fileuploader/)
- [source](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/js/FileUploader.ts)

Spirit – React implementation:
- [live demo](https://spirit-design-system.netlify.app/packages/web-react/src/components/fileuploader/)
- [source](https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web-react/src/components/FileUploader)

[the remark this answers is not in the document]
bedrich-schindler: This is due to pointer events, see my comment: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop#prevent_the_browsers_default_drag_behavior

bedrich-schindler:
During our call with @mbohal, we agreed on several things:

* Change of `onDragOver` is described here: https://github.com/react-ui-org/react-ui/pull/601#discussion_r1983050339
* Due to complexity, we do not want to use synthetic events but rather custom prop that would cover input change and file drop at one place, something like `onFilesChanged` and document it. This is missing in the documentation yet. Due to its complexity, this can't be omitted.
* We do not want to make drag props available as we do not see benefits now and while do not want to use synthetic event. `onChange`, it would follow the same approach.
* We are not sure about transferProps. Currently, input is hidden and logic is provided through button and droppable area. While we do not want to provide traditional synthetic event `onChange` and input is hidden, are transfer props placed on correct element? Is input where all aria attributes should be placed? We frankly do not know yet.

bedrich-schindler (changes requested):
During our call with @mbohal, we agreed on several things:

* Change of `onDragOver` is described here: https://github.com/react-ui-org/react-ui/pull/601#discussion_r1983050339
* Due to complexity, we do not want to use synthetic events but rather custom prop that would cover input change and file drop at one place, something like `onFilesChanged` and document it. This is missing in the documentation yet. Due to its complexity, this can't be omitted.
* We do not want to make drag props available as we do not see benefits now and while do not want to use synthetic event. `onChange`, it would follow the same approach.
* We are not sure about transferProps. Currently, input is hidden and logic is provided through button and droppable area. While we do not want to provide traditional synthetic event `onChange` and input is hidden, are transfer props placed on correct element? Is input where all aria attributes should be placed? We frankly do not know yet.

src/components/FileInputField/FileInputField.jsx
mbohal: Talked bout on video call:
* We have to ensure that pointer events work on `<button>` and in general in the drop zone

mbohal:
> We are not sure about transferProps. 

We want `transferProps` to point on the `<input />`.

src/components/FileInputField/FileInputField.jsx
adamkudrna: I got it working now:

- I changed the CSS a bit so the dragging state is not being deactivated when cursor enters dropzone children (pseudo element on dropzone "link").
- I found out that `setIsDragging` in `handleDragOver` actually is necessary. Otherwise the dragging state can be turned off during complex hovering over the drop zone.

src/components/FileInputField/FileInputField.jsx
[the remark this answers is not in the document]
adamkudrna: `handleDragOver` seems to be the right way. Big boys have it the same: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/FileUploader/FileUploaderDropContainer.tsx

src/components/FileInputField/FileInputField.jsx
adamkudrna: I think this is resolved, right?

src/translations/en.js
adamkudrna: I think this can be closed as we don't want to address it in this PR, do we?

adamkudrna:
> * Due to complexity, we do not want to use synthetic events but rather custom prop that would cover input change and file drop at one place, something like `onFilesChanged` and document it. This is missing in the documentation yet. Due to its complexity, this can't be omitted.
> 
> * We do not want to make drag props available as we do not see benefits now and while do not want to use synthetic event. `onChange`, it would follow the same approach.

@bedrich-schindler @mbohal I admit I need a guidance regarding this change. Could you drop a suggestion somewhere in the code so I can fully understand it and do the rest of the work? 🙏🏻

bedrich-schindler:
> Due to complexity, we do not want to use synthetic events but rather custom prop that would cover input change and file drop at one place, something like onFilesChanged and document it. This is missing in the documentation yet. Due to its complexity, this can't be omitted.

This means that prop `onFilesChanged` should be introduced and then on input type="file", do something like:

```jsx
<input onChange={(e) => {
  onFilesChanged(e.target.files, e);
}}
```

and on droppable label (I guess):

```jsx
<label onDrop={(e) => {
  onFilesChanged(dataTransfer.files, e);
}}
```

Maybe prevent default actions if needed. And document this.

bedrich-schindler:
> We do not want to make drag props available as we do not see benefits now and while do not want to use synthetic event. onChange, it would follow the same approach.

This was request from @mbohal. You can for example use spread to omit all drag events from the transfer props, so even if defined, it never gets into component. This should be documented as well.

mbohal:
> You can for example use spread to omit all drag/drop events from the transfer props, so even if defined, it never gets into component. 

Since the drop related event handlers must be specified on the root `<div>` and the `transferProps` are passed to the `<input />` element I dont think we need to filter anything. Passing them via `transferProps` will not work anyway.

So, for example, I think that in [`dropHandler`](https://github.com/react-ui-org/react-ui/pull/601/files#diff-7b306d6537076cea83844613d9c1275f1a61d1f9130395cb11ff410a7b39b0a4R75) we can just remove [calling the `onDrop` function](https://github.com/react-ui-org/react-ui/pull/601/files#diff-7b306d6537076cea83844613d9c1275f1a61d1f9130395cb11ff410a7b39b0a4R80-R82).
The final code would thus look like this:

```jsx
  const handleDrop = (event) => {
    event.preventDefault();
    handleFileChange(event.dataTransfer.files);
    setIsDragging(false);
}
```

I hope I answered. If not, just give me a call so we can get this out of the way.

bedrich-schindler:
I haven't checked the code itself yet, but there is still missing information in the docs.

As this is more complex input with custom event handlers, I would like to have at least some note or section about this. The only documentation of `onFilesChanged` is this:

image

There are no function parameters visible nor the example is present.

Currently, section `Forwarding HTML Attributes` is telling lies as not all events are passed.

I would like to:

* Add example how the event should be handled and tell user that this is abstraction for onChange and onDrop.
* Tell users that some events are not available and should not be used.

bedrich-schindler:
If I have:

```
<FileInputField
label="Attachment"
onFilesChanged={(e) => {
  console.log(e);
}}
/>
```

`onFilesChanged` is not called when file is dropped, only label changes.

bedrich-schindler (changes requested):
I have to provide comment for "Request changes", but the issues are already top comments.

src/components/FileInputField/FileInputField.jsx
```diff
@@ -17,49 +25,142 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+
+  const handleDrop = (event) => {
+    event.preventDefault();
+    handleFileChange(event.dataTransfer.files);
+    setIsDragging(false);
+  };
+
+  const handleDragOver = (event) => {
+    event.preventDefault();
+    setIsDragging(true);
```
bedrich-schindler, line 79: This is still present. We wanted `setIsDragging(true);` to disappear from  `handleDragOver` as it is called hundredt times per second.

src/translations/en.js
bedrich-schindler: I think that you are right. We are just in problematic situation, because those translations won't work at Czech launguage. We should at least map this issue in new issue. @mbohal?

src/components/FileInputField/FileInputField.jsx
```diff
@@ -17,49 +25,142 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+    }
+  };
+
+  const handleClick = (event) => {
+    event.currentTarget.previousElementSibling.click();
+  };
+
+  const handleDrop = (event) => {
+    event.preventDefault();
+    handleFileChange(event.dataTransfer.files);
```
bedrich-schindler, line 73: `handleDrop` does not call `props.onFilesChanged`, so this is reason why it does not work on drop.
Please, handle both events as I requested earlier this day and pass event itself and files, two parameters.

props.onFilesChanged(files, e) or props.onFilesChanged(e, files), i do not care which way. Also `onFilesChanged` should be required in my opinion, there is no way to work with files uncontrolled way.

src/components/FileInputField/FileInputField.jsx
```diff
@@ -17,49 +25,142 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+  const handleDragEnter = () => {
+    setIsDragging(true);
+  };
+
+  const handleDragLeave = () => {
+    setIsDragging(false);
+  };
+
+  useEffect(() => {
+    setIsDragAndDropSupported('draggable' in document.createElement('span'));
```
bedrich-schindler, line 91: I think that this check is not necessary:

image


We support latest browser versions as you can see in browser list and `draggeble` was even supported in era of IE.

src/components/FileInputField/FileInputField.jsx
adamkudrna: It is present and I explained it's needed. If I remove it, the "is dragging" state may not work correctly during more complex dragging. I am aware of how often it's being called, but I decided to leave it as is not just because Spirit has it, but I found the same solution in the respected Carbon design system by IBM.

❌ This is how it behaves when I remove it (I emphasized the "is dragging" state with red color):

https://github.com/user-attachments/assets/2e54a5d3-7c32-430a-aba0-74288ff2a1aa

✅ This is correct:

https://github.com/user-attachments/assets/2d5a6657-c97e-4c2a-8cc2-85553fe3310b

src/translations/en.js
mbohal: Lets deal with it in another issue. For now all we need is english.

src/components/FileInputField/FileInputField.jsx
adamkudrna: I know, it may look like a sure thing, but I am not persuaded about various mobile browsers. The detection is a light-weight feature and I'd rather leave it here so we can be sure.

Copilot thinks the same BTW:

> Drag-and-drop functionality is not supported in all current desktop and mobile browsers. It is generally supported in most modern desktop browsers, but support on mobile browsers can be inconsistent. Therefore, it is a good practice to detect its availability.

src/translations/en.js
adamkudrna: Filed under #611, closing here.

src/components/FileInputField/FileInputField.jsx
bedrich-schindler: You can discuss this with @mbohal, but from my point of view, this is type of check should not be here. This check is costly as it must be done for each instance and it does not have any advantage as this kind of API is expected by our configuration of browserlist. This is not bleeding-edge technology, this is there with us for decade. It is like writing fallback for IE in 2025. Even CSS Grid is younger that this.

src/components/FileInputField/FileInputField.jsx
bedrich-schindler: I already spoke about this at this comment: https://github.com/react-ui-org/react-ui/pull/601#discussion_r1983050339

With pointer events configuration, this should be solvable. It was working for me when I tried it last week. I just did not tested consequences.

I want to point out that prevent default still needs to be called inside of onDragOver (https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop#prevent_the_browsers_default_drag_behavior), but it should not set state there.

I disagree with implementation setting state over and over. If this is not solvable using ideas I presented in the comment, we should at least create follow-up issue to solve this later. But this implementation is not correct and If I can recall correctly, @mbohal had same opinion last week.

src/components/FileInputField/README.md
```diff
@@ -162,6 +172,40 @@ It's possible to disable the whole input.
 <FileInputField
   disabled
   label="Attachment"
+  onFilesChanged={() => {}}
+/>
+```
+
+## Handling Files
```
bedrich-schindler, line 179: This is good for me! 

I am just not sure about `can be handled`. It must be/have to be handled as it is required prop.

src/components/FileInputField/FileInputField.jsx
bedrich-schindler, line 1: This implementation look good!

src/components/FileInputField/FileInputField.jsx
bedrich-schindler: I pushed one commit that should demonstrate that is works. Or at least I have not found a case when not. 

I might require some tweaking of cursor, I just pushed raw changes to evaluate it.

What more, it improves working with input field. It now uses ref instead of trying to find previous element in the DOM. This is more React way.

src/components/FileInputField/FileInputField.jsx
```diff
@@ -17,49 +27,155 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+        getRootSizeClassName(
+          resolveContextOrProp(inputGroupContext && inputGroupContext.size, size),
+          styles,
+        ),
         getRootValidationStateClassName(validationState, styles),
       )}
-      htmlFor={id}
-      id={id && `${id}__label`}
+      id={id && `${id}__root`}
+      onClick={handleClick}
```
adamkudrna, line 127: @bedrich-schindler Why did you move `onClick` here? 👀 

- It belongs to the `<button>` element, not to the root `<div>`.
- Doing `<div onClick="">` is generally wrong because it break semantics and accessibility (keyboard control, assistive technologies). In HTML, we have `<button>`s to be clicked on. `<div>`s are for visual grouping with no semantic meaning and no accessibility role.

The linter reports it for a reason.

bedrich-schindler: Sadly, we are heading to component that will never be perfect. We have to balance and find the most suitable scenario. That is fact coming from the fact that we have to use hidden input file field. Known hack, oficially promoted hack, but still hack.

The only way I was able to came up with in the short time was combination with disabling pointer on child elements and moving onClick to root element that has pointer events enabled. If we make it this way, we can add all the semantics to the top level element to behave like a button and we can completely omit underlying button as well. That being said, we do not need to have any visible interaction with the child component. Everything is solvable at CSS level pretty easy.

I understand that there is a "link" in the button and that hover and other pseudo classes are not called on child component, but all the interaction can be solved on top layer. I just came up with the JS part of the solution that have to be evaluated first. I did not want to speed time on solution where that might not be accepted.

I want @mbohal to continue with the review as we are stuck at this point. There is no way I will aprove Adam's solution and he disagrees with mine. I stand strong in my opinion on this.

As I said yesterday, I see way that we can proceed with former solution due to time press, but would then require to start imediatelly on better solution. Mainstream solution are not often the best. Nobody even said that calling setter dozens/hundred times a second is the solution. It is simply not.

src/components/FileInputField/FileInputField.module.scss
```diff
@@ -1,13 +1,25 @@
...
+@use "settings";
 
 @layer components.file-input-field {
     // Foundation
     .root {
         @include foundation.root();
+
+        * {
+            pointer-events: none; // 2
+        }
```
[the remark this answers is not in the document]
bedrich-schindler: Discussion continues here: https://github.com/react-ui-org/react-ui/pull/601#discussion_r2005100132

src/components/FileInputField/FileInputField.jsx
adamkudrna: OK, removed.

adamkudrna:
FYI @mbohal @bedrich-schindler, I made the `id` mandatory so the input is paired with its label. That also allowed me to re-enable the previously disabled `disabledPropTest`.

adamkudrna:
Approved on a call last week, merging.

Issue #592, opened 2025-02-21
Transform `Action` components from Jest to Playwright CT


bedrich-schindler:
Follow transformation guidelines defined in https://github.com/react-ui-org/react-ui/issues/591

Transform following components:

<img width="199" alt="Image" src="https://github.com/user-attachments/assets/1bd7d5d6-af3b-490d-a482-c8dc20641b57" />

If any component from this group is transformed, check whether it is OK or need improvement.

PR #604, opened 2025-03-05
Transform Action jest tests into playwright (#592)

lukasbriza:
Refactoring components from action sections of RUI from Jest to Playwright tests.


bedrich-schindler (changes requested):

src/components/TextLink/__tests__/TextLink.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+          href="/test/uri"
+          label="test-label"
+        />,
+      );
+
+      await expect(component).toHaveAttribute('href', '/test/uri');
+    });
+  });
+
+  test.describe('functionality', () => {
```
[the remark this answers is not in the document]
lukasbriza: Added test for native behaviour (click, enter). Also preserved test for onclick behaviour.

src/components/TextLink/__tests__/TextLink.spec.tsx
```diff
@@ -38,23 +38,46 @@ test.describe('TextLink', () => {
...
+      await Promise.all([
+        page.waitForURL(testHref),
+        component.click(),
+      ]);
+
+      expect(page.url()).toContain(testHref);
+    });
+
+    test('calls native redirect when enter pressed', async ({ mount }) => {
+      const testHref = '/test/uri';
```
bedrich-schindler, line 64: I would move `testHref` higher to share it by all tests. There is one forgotten string not converted to this variable.

bedrich-schindler:
Resolve the last task, squash and merge.

bedrich-schindler (approved):

src/components/TextLink/__tests__/TextLink.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+        const screenshot = await component.screenshot();
+        expect(screenshot).toMatchSnapshot();
+      });
+    });
+  });
+
+  test.describe('non-visual', () => {
+    test('href', async ({ mount }) => {
+      const component = await mount(
+        <TextLinkForTest />,
```
bedrich-schindler, line 41: Story should contain all required values so it renders without error. But for purpose of the test, it might be better to explictely write tested prop here so it does not looks like a magic.

variants:

1. Use the same value as it is in story
2. Use another value

@lukasbriza agrees that is would be more explicit. What do you think, @mbohal?

bedrich-schindler (changes requested):

src/components/TextLink/__tests__/TextLink.spec.tsx
mbohal: I think 2) is the correct solution.

bedrich-schindler (approved):

PR #605, opened 2025-03-07
fixup! Transform Action jest tests into playwright (#592)

lukasbriza:
Radio and checkboxfield transformation into playwright.


bedrich-schindler:
To be discussed:

Interesting idea to include `undefined` as optional parameter. Neverthless I do not know whether this is there for optional props like helper texts that are not present or whethere it is there due to testing defaults. It is not clear from that. I want to point out, that some prop tests accepts undefined, but API documentation of component (e.g. SelectField) says that the default is `null`

src/components/CheckboxField/__tests__/CheckboxField.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+} from './CheckboxField.story';
+import { checkedPropTest } from './_propTests/checkedPropTest';
+
+test.describe('CheckboxField', () => {
+  test.describe('base', () => {
+    test.describe('visual', () => {
+      [
+        ...propTests.labelPropTest,
+        ...propTests.renderAsRequiredPropTest,
+        ...propTests.isLabelVisiblePropTest,
```
bedrich-schindler, line 23: I would like to have uniform approach so we do not have to thing whether is it is array or function. Lets discuss it first.

src/components/CheckboxField/__tests__/CheckboxField.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+        const component = await mount(
+          <CheckboxFieldForTest
+            helpText={helpText}
+            id={idValue}
+            label={label}
+            validationText={validationText}
+          />,
+        );
+
+        await expect(component.getByRole('checkbox')).toHaveAttribute('id', idValue);
```
bedrich-schindler, line 85: This is quite interesting! But as I know RUI, there are components for those such approach will not be enough. lets discuss it

src/components/CheckboxField/__tests__/CheckboxField.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+            testRefAttrValue="test-ref-value"
+          />,
+        );
+
+        await expect(component.getByRole('checkbox')).toHaveAttribute('test-ref', 'test-ref-value');
+      });
+    });
+
+    test.describe('functionality', () => {
+      test('calls synthetic event onChange()', async ({ mount }) => {
```
bedrich-schindler, line 105: This is draft, so I just want to remind not to forget to test keyboard interaction

src/components/CheckboxField/__tests__/CheckboxField.spec.tsx
bedrich-schindler: This comment is outdated as we decided to abandon parametrization.

lukasbriza:
This PR is only draft so I am closing it. Full PR will be introduced afterward.

PR #608, opened 2025-03-12
Refactoring/inputs jest to playwright

lukasbriza:
Rewrite jest tests for Inputs components into playwright.

I have a few questions for @mbohal or @bedrich-schindler.

When running full-page screenshots (in the case of SelectField), the rendering seems inconsistent. The images for the full page are re-generated every time I run the tests with the --with-update flag. Also, when testing components within the FormLayout context, they look oddly rendered when the horizontal layout is applied.

Should we address/fix this issue directly in this PR? Do you consider it a bug?

Additionally, would you like to add tests for ButtonGroup and InputGroup contexts at the component level, similar to how we are testing the FormLayout context?

Closes #607


tests/playwright/propTests/labelPropTest.tsx
```diff
@@ -1,8 +1,13 @@
+import React from 'react';
 import type { PropTests } from '../types';
 
 export const labelPropTest: PropTests = [
   {
     name: 'label',
     props: { label: 'Extra long label that exceeds regular length that anyone would expect' },
   },
+  {
+    name: 'nodeLabel',
```
bedrich-schindler, line 10: To keep it unified with other prop tests, please change this to `label=node` and also change `label=normal`. But first check the rest of the comment ...

---

First, I would like to say that you did good job identifying that different prop types can be passed into `label`, `helpText`, `validationText` etc. But I am thinking about improving the convention a bit.

Until now, we had been using `<prop name>=<prop value>` as a name of prop test item.  You introduced different types and variants like `<prop name>=normal`, `<prop name>=node`, `<prop name>=longText`. I would like to distinguish between value, data type (string, node, ...) and some variant (longText, normalText).

 I am not sure about the format, but I would suggest something like:

* `<prop name>:<prop type>=<prop value>`

So it will look like:

* `label:string`
* `label:node`

If we really need more variants, I would choose format like:

* `<prop name>:<prop type>=<prop value>[<test variant>]`

So it can look like:

* `label:string[normal]`
* `label:string[long]`
* `label:node`

As you can see, if we need to distinguish between multiple, we add it there. If not, we skip it.

I think that there is no such component (so this example it totally hypotetical), but as this is very versatile, we can endup in things like:

* someProp:string`
* someProp:bool=true`
* someProp:bool=false`
* someProp:shape[flat]
* someProp:shape[nested]

What do you think about it @adamkudrna @mbohal @lukasbriza? I just want to visibly make difference between values, prop types (data types) and some additional variants of tests.

bedrich-schindler (changes requested):

src/components/SelectField/__tests__/SelectField.spec.tsx
```diff
@@ -0,0 +1,283 @@
...
+            <SelectFieldForTest
+              {...props}
+            />,
+          );
+
+          if (onBeforeSnapshot) {
+            await onBeforeSnapshot(page, component);
+          }
+
+          const screenshot = await component.screenshot();
```
bedrich-schindler, line 102: To fixed errors caused by animations, pass following options into `screenshot()`:

```
{
          animations: 'disabled',
}
```

This should be used for all screenshot calls of this component.

bedrich-schindler: See: https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1-option-animations

src/components/SelectField/__tests__/SelectField.spec.tsx
```diff
@@ -0,0 +1,283 @@
...
+          expect(screenshot).toMatchSnapshot();
+        });
+      });
+    });
+
+    /**
+     * Full page screenshot is required for option tests because native select options are rendered outside of the DOM.
+     */
+    test.describe('visual-fullPage', () => {
```
bedrich-schindler, line 111: `visual-fullPage` does not make sence. Nest describe blocks instead (visual > fullpage > ...)

src/components/SelectField/__tests__/SelectField.spec.tsx
```diff
@@ -0,0 +1,283 @@
...
+            <SelectFieldForTest
+              {...props}
+            />,
+          );
+
+          if (onBeforeSnapshot) {
+            await onBeforeSnapshot(page, component);
+          }
+
+          const screenshot = await page.screenshot();
```
bedrich-schindler, line 138: I would consider cropping image to something smaller. It is not necessery to have huge screenshot with small component in left top corner.

See https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1-option-animations

bedrich-schindler (changes requested):

src/components/Toggle/__tests__/Toggle.spec.tsx
```diff
@@ -0,0 +1,175 @@
...
+      [
+        ...propTests.defaultComponentPropTest,
+        ...propTests.labelPropTest,
+        ...propTests.isLabelVisiblePropTest,
+        ...propTests.labelPositionPropTest,
+        ...propTests.renderAsRequiredPropTest,
+        ...propTests.requiredPropTest,
+        ...mixPropTests([
+          disabledWithCheckedPropTest,
+          propTests.variantPropTest,
```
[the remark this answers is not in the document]
lukasbriza: Sorry I overlooked it.

src/components/CheckboxField/__tests__/_propTests/checkedPropTest.ts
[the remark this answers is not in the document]
lukasbriza: checkPropTest added to global tests

src/components/Toggle/__tests__/_propTests/disabledWithCheckedPropTest.ts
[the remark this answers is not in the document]
lukasbriza: replace with mix disabledProps and checkedProps

src/components/Toggle/__tests__/Toggle.spec.tsx
```diff
@@ -0,0 +1,175 @@
...
+  ToggleForTest,
+  ToggleForRefTest,
+  ToggleForFormLayoutTests,
+} from './Toggle.story';
+import type { ToggleForFormLayoutTestsProps } from './Toggle.story';
+import { disabledWithCheckedPropTest } from './_propTests/disabledWithCheckedPropTest';
+
+test.describe('Toggle', () => {
+  test.describe('base', () => {
+    test.describe('visual', () => {
```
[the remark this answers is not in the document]
lukasbriza: Not sure if I get it right, but I:
1) added to all components that support it mix(validationStatePropTest, requiredPropTest) and mix(renderAsRequiredPropTest, validationStatePropTest)
2) added helperTextAndValidationTextPropType combinations test that replaced mix(validationTextPropTest, helpTextPropTest) and mix(validationTextPropTest, validationStatePropTest) and mix(helpTextPropTest, validationStatePropTest) => because we do not want to test color combinations for combinations of validationText and helperText.

src/components/SelectField/__tests__/SelectField.spec.tsx
lukasbriza: It is good point, but after testing. When I add clip property to screenshot, the options sliders became invisible. And that was reason why I created this "fullpage" test.

src/components/SelectField/__tests__/SelectField.spec.tsx
```diff
@@ -0,0 +1,283 @@
...
+          <SelectFieldForTest
+            onChange={() => { changeCalled = true; }}
+            options={baseOptions}
+            value={baseOptions[0].value}
+          />,
+        );
+
+        const select = component.getByRole('combobox');
+        await select.focus();
+        await select.press('ArrowDown');
```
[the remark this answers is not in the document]
lukasbriza: Added to same test.

tests/playwright/propTests/labelPropTest.tsx
mbohal: I liek this:

```
    someProp:string`
    someProp:bool=true`
    someProp:bool=false`
    someProp:shape[flat]
    someProp:shape[nested]
```

lukasbriza:
I had to add `await page.waitForTimeout(500);` inside `onBeforeSnapshot`, and also include the `{ maxDiffPixelRatio: 0.001 }` option in `toMatchSnapshot` within the full-page block to capture snapshots of the opened `SelectField`. These tests are getting rewritten on every update, but I have no idea why. I think it might be some inconsistency or an issue with the slider option and how it's rendered in the browser or Playwright.

tests/playwright/propTests/labelPropTest.tsx
lukasbriza: Transformed all name of propTest as @mbohal  suggested.

bedrich-schindler:
> I had to add await page.waitForTimeout(500); inside onBeforeSnapshot, and also include the { maxDiffPixelRatio: 0.001 } option in toMatchSnapshot within the full-page block to capture snapshots of the opened SelectField. These tests are getting rewritten on every update, but I have no idea why. I think it might be some inconsistency or an issue with the slider option and how it's rendered in the browser or Playwright.

@lukasbriza, I suggested change over there:  

image

Did you try it?

lukasbriza:
> > I had to add await page.waitForTimeout(500); inside onBeforeSnapshot, and also include the { maxDiffPixelRatio: 0.001 } option in toMatchSnapshot within the full-page block to capture snapshots of the opened SelectField. These tests are getting rewritten on every update, but I have no idea why. I think it might be some inconsistency or an issue with the slider option and how it's rendered in the browser or Playwright.
> 
> @lukasbriza, I suggested change over there:
> image
> 
> Did you try it?

Yes i tried it. I added `{ animations: "disabled" }` into every screenshot method in SelectField tests and it did not worked on that issue. Snapshots are still rewritten every time. Adding `{ maxDiffPixelRatio: 0.001 }` was only thing that helped in combination with timeout on  500ms.

src/components/Alert/__tests__/Alert.spec.tsx-snapshots/Alert-visual-color-danger-1-chromium-linux.png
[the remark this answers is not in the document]
lukasbriza: On call with @bedrich-schindler  decided that no reset will be applied.

tests/playwright/propTests/helperTextAndValidationTextPropType.ts
```diff
@@ -0,0 +1,9 @@
+import type { PropTests } from '../types';
+import { mixPropTests } from '../utils/mixPropTests';
+import { helpTextPropTest } from './helpTextPropTest';
+import { validationTextPropTest } from './validationTextPropTest';
+
+export const helperTextAndValidationTextPropType: PropTests = mixPropTests([
```
bedrich-schindler, line 6: We talked about following combinations yesterday, should looks like this:

```js
import type { PropTests } from '../types';

export const helperTextAndValidationTextPropType: PropTests = [
  {
    name: 'helpText:string & validationText:undefined',
    props: {
      helpText: 'Some normal helpText.',
      validationText: undefined,
    },
  },
  {
    name: 'helpText:undefined & validationText:string',
    props: {
      helpText: undefined,
      validationText: 'Some normal validationText.',
    },
  },
  {
    name: 'helpText:string & validationText:string',
    props: {
      helpText: 'Some normal helpText.',
      validationText: 'Some normal validationText.'
    }
  }
];
```

I hope that covers alls.

tests/playwright/propTests/validationStatePropTest.ts
```diff
@@ -0,0 +1,16 @@
+import type { PropTests } from '../types';
+
+export const validationStatePropTest: PropTests = [
```
bedrich-schindler, line 3: We have talked about this yesteaday and we found out that this is missing:

```js
{
    name: 'validationState:null',
    props: { validationState: null },
  },
``` 

@lukasbriza will also check if there are cases when `null` works as one of parameters. Please, if you found more props like `validationState`, contact me to discuss it first over Slack.

bedrich-schindler (changes requested):

tests/playwright/propTests/validationStatePropTest.ts
lukasbriza: I added to this tests object like:
`
{
    name: 'validationState:undefined',
    props: { validationState: undefined },
  },
`
Because the `null` type here serves as a representation of an undefined value and functions in the same way as JavaScript's default `undefined` value.

bedrich-schindler:
> I had to add await page.waitForTimeout(500); inside onBeforeSnapshot, and also include the { maxDiffPixelRatio: 0.001 } option in toMatchSnapshot within the full-page block to capture snapshots of the opened SelectField. These tests are getting rewritten on every update, but I have no idea why. I think it might be some inconsistency or an issue with the slider option and how it's rendered in the browser or Playwright.

I have tried several things but unfortunately, I have the same problem as you report. I do not know why it worked when I tried it with disabled animations. It may be the flakiness.

That being said, please [create bug report into the Playwright](https://github.com/microsoft/playwright/issues/new/choose). You must provide reproducible demo with the bug. I already removed the reproducible repository, but I have local copy of that. I will send it to you so you can just replace my issue with this bug and sent it do Playwright team so than can review that and hopefully fix that or give us a hint.

Example of the bug report: https://github.com/microsoft/playwright/issues/34441

Issue #609, opened 2025-03-12
Update symlinks


lukasbriza:
Remove RELEASING.md and CONTRIBUTING.md in root.
Add links to README.md into docs/*
Verify links to be consistent.

PR #610, opened 2025-03-12
Update symlinks and remove md files from root (#609)

lukasbriza:
Remove RELEASING.md and CONTRIBUTING.md and updating corresponding symlinks in docs/*.


lukasbriza:
@mbohal Should all links referencing parts of our documentation be internal links? I added links to the testing and releasing guidelines as external links, but for the testing link specifically, there’s no testing-guidelines.md file in master, so the GitHub check will fail due to a dead link. I'm wondering if it wouldn't be better to have all such references as internal links for consistency?

mbohal:
> I'm wondering if it wouldn't be better to have all such references as internal links for consistency?

I think we want this. We need links to work in IDE and if relevant also in the mkdocs output.
I think we need to use the relative syntax (`../../abc`) for this.

mbohal (approved):

lukasbriza:
After yesterday's call with @bedrich-schindler , it was decided to merge this task to enable further work on documenting the rules for writing Playwright tests.

PR #614, opened 2025-03-24
Introduce `color` property to the `Modal` component (#468)

### LINKED ISSUE #468: Add support of feedback colors to `Modal`

adamkudrna:
- `Modal` can now be colored via the `color` prop, using a value from the Feedback Colors collection.
   New custom properties in shape of `--rui-Modal--<COLOR>__<PROPERTY>` have been introduced where:
   - `<COLOR>` is a value from supported color collections (check the docs to see which collections are supported),
   - `<PROPERTY>` is one of `border-color` or `background-color`.
- Add description of our color collections so it's clearer when they can be used.

Closes #468.

### DIFF
--- a/src/components/Modal/Modal.module.scss
+++ b/src/components/Modal/Modal.module.scss
@@ -7,6 +7,7 @@
 @use "../../styles/theme/typography";
 @use "../../styles/tools/accessibility";
 @use "../../styles/tools/breakpoint";
+@use "../../styles/tools/collections";
 @use "../../styles/tools/reset";
 @use "../../styles/tools/spacing";
 @use "animations";
@@ -81,5 +82,15 @@
     .isRootPositionTop {
         top: var(--rui-local-outer-spacing);
         bottom: auto;
+    }
+
+    @each $color in settings.$colors {
+        @include collections.generate-class(
+            $prefix: "rui-",
+            $component-name: "Modal",
+            $variant-name: "color",
+            $variant-value: $color,
+            $properties: settings.$themeable-properties,
+        );
     }
 }
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -690,6 +690,90 @@ React.createElement(() => {
 });
 ```
 
+## Color Variants
+
+Modal can be colored using the `color` prop. The `color` prop implements the
+[Feedback color collection](/docs/foundation/collections#colors)
+and is applied to the border of the modal and the modal footer.
+
+```docoff-react-preview
+React.createElement(() => {
+  const [modalOpen, setModalOpen] = React.useState(false);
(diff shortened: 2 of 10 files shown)


src/components/Modal/Modal.module.scss
bedrich-schindler, line 1: As I see, we did not specify which parameters should be configurable. When I see final result, it does not seem nice visually.

image

To compare it with RADES library:

image

Practically, we will need this to have configurable in our projects otherwise. Otherwise this change will not suit our needs that initialized this issue. I would recommend enabling/disabling followings:

* modal border (around the modal)
* footer top border
* header bottom border

This extension would allow enabling/disabling specific borders as we need. We do not need to specify color but just width or something like this to enable/disable specific border. 

What do you think @adamkudrna? I know that this should be solved before the issue was implemented and I am OK with postponing those changes and implementing them in separate issue. But practically, this implementation will not help us as we would not be able to configure our Dialogs without hacking it.

This might require change of logic of `--rui-Modal__separator__color` and `--rui-Modal__separator__width` properties

src/components/Modal/README.md
```diff
@@ -313,6 +313,7 @@ React.createElement(() => {
             <ModalBody>
               <ModalContent>
                 <Radio
+                  id="modal-footer-justify"
```
mbohal, line 316: Do we need this `id`?

src/components/Modal/Modal.module.scss
mbohal: > This extension would allow enabling/disabling specific borders as we need. We do not need to specify color but just width or something like this to enable/disable specific border.

Presently for RACOM we only need the border around the `Modal`. The header/footer separators (`separators`) are not needed now. I think at least…

Generally we can:

1. Remove the `separators` and be done with it.
2. Add CSS custom variable to toggle each one of the `separators` and use the same border width as the border around modal
3. Add CSS custom variable to set each one of the `separators` border width where `0` would effectively disable them.

I think 1) should be enough for now. Support for `separators` can be added in the future if need arises and it can be paid for by whoever needs it. Its not just the time investment, but also from the perspective of our project it is effectively dead code that requires upkeep.

However if we need/want the `separators`, I have no strong preference between 2) and 3), I would go with whatever @adamkudrna decides is more consistent wit the rest of RUI. If pressured I would go with 3) (I think this is what @bedrich-schindler was suggesting)

mbohal (changes requested):

src/components/Modal/Modal.module.scss
mbohal: On call @adamkudrna explained that 2) is already implemented, except not for each separator separately, but for both.

PR #616, opened 2025-03-24
Drop Neutral Colors collection from `Alert` and `Card` (#615)

### LINKED ISSUE #615: Drop Neutral Colors collection from `Alert` and `Card` components
1. We have no use for dark component variants.
2. We could get rid of the link color switch (dark links for light backgrounds and vice versa) in our Sass code.
3. If we need dark component variants in the future, themes are probably a better and more robust solution.

adamkudrna:
We found the Neutral Colors collection insufficiently flexible to use, so we are removing it from more complex components where its usage was unclear and confusing.

Features:

- New `--rui-Card__background-color` custom property has been introduced to define `Card`'s default background color.

Breaking changes:

- `Alert`: light and dark colors (Neutral Colors collection) have been dropped with no replacement.
- `Card`: light color now corresponds to the default appearance. The dark color option has been dropped with no replacement.

Migration:

- `Alert`: pick another color from the Feedback Colors collection.
- `Card`: remove `color="light"` to keep the default light appearance. If using `color="dark"`, pick another color from the Feedback Colors collection.

Closes #615.

### DIFF
--- a/src/components/Card/README.md
+++ b/src/components/Card/README.md
@@ -148,7 +148,7 @@ for card content.
 ## Color Variants
 
 To cover all possible needs of your project, Card is available in colors from
-[Feedback and Neutral color collections](/docs/foundation/collections#colors).
+[Feedback color collection](/docs/foundation/collections#colors).
 
 ```docoff-react-preview
 <Card color="success">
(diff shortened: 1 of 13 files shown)


src/components/Card/README.md
```diff
@@ -211,26 +211,6 @@ To cover all possible needs of your project, Card is available in colors from
     <Button label="Read more" priority="outline" color="note" />
   </CardFooter>
 </Card>
-<Card>
```
bedrich-schindler, line 214: I am just thinking about `Color Variants` generally. You removed `light` color, but it is still available through `null`, but this is not part of `Color Variants`. It is true that this is same for all components, but on the other hand, `null` is also color variants. This applies to Modal and other components as well.

This is just idea. If you think that it is valid, you can create an issue for that. If not, just let it be as it is. Anyway this is definitely scope of this issue.

bedrich-schindler (approved):

src/components/Card/README.md
adamkudrna: I see it the same as @mbohal described. What we need is this:

1. a neutral (light) variant for general purposes,
2. an option to highlight selected components with our feedback colors.

The key difference is that while the feedback variants define which border color or background color should be used, there is no such rule for the neutral appearance that would be applicable for all components. I mean I tried, but, with `Modal` colors (#468), it turned out it doesn't scale.

PR #617, opened 2025-03-24
Use native font stack (#417)

### LINKED ISSUE #417: Switch to system fonts

adamkudrna:
It ensures that the text is displayed in the font that the user is most comfortable with, and it also reduces the size of the data transferred.

Breaking change:

The Titillium web font has been replaced with native font stack.

Migration:

To keep the Titillium web font in your project, load the web font and override the `--rui-font-family-base` custom property:

```
--rui-font-family-base: 'Titillium Web', helvetica, roboto, arial, sans-serif;
```

Closes #417.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,16 +53,7 @@ To install React UI in your app:
    npm install --save @react-ui-org/react-ui
    ```
 
-2. Load the Titillium Web font:
-
-   ```html
-   <link
-     href="https://fonts.googleapis.com/css2?family=Titillium+Web:wght@300;400;600;700&display=swap"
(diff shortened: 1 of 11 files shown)


mbohal (approved):

adamkudrna:
@bedrich-schindler I moved the font customization to the `Customize` section for better discoverability.

However, I don't the font is missing from the places you mentioned. I mean, I deleted it so it's missing now. But when you look at the docs as a newcomer, you don't have to think about the font until you actually want to change it. Then you head to the Customize section.

Issue #627, opened 2025-04-28
FileInputField does not react to a form reset event


atmelmicro:
The FileInputField does not react to a form reset event when it is in a form.

Repro:
```jsx
<form>
  <FileInputField
    fullWidth
    id="my-file"
    label="First name"
    onFilesChanged={() => {}}
  />
  <input type="file" />
  <button type="reset">Reset</button>
</form>
```

When a file is added to both of the inputs, only the native one will reset when the reset button is pressed.

PR #629, opened 2025-04-29
Fix `FileInputField` not resetting properly (#627)

atmelmicro:
Fix `FileInputField` not resetting properly on `reset` event inside of a form

### DIFF
--- a/src/components/FileInputField/FileInputField.jsx
+++ b/src/components/FileInputField/FileInputField.jsx
@@ -1,5 +1,7 @@
 import PropTypes from 'prop-types';
 import React, {
+  useCallback,
+  useEffect,
   useContext,
   useImperativeHandle,
   useRef,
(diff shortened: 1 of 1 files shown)


src/components/FileInputField/FileInputField.jsx
```diff
@@ -99,6 +101,30 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+    }
+
+    const { form } = inputEl;
+    if (!form) {
+      return;
+    }
+
+    form.addEventListener('reset', handleReset);
+
+    // eslint-disable-next-line consistent-return
```
bedrich-schindler, line 122: Please, remove the `eslint-disable` and replace `return;` with `return () => {};` to prevent need of this rule. This is how we do it across RUI.

bedrich-schindler (approved):
This

Issue #630, opened 2025-04-29
Unable to manually reset value of `FileInputField`


bedrich-schindler:
Reset using form's input is being implemented in https://github.com/react-ui-org/react-ui/issues/627.  There is one more problem with reset and it is reset of its internal value. For native input, this can be achieved by setting `value` to `null` over its ref. This is not possible for `FileInputField`.

**The aim of this issue is to introduce reset over its ref.**

Aditional improvements to be discussed in the future.

PR #631, opened 2025-04-29
Allow to reset `FileInputField` internal state by  calling `resetState` function on its ref (#630)

bedrich-schindler:


src/components/FileInputField/FileInputField.jsx
```diff
@@ -37,20 +37,30 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
 
+  const internalInputRef = useRef();
+
+  // We need to have a reference to the input element to be able to call its methods,
+  // but at the same time we want to expose this reference to the parent component for
+  // case someone wants to call input methods from outside the component.
+  useImperativeHandle(
+    ref,
+    () => {
+      const inputEl = internalInputRef?.current ?? {};
```
bedrich-schindler, line 55: I did not use spread syntax on purpose and ESlint is OK with that as well. Reason is that it would somehow change final shape of the object. If you return it as it is, it look like HTML input in the console. If you change it to the spread syntax, it would return object with bunch of attributes. In the end, it will works the same, but somehow Chrome presents it differently.

I think that this is not necessary to document in the code as it only improves debugablitiy for end user.

lukasbriza (approved):

src/components/FileInputField/FileInputField.jsx
```diff
@@ -37,20 +37,30 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
+      inputEl.resetState = () => {
+        setSelectedFileNames([]);
```
atmelmicro, line 57: `resetState` should call `onFilesChanged([])` - I would use the `handleReset` function here.

src/components/FileInputField/README.md
```diff
@@ -236,6 +236,37 @@ set the `multiple` prop to `true`.
 />
 ```
 
+## Resetting Input State
+
+If you need to reset the input state, you can do it by calling the
+`resetState` method on the component's ref.
+
+```docoff-react-preview
+const fileInputRef = React.useRef();
```
atmelmicro, line 245: This example does not work. It should be wrapped in a `React.createElement(() => { ... })`

Screenshot 2025-04-30 at 12 29 26

src/components/FileInputField/README.md
```diff
@@ -236,6 +236,37 @@ set the `multiple` prop to `true`.
...
+
+return (
+  <>
+    <FileInputField
```
atmelmicro, line 249: The `FileInputField` element should have an `id`.

Screenshot 2025-04-30 at 12 30 45

src/components/FileInputField/README.md
bedrich-schindler: I was expecting to use only syntax highlight feature, but used live preview instead. In the end, I end up with improved example . Thanks for catching this!

src/components/FileInputField/FileInputField.jsx
bedrich-schindler: I changed as you wish.

The only difference is that now `onFilesChanged` handler is called. I am not sure whether it is wanted or unwanted behaviour. But I decided to stick with it.

src/components/FileInputField/FileInputField.jsx
```diff
@@ -37,20 +37,30 @@ export const FileInputField = React.forwardRef((props, ref) => {
...
   const translations = useContext(TranslationsContext);
 
   const [selectedFileNames, setSelectedFileNames] = useState([]);
   const [isDragging, setIsDragging] = useState(false);
 
+  const internalInputRef = useRef();
+
+  // We need to have a reference to the input element to be able to call its methods,
+  // but at the same time we want to expose this reference to the parent component for
+  // case someone wants to call input methods from outside the component.
```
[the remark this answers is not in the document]
bedrich-schindler: I did not change this, just changed the order of the code. But fixing this anyway.

PR #636, opened 2025-05-05
Refactoring/miscellaneous jest to playwright

### LINKED ISSUE #595: Transform `Miscellaneous` (excluding `Modal`, `Popover`, `ScrollView`) components from Jest to Playwright CT
Follow transformation guidelines defined in https://github.com/react-ui-org/react-ui/issues/591

Transform following components without `Modal`, `Popover`, `ScrollView` (that are is separate issue)

Image

If any component from this group is transformed, check whether it is OK or need improvement.

lukasbriza:
Refactor jest tests from miscellaneous section into playwright
Closes #595


src/components/Badge/__tests__/Badge.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  test.describe('visual', () => {
+    [
+      ...propTests.defaultComponentPropTest,
+      ...labelPropTest,
+      ...mixPropTests([
+        [
+          ...propTests.feedbackColorPropTest,
+          ...propTests.neutralColorPropTest,
+        ],
+        propTests.priorityPropTest,
```
bedrich-schindler, line 23: There is subset of priorities (filled and outline, flat is not present). Remove testing of priority=flat.

Please, review generated snapshots against documentation. Ideally, this should be caught by you during self review.

src/components/Card/__tests__/Card.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from '../../../../tests/playwright';
+import {
+  CadForTest,
+  CadWithScrollableForTest,
+} from './Card.story';
+import { densePropTest } from './_propTests/densePropTest';
+
+test.describe('Card', () => {
+  test.describe('visual', () => {
+    [
```
bedrich-schindler, line 18: Card is composition of body and footer. I would expect to test card with body only and card + footer (this is already tested). This is what documentation says. It is also possible to include footer only, but it does not make much sense to be  - documentation does not forbid this, but I would skip such test.

tests/playwright/propTests/raisedPropTest.ts
```diff
@@ -0,0 +1,8 @@
+import type { PropTests } from '../types';
+
+export const raisedPropTest: PropTests = [
+  {
```
bedrich-schindler, line 4: Please, follow the rules set in parent issue. Again, boolean values must have both values defined in prop tests.

src/components/Tabs/__tests__/TabItem.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+      );
+
+      expect(component.locator(`li[id=${id}]`)).toBeDefined();
+      await expect(component.getByRole('link')).toHaveAttribute('id', `${id}__link`);
+      await expect(component.getByText(label)).toHaveAttribute('id', `${id}__label`);
+    });
+  });
+
+  test.describe('functionality', () => {
+    test('call on click callback', async ({ mount }) => {
```
bedrich-schindler, line 66: `onClick` is optional, but I am glad you test it! Neverhless, if it is possible, test also `href` functionality which is default behavior. See `TextLink.spec.tsx` where everything is tested. Also, keyboard accessibility should be part of testing as well.

bedrich-schindler (changes requested):

src/components/Badge/__tests__/Badge.spec.tsx
lukasbriza: I think that there should be specified in documentation, that there is only subset of values, like it is specified in color property. During self review I try to focus primary on differences in screenshots and if prop change is reflected in screenshot.

src/components/Tabs/__tests__/TabItem.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+} from '../../../../tests/playwright';
+import { TabItemForTest } from './TabItem.story';
+import { isActivePropTest } from './_propTests/tabItem/isActivePropTest';
+
+test.describe('Tabs', () => {
+  test.describe('visual', () => {
+    [
+      ...propTests.defaultComponentPropTest,
+      ...propTests.afterLabelPropTest,
+      ...mixPropTests([
```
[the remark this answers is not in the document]
lukasbriza: I decided to combine the isActive prop test with a visual change in the component using an icon. I wanted to avoid generating unnecessary screenshots—testing the combination of isActive and afterLabel felt redundant, since the class binding for isActive has already been verified, and there's a separate test for the afterLabel property. But of course, I can include it if needed.

tests/playwright/propTests/raisedPropTest.ts
bedrich-schindler: I am sorry if it's missing. Nevertheless, we do this for all the boolean values as propTests are independent to default component props.

src/components/Card/__tests__/Card.spec.tsx
lukasbriza: @bedrich-schindler  I think that is should be like you mentioned.

src/components/Card/__tests__/_propTests/densePropTest.ts
```diff
@@ -0,0 +1,8 @@
+import type { PropTests } from '../../../../../tests/playwright/types';
+
+export const densePropTest: PropTests = [
+  {
+    name: 'dense:boolean=true',
```
bedrich-schindler, line 5: Chybí dense = false.

bedrich-schindler (changes requested):

bedrich-schindler (approved):

PR #655, opened 2025-06-24
Export isChildrenEmpty from RUI.

lukasbriza:
Export function `isChildrenEmpty` to be used outside of RUI project.


src/helpers/isChildrenEmpty/README.md
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div>Children not provided</div>
+  );
+});
+```
+
+```docoff-react-preview
+React.createElement(() => {
+  const children = "Meaningful content";
+  const isEmpty = isChildrenEmpty(children);
```
bedrich-schindler, line 40: I would consider replacing this with more React-style example. E.g.:
```
const children = (
  <>
    <h1>Title</h1>
    <p>Content</p>
  </>
)
```
This is the example from real life implementation of component and represents advanced usage

PR #658, opened 2025-06-27
Add text when the user is dragging files over FileInputField

atmelmicro:


mbohal (approved):

src/translations/en.js
```diff
@@ -5,6 +5,7 @@ export default {
   FileInputField: {
     browse: 'Browse',
     drop: 'or drop file here',
+    dropHere: 'Drop here!',
```
bedrich-schindler, line 8: I know that @davidurban  designed it with `Drop here!`, but I would expect `Drop file here` instead to match with translation `Browse or drop file here`. Subject is missing here. 

What do you think @atmelmicro @mbohal?

@davidurban has time off now so we need to decide it without him.

mbohal: I agree with @bedrich-schindler

bedrich-schindler:
When last issue is resolved, you are free to squash and merge it

PR #666, opened 2025-10-16
Change default properties from `null` to `undefined` (#662)

### LINKED ISSUE #662: Use `undefined` instead of `null`
Proposal from @bedrich-schindler and @mbohal.

In dependent projects, we often question whether `null` is correct default value for "empty values" and we ofter suggest to use `undefined` instead. This is something we would like to introduce into React UI.

Main advantage is that `undefined` does not propagate into DOM, while `null` does and it results in console errors.

lukasbriza:
Change all component null default properties to default.

Closes #662


bedrich-schindler:
@lukasbriza, Playwright tests are failing. Can you please make it passing before we just on code review?

bedrich-schindler (approved):

PR #670, opened 2025-10-30
Add popover z index property.

lukasbriza:


adamkudrna (changes requested):
I don't think this approach is robust and systematic. `z-index` is one thing, but current stacking context is actualy the key for it to work. Setting a higher `z-index` does not help when you are locked inside a stacking context. We need to make sure we are rendering the `Popover` either a) in the highest stacking context (usually solved via React Portal), or b) in the browser's top layer.

https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Positioned_layout/Stacking_context

adamkudrna:
Closing in favour of a more robust solution than `z-index`.

Issue #575, opened 2024-10-22
Update to React 18.3


bedrich-schindler:
Update codebase to React 18.3 in preparation of React 19

See React 18.3 deprecation warnings and https://react.dev/blog/2024/04/25/react-19-upgrade-guide

bedrich-schindler:
We generally support `18.x`. Restriction to `~18.2.0` came up from internal projects as we did not want to have deprecation warnings in clients. As we speak about this, we should have set something like `>=18.0.0 && <= 18.3.0` as version 18.3.0 would result in warnings. ~18.2.0 is to restrictive. After the upgrade to 18.3, we can cagenerally support `18.x`.

After upgrade to React 19, we can either support only `19.x` or support both. But it would required testing with both versions.

Am I right?

mbohal:
> Am I right?
Yes, you are right.

> we can either support only 19.x or support both
I guess this is what I meant. I guess suppose we could support both versions if we have visual tests? Would you consider that good enough testing? I think I would.

adamkudrna:
I will take whatever you two agree on. Go ahead.

adamkudrna:
@bedrich-schindler We are probably going straight for React 19, right?

bedrich-schindler:
@adamkudrna, not sure yet. If we want, we can support both. Important thing for us are prop types that are being deprecated in 18.3 and removed in 19. If we remove prop types, we can support both without raising errors.

PR #677, opened 2025-12-05
Update dependencies (#674) + Support Node 24 LTS (#678)

bedrich-schindler:
The rest should be updated in #575, #675, #676


package.json
```diff
@@ -29,12 +29,12 @@
...
   },
   "engines": {
-    "node": ">=20"
+    "node": ">=22"
   },
   "devEngines": {
     "runtime": {
       "name": "node",
-      "version": "22.x",
+      "version": "22.x || 24.x",
```
bedrich-schindler: For some reason, single version worked for matrix testing of both Node 20 and Node 22 (see `.github/workflows/build.yml`), but it does not work for Node 22 and Node 24. So I must entered both version (and not just latest one) to make it working. But I think this is correct approach.

bedrich-schindler: cc @adamkudrna @mbohal

PR #689, opened 2026-02-09
Refactor Sass inline `if()` to `@if` blocks to get ready for native CSS `if()`

adamkudrna:
After this change, logs should be free of the similar warnings:

```
Deprecation Warning on line 65, column 19 of file:///workspace/node_modules/@react-ui-org/react-ui/src/styles/tools/_collections.scss:65:19:
The Sass if() syntax is deprecated in favor of the modern CSS syntax.

Suggestion: if(sass($modifier-value): "--" + $modifier-value; else: "")

More info: https://sass-lang.com/d/if-function
```

### DIFF
--- a/tests/playwright/env/parseDotEnvFile.ts
+++ b/tests/playwright/env/parseDotEnvFile.ts
@@ -6,7 +6,10 @@ const PLAYWRIGHT_ENV_FILE = '.env.playwright';
  * Load and parse Playwright environment file into an object.
  */
 export const parseDotEnvFile = () => {
-  const dotEnvConfigOutput = dotenv.config({ path: PLAYWRIGHT_ENV_FILE });
+  const dotEnvConfigOutput = dotenv.config({
+    path: PLAYWRIGHT_ENV_FILE,
+    quiet: true,
+  });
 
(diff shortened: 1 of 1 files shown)


bedrich-schindler:
It's great we are getting rid of deprecation warnings, but I would like to point out that clients are polluted with `mixed-decls` longer (and more) than with `if-function`. Are you opening second PR for that fix?

src/styles/tools/_collections.scss
```diff
@@ -202,14 +211,15 @@
...
-        if(
-            $modifier-name and $modifier-value,
-            ".isRoot#{rui-string.capitalize($modifier-name)}#{rui-string.capitalize($modifier-value)}",
-            ""
-        );
+    $modifier-class-name: "";
     $variant-class-name: ".isRoot#{rui-string.capitalize($variant-name)}#{rui-string.capitalize($variant-value)}";
 
+    @if $modifier-name and $modifier-value {
+        $modifier-class-name: ".isRoot"
```
bedrich-schindler: My knowledge is not sufficient to review this, I need to educate on this topic. If connecting strings outside of "", do we need to wrap function inside `#{...}` as you did?

Issue #594, opened 2025-02-21
Transform `Layouts` components from Jest to Playwright CT


bedrich-schindler:
Follow transformation guidelines defined in https://github.com/react-ui-org/react-ui/issues/591

Transform following components:

<img width="199" alt="Image" src="https://github.com/user-attachments/assets/c85b28a6-f17d-469c-bde0-ba858d63a5ed" />

This is quite different group of components which might require different testing approach. Especially `FormLayout` would be atypical. Do not hesitate to consult proof of concept before spending exhaustive time on the work.

PR #699, opened 2026-03-18
Auto-hide labels of box fields nested in `FormLayoutCustomField` via context

adamkudrna:
## Summary

- Introduces `FormLayoutCustomFieldContext` and provides it from
  `FormLayoutCustomField`, so `TextField`, `TextArea`, `SelectField`, and
  `FileInputField` automatically hide their own label when nested inside
  `FormLayoutCustomField` — no more manual `isLabelVisible={false}` needed.
  **This is a breaking change**: any nested field that previously showed its
  own label will now always have it hidden, since `FormLayout` forces
  horizontal layout on all descendants and a visible nested label would
  conflict with the grid regardless of `fieldLayout`.
- Adds a "Nested Fields Labels" section to the `FormLayout` README
  documenting the automatic behavior.
- Adds `labelWidth="100px"` visual test coverage (via a real `FormLayout`
  wrapper) for `Button`, `CheckboxField`, `FileInputField`, `Radio`,
  `SelectField`, `TextArea`, `TextField`, and `Toggle`, and renames the
  previous `30px` FormLayout label-width test case to the more representative
  `100px`.
- Adds `Button` to the `FormLayout` story.
- Removes now-redundant `isLabelVisible={false}` wiring from custom-field
  prop tests since label-hiding is now automatic.


mbohal (changes requested):

src/components/FormLayout/README.md
```diff
@@ -330,6 +330,30 @@ and its styles may affect contained form fields through CSS cascade, don't
 forget to mirror the aforementioned properties to the contained form fields too
 as API options as such are **not** inherited.
 
+### Nested Fields Labels
+
+When placing box form fields — TextField, TextArea, SelectField, or
+FileInputField — inside a FormLayoutCustomField, we recommend to hide their
```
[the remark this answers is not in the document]
adamkudrna: Now the form field labels inside `FormLayoutCustomField` hide automatically. It's a BC though.

src/components/Radio/__tests__/Radio.story.tsx
```diff
@@ -37,6 +40,13 @@ export const RadioForTest = ({
   />
 );
 
+export const RadioForFormLayoutLabelWidthTests = () => (
```
bedrich-schindler, line 43: General question.

Why there two types of form layout story components? One uses `FormLayoutContext` and you introduced `FormLayout`. Why to not we extend `RadioForFormLayoutTests` with `labelWidth` (as it is done with `layout`)?

It would make sense to me to have just one component for such tests as everything can be simplified then. Including just simple propTest instead of writing specific test case in spec file.

We do test component here, so setting `FormLayoutContext` is more correct. Otherwise you test composition of two components.

bedrich-schindler (changes requested):

src/components/Radio/__tests__/Radio.story.tsx
adamkudrna, line 43: labelWidth is CSS-only (grid-template-columns / --rui-custom-label-width on the real FormLayout root, consumed via CSS Grid subgrid), so it needs a real <FormLayout> ancestor — it can't be faked through FormLayoutContext like `layout` can. See reply below for the full explanation.

adamkudrna (commented):
Reply to @bedrich-schindler's comment above about two story component types for FormLayout tests:

Agreed for `layout`/`fieldLayout` — that's purely context-driven (`FormLayoutContext` → `{ layout }`), so `RadioForFormLayoutTests` wrapping in `FormLayoutContext.Provider` is the right, more isolated approach.

`labelWidth` is different: it's CSS-only, set as `grid-template-columns`/`--rui-custom-label-width` on the real `FormLayout` root, and consumed by fields via CSS Grid `subgrid`. That only resolves with a real grid ancestor in the DOM, so it can't be faked through context. `RadioForFormLayoutLabelWidthTests` wrapping in a real `<FormLayout>` is required for that case, not composition-testing overreach.

src/components/Radio/__tests__/Radio.story.tsx
adamkudrna: Agreed for `layout`/`fieldLayout` — that is purely context-driven (`FormLayoutContext` → `{ layout }`), so `RadioForFormLayoutTests` wrapping in `FormLayoutContext.Provider` is the right, more isolated approach.

`labelWidth` is different: it is CSS-only, set as `grid-template-columns`/`--rui-custom-label-width` on the real `FormLayout` root, and consumed by fields via CSS Grid `subgrid`. That only resolves with a real grid ancestor in the DOM, so it cannot be faked through context. `RadioForFormLayoutLabelWidthTests` wrapping in a real `<FormLayout>` is required for that case, not composition-testing overreach.

src/components/FormLayout/FormLayoutCustomField.jsx
```diff
@@ -73,7 +74,9 @@ export const FormLayoutCustomField = ({
         className={styles.field}
         id={id && `${id}__field`}
       >
-        {children}
+        <FormLayoutCustomFieldContext.Provider value>
```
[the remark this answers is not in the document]
adamkudrna: Eslint requires empty value, keeping as is.

src/components/Button/__tests__/Button.story.tsx
```diff
@@ -42,10 +50,29 @@ export const ButtonForRefTest = ({
...
+export const ButtonForFormLayoutTests = ({
+  layout,
+  ...props
+}: ButtonForFormLayoutTestsProps) => {
+  const values = useMemo(() => ({ layout }), [layout]);
+
+  return (
+    <FormLayoutContext.Provider
+      value={values}
+    >
```
mbohal: I think we should add here `<TextField label="Text field" />`. Otherwise the test snapshot just shows the button and it does not test the layout at all. 

Or is omitted on purpose and I'm missing something? Other components have a test named `…visual-label-hidden…` so maybe it is just a misleading name?

src/components/FormLayout/__tests__/FormLayout.spec.tsx-snapshots/FormLayout-base-visual-autoWidth-boolean-false-1-chromium-linux.png
mbohal, line 1: Are we dropping the snapshots on purpose? Why?

mbohal (changes requested):

src/components/Button/__tests__/Button.story.tsx
adamkudrna: Good catch, but adding a `TextField` here wouldn't actually fix it. `ButtonForFormLayoutTests` only wraps `Button` in `FormLayoutContext.Provider` — a plain React context, no DOM element. Button's horizontal-layout style is `.isRootLayoutHorizontal { grid-column-start: 2; }`, which only has a visual effect inside a real CSS grid, and `display: grid` lives on `FormLayout`'s root `<div>`, not on the context provider. So this mount never has a grid ancestor, and the vertical/horizontal snapshots were identical regardless of what's rendered alongside the button.

This is different from fields like `TextField`/`Radio`, where `layout` drives each field's own internal subgrid, so the context-only mount is meaningful for them — just not for `Button`, which relies on positioning from an ancestor grid.

The pre-existing `vertical`/`horizontal` tests (mounting `ButtonInVerticalFormLayoutForTest`/`ButtonInHorizontalFormLayoutForTest`, which use a real `<FormLayout>` with a `TextField`) already cover this correctly, so I removed the redundant `layoutPropTest`-driven block instead of trying to patch it.

src/components/FormLayout/__tests__/FormLayout.spec.tsx-snapshots/FormLayout-base-visual-autoWidth-boolean-false-1-chromium-linux.png
adamkudrna: Intentional cleanup, not a regression. `autoWidth` hasn't actually been exercised by any test since #594 — see the comment at the top of `FormLayout.spec.tsx`'s `base` describe block explaining it was dropped because subgrid support made the issue it addressed obsolete. These two PNGs were leftover artifacts from before that change that never got deleted. I'm touching this snapshot directory anyway in this PR, so I removed the orphaned files.

PR #706, opened 2026-03-27
Introduce `devcontainer` service container based development environment (#705)

bedrich-schindler:
All development is now done inside the container named `devcontainer` which contains all necessary tools and dependencies. The devcontainer orchestrates other service containers behind the scenes via Docker-from-Docker. Other containers are implementation details and should not be accessed directly.

This is breaking change commit for development environment, so it is required to remove all the env files and start with fresh one.

There are two supported ways to access the development environment. Recommended way is to use Development Containers with an IDE. The alternative is to use Docker Compose directly. Local development is officially no more supported.

What more, documentation of contribution and CLAUDE.md was updated to match with the changes.

---

Closes #705


docker/react_ui_devcontainer/files/home/developer/shell-init.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Block SSH agent forwarding if the environment variable `BLOCK_SSH_AUTH_SOCK` is set to "true"
+if [ "$BLOCK_SSH_AUTH_SOCK" = "true" ]; then
```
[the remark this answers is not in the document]
adamkudrna: Maybe worth a comment right in the file?

docker/react_ui_devcontainer/files/usr/local/bin/copilot
adamkudrna, line 1: Don't these "replacement" binary files need to be executable? Maybe you have it elsewhere in the PR, maybe GitHub just does not display file permissions.

docker/react_ui_devcontainer/files/usr/local/bin/mkdocs
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+sudo docker compose -p "$COMPOSE_PROJECT_NAME" exec docs mkdocs "$@"
```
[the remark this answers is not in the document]
adamkudrna: I would put down a comment on this, using `sudo` is very unusual so we should note the reason for our future selves.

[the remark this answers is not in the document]
adamkudrna: > AI from devcontainer cannot connect to other containers on host machine without using sudo

Does it mean AI clients cannot reach the containers at all? Won't it be an unwanted limitation for us when working with AI?

src/docs/contribute/testing-guidelines.md
```diff
@@ -39,52 +35,18 @@ npm run <eslint|markdownlint|stylelint>
...
-from other tools.
-
 #### Configuration
 
-Test parameters can be tweaked by creating and tweaking `.env.playwright` file:
-
-```bash
-cp .env.playwright.dist .env.playwright
-```
+Test parameters can be tweaked by creating and tweaking `.env` file.
```
adamkudrna, line 46: Don't we mention `workers` at least?

docker/react_ui_devcontainer/Dockerfile
```diff
@@ -0,0 +1,128 @@
+FROM debian:bookworm AS react-ui-devcontainer
```
adamkudrna: How about trimming the size a bit (about 40 %)?


```suggestion
FROM debian:bookworm-slim AS react-ui-devcontainer
```

docker/react_ui_devcontainer/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+        > /etc/apt/sources.list.d/docker.list && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends docker-ce-cli docker-compose-plugin && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Switch to the `developer` user for all user-level installations
+USER $USERNAME:$USERNAME
+
+# Install Oh My Zsh
+RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
```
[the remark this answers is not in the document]
adamkudrna: @bedrich-schindler Is it really OK to leave it without `--unattended`?

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,206 @@
...
+Run the setup script to automatically create and configure all necessary files
+and build Docker images:
 
 ```bash
-docker compose run --rm node_shell
+sh setup.sh
 ```
 
-If you want to run single command, run:
+[Development Containers] run this script automatically if the project has not
```
adamkudrna: I would say:

```suggestion
IDEs that support [Development Containers] run this script automatically if the project has not
```

Is it correct?

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,206 @@
...
-#### Node shell
+### Setup
 
-All npm commands such as `npm ci`, `npm test`, `npm run eslint` and others you
-need to run them within the `node_shell` Docker container.
+#### Automatic setup
 
-To log into the container, run:
+Run the setup script to automatically create and configure all necessary files
+and build Docker images:
```
adamkudrna, line 37: I would add a big note with ⚠️ for those what had the project on their machines prior to switching to dev container. If you forget to delete your old `.env` file, the build of the dev container fails because the script cannot find the project name variable and tries to call `_devcontainer` image instead of `react-ui_devcontainer`.

Image

setup.sh
```diff
@@ -0,0 +1,104 @@
...
+if [ ! -f docker-compose.yml ]; then
+  echo "Creating and configuring docker-compose.yml file..."
+  cp docker-compose.yml.dist docker-compose.yml
+
+  DEFAULT_PROJECT_DEVCONTAINER_IMAGE="${DEFAULT_PROJECT_NAME}_devcontainer"
+  PROJECT_NAME=$(grep -E '^COMPOSE_PROJECT_NAME=' .env | cut -d '=' -f 2-)
+  PROJECT_DEVCONTAINER_IMAGE="${PROJECT_NAME}_devcontainer"
+
+  sed_cmd "s|image: $DEFAULT_PROJECT_DEVCONTAINER_IMAGE|image: $PROJECT_DEVCONTAINER_IMAGE|" docker-compose.yml
+else
```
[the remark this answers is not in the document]
bedrich-schindler: I would skip this. If created using `setup.sh`, variable is automatically set in `.env`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "React UI (${localWorkspaceFolderBasename})",
```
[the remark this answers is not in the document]
adamkudrna: @mbohal I'd encourage to use Git Worktree instead of multiple clones.

But the result is the same for this scenario — a directory is created for each worktree, and its name should appear here.

- https://git-scm.com/docs/git-worktree
- https://dev.to/nickytonline/git-worktrees-git-done-right-2p7f

setup.sh
```diff
@@ -0,0 +1,104 @@
...
+  echo "docker-compose.yml file already exists, skipping creation."
+fi
+
+# Build Docker images
+sh ./docker/build-docker-images.sh
```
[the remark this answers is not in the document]
adamkudrna: I agree with your approach. Docker images and container layers are cached, so it shouldn't take long to hook up the environment. And we prefer being up-to-date over fast startups, at least at the time being.

⚠️  BUT — there is something that is being downloaded each time the dev container starts. Its size is over 1 GB so it takes time. I am not sure what it is, maybe it's even JetBrains specific, but I'd try to fix it:

Snímek obrazovky 2026-03-30 v 12 42 04

docker-compose.base.yml
```diff
@@ -0,0 +1,91 @@
...
+      - copilot-config:/home/developer/.config/copilot
+      - opencode-config:/home/developer/.config/opencode
+      - opencode-share:/home/developer/.local/share/opencode
+      - opencode-state:/home/developer/.local/state/opencode
+
+  # This service provides Node environment and NPM
+  node:
+    build: docker/node
+    user: ${COMPOSE_UID}:${COMPOSE_GID}
+    command: sh -c 'if [ "$$COMPOSE_START_JS_FILES_WATCHER_AT_START" = "true" ]; then npm ci && npm start; else sleep infinity; fi'
```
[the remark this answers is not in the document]
adamkudrna: Question: can I stop the server without stopping the dev container, run `npm ci` manually, and restart the server? If not, we need `npm ci` on every dev container start anyway.

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+contains all necessary tools and dependencies. All commands in the documentation
+are container-agnostic and are meant to be run directly inside the `devcontainer`.
 
-This allows running the documentation site which serves as a development platform.
+The `devcontainer` orchestrates other service containers behind the scenes via
+Docker-from-Docker. Other containers are implementation details and should not
+be accessed directly.
 
-### Configure Docker Compose
+There are two supported ways to access the development environment. Recommended
```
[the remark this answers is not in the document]
adamkudrna: As a minimum, I'd start a new paragraph with "The alternative is…".

Nevertheless, in general, I'm not fancy of having two supported approaches — again. That's why we decided for Dev Containers, isn't it?

> … but the IDE will run on the host outside the container,
> so mentioned features may require additional manual configuration.

I would be even more straightforward. I don't want to support this use case for anyone except CI/CD. If you want to use IDE, use the Dev Container. Everything else is at your own risk, and we shouldn't waste the limited capacities of our small team on supporting it.

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+Run the setup script to automatically create and configure all necessary files
+and build Docker images:
 
 ```bash
-docker compose run --rm node_shell
+bash setup.sh
 ```
 
-If you want to run single command, run:
+[Development Containers] run this script automatically if the project has not
```
adamkudrna: ⚠️ This sentence should go **before** the instruction to run the bash command. By the time I would be reading this, I would already have run the bash command…

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+    docker compose exec devcontainer bash
     ```
 
-4. **On host:** Build mkDocs:
+3. To stop the environment:
 
     ```bash
-    docker compose run --rm mkdocs_build_site
+    docker compose down
     ```
```
adamkudrna, line 98: I am OK with keeping it as an escape-hatch instruction. I'd only prefer to introduce it with something like:

> In case the Dev Containers feature is not available in your IDE, you can run the `devcontainer` manually:

And warn users that not everything may work as it would in a real Dev Container.

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+
+## Running
+
+> See `.env` whether both the application and documentation server are not configured
+> to auto-start on container startup. If they are, you can skip the following steps.
+
+To start building JavaScript files in watch mode:
+
+```bash
+npm run start
```
adamkudrna:
```suggestion
npm start
```

?

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+To build the documentation:
+
+```bash
+mkdocs build
+```
+
+## Running
+
+> See `.env` whether both the application and documentation server are not configured
+> to auto-start on container startup. If they are, you can skip the following steps.
```
adamkudrna: They start automatically by default, don't they? If yes, you could say it here.

.env.dist
```diff
@@ -1,13 +1,52 @@
...
+# ⚠️ [Linux] This needs to be set to the output of `id --user`
+# ⚠️ [MacOS] This needs to be set to 1000
 COMPOSE_UID=1000
+# ⚠️ [Linux] This needs to be set to the output of `id --group`
+# ⚠️ [MacOS] This needs to be set to 1000
 COMPOSE_GID=1000
+
+#############################
+# Devcontainer configuration #
+#############################
```
adamkudrna, line 29:
```suggestion
##############################
# Devcontainer configuration #
##############################
```

.env.dist
```diff
@@ -1,13 +1,52 @@
...
+# Select your preferred editor and visual (vim, nano)
+EDITOR=vim
+VISUAL=vim
+
+# Select your preferred shell (/bin/bash, /bin/fish, /bin/zsh)
+SHELL=/bin/bash
+
+###########################
+# Playwright configuration #
+###########################
```
adamkudrna, line 45:
```suggestion
############################
# Playwright configuration #
############################
```

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+npm run build
 ```
 
-Argument `--service-ports` is used to expose the ports of the container to the host
-to serve the test report.
+To build the documentation:
+
+```bash
+mkdocs build
+```
```
adamkudrna, line 216: This is unclear to me. When I start the Dev Container, it should automatically start the dev server (that's the default configuration in `.env.dist`). So, to be able to re-install dependencies and run the build, I need to shut down the server first, right? How should I do it? Could you clarify it?

As I think about it, it seems to me that a section "Stopping" is missing. Or is it me who is missing something?

docker-compose.base.yml
```diff
@@ -0,0 +1,91 @@
...
+    depends_on:
+      node:
+        condition: service_started
+      playwright:
+        condition: service_started
+      docs:
+        condition: service_started
+    user: ${COMPOSE_UID}:${COMPOSE_GID}
+    # Keep the container running indefinitely to allow developers to attach to it and use it as their development environment
+    command: sleep infinity
```
adamkudrna, line 23: RIP ☠️ 🎉

setup.sh
```diff
@@ -0,0 +1,104 @@
...
+  LOCAL_VISUAL=
+  if [ -n "$VISUAL" ]; then
+    sed_cmd "s|^VISUAL=.*|VISUAL=$VISUAL|" .env.temp
+    LOCAL_VISUAL=$VISUAL
+  fi
+
+  cp .env.temp .env
+  rm .env.temp
+
+  echo "Configured .env file with the following values:"
```
adamkudrna, line 68: 👍🏻

adamkudrna (changes requested):
I have a lot of comments, but I really like it! You have done a great job 👏🏻.

docker/react_ui_devcontainer/files/home/developer/shell-init.sh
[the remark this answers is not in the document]
bedrich-schindler: There is already comment mentioning VS Code:

```
# Remove any existing SSH agent socket files created by VS Code
```

This is part that covers VS Code issue. Resetting `SSH_AUTH_SOCK` is just to wipe it out completely.

But sure, I will add complete explanation.

setup.sh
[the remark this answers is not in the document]
bedrich-schindler: I do not have answers yet. This is not an issue with VS Code and it is not happening in WebStorm as well. Do you use `Mount Sources` or `Clone Sources`?

image

`Clone Sources` is not working for me in any repo and JetBrains team stopped responding to my ticket. But `Mount Sources` in 2026.1 does not have the issue you mention.

My guess is that they are publishing IDE backends instantly, so it might get downloaded every time they publish new version of IDE Backend. But it is just guess. It is probably up to you, I am sorry.

I think that this is not related to this PR, so I suggest closing it here.

setup.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+DEFAULT_PROJECT_NAME="react-ui"
+DEFAULT_USER_ID=1000
+DEFAULT_GROUP_ID=1000
+
+set -e
+trap 'echo "Failed to setup project"; rf -f .env.temp; exit 1' ERR
```
adamkudrna: Typo: `rf` should be `rm`. If the script fails mid-run, `.env.temp` won't be cleaned up.

— Claude Code

docker/react_ui_devcontainer/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+
+# Ensure /usr/local/bin takes precedence over VS Code injected paths
+# and run profile script on every shell startup to apply .env configuration
+RUN mkdir -p $HOME/.config/fish && \
+    echo 'export PATH="/usr/local/bin:$PATH"' >> $HOME/.bashrc && \
+    echo 'export PATH="/usr/local/bin:$PATH"' >> $HOME/.zshrc && \
+    echo 'set -gx PATH /usr/local/bin $PATH' >> $HOME/.config/fish/config.fish && \
+    echo '$HOME/shell-init.sh' >> $HOME/.bashrc && \
+    echo '$HOME/shell-init.sh' >> $HOME/.zshrc && \
+    echo '$HOME/shell-init.sh' >> $HOME/.config/fish/config.fish
```
adamkudrna: `shell-init.sh` is executed as a subprocess here, not sourced. When bash/zsh evaluates `$HOME/shell-init.sh` from `.bashrc`/`.zshrc`, it spawns a child process — so `export SSH_AUTH_SOCK=/dev/null` inside the script only affects that subshell. The `BLOCK_SSH_AUTH_SOCK` feature is silently broken for bash and zsh.

Fix: use `. $HOME/shell-init.sh` (dot-source) so the export affects the current shell session.

Note: fish doesn't support sourcing sh scripts natively, so the fish line would need a separate fish-compatible snippet (or be removed).

— Claude Code

docker-compose.base.yml
```diff
@@ -0,0 +1,91 @@
...
+    ports:
+      - ${COMPOSE_PLAYWRIGHT_REPORT_PORT}:9323
+    volumes:
+      - .:/workspace:z
+
+  # This provides server for documentation
+  docs:
+    build: docker/mkdocs
+    user: ${COMPOSE_UID}:${COMPOSE_GID}
+    entrypoint: sh -c 'if [ "$$COMPOSE_START_DOCS_SERVER_AT_START" = "true" ]; then mkdocs serve; else sleep infinity; fi'
```
adamkudrna: The `node` service (line 56) uses `command:` for its conditional start logic, but `docs` uses `entrypoint:` for the same pattern. This inconsistency matters for extending files: `command:` can be overridden in a child Compose file, but `entrypoint:` requires an explicit `entrypoint:` override. Consider making them consistent.

— Claude Code

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,205 @@
...
+the `devcontainer` shell, editor and SSH agent forwarding, as well as application
+settings. See `.env.dist` for available options.
+
+### Accessing the Development Environment
+
+#### Using Development Containers
+
+Open the project in an IDE that supports [Development Containers] (e.g.
+[Visual Studio Code][vscode-devcontainers], [JetBrains IDEs][jetbrains-devcontainers]).
+The IDE will automatically setup the environment using the configuration in
```
adamkudrna: "setup" is a noun/adjective; as a verb it should be two words: "set up".

— Claude Code

.env.dist
```diff
@@ -1,13 +1,52 @@
-###############################
-# Docker compose configuration #
-###############################
+################################
+# Docker Compose configuration #
+################################
 
-# Host system port where the live documentation is to be made accessible
-COMPOSE_START_PORT=8000
+# Docker compose project name must match directory name of the project to allow devcontainer to communicate with other containers
```
adamkudrna: The comment says "must match directory name of the project", but the real constraint is that this value must match the project name used when running the other containers. `setup.sh` derives it from the directory basename, but if someone starts the stack with a custom project name (e.g. `docker compose -p my-name up`), the wrapper scripts inside the devcontainer will silently target the wrong containers. A more accurate description: "must match the Docker Compose project name used to start the other service containers."

— Claude Code

setup.sh
adamkudrna: And again, it should be "set up" when used as verb.

adamkudrna (commented):
Here are a few points by Claude. I believe they are relevant.

(Once you rebase the branch, we can try asking Claude to do CR right here, just by adding the `claude` label.)

docker/react_ui_devcontainer/files/usr/local/bin/mkdocs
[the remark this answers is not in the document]
bedrich-schindler: > AI from devcontainer cannot connect to other containers on host machine without using sudo

Again and simplier. 

`root` is owner of Docker Socket, so if you want to use Docker Socket (e.g. by calling `docker ps`), you need to be either `root` or execute it with `sudo` (`developer` use is part of sudoers, password is not required).

These wrapper use `sudo` to access Docker Socket, so if you execute `mkdocs` command, you call `/usr/local/bin/mkdocs` (this file). This file is executable by every user (see `COPY` command). And this file calls `sudo docker compose ...` so its works OK.

Remember that all this is container agnostic. 

If AI (or you) for some reason (for example due to some attack) tries to access Docker Socket without `sudo`, it will be rejected due to insufficient permissions. If you call it with `sudo`, you gain access to it. But we can tell CLAUDE.md to never use `sudo`, so it should prevent those types of attacks. I do not want to be pessimistic, but this is side effect of the architecture I came up with. This is why I mention it.

So if you want, I can place comments to all the files.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
bedrich-schindler: As discussed in Slack, I think it does not care how you approach it. The result is still the same. It is even not part of documentation and I point it out just to let you know, the reason is not obvious. Bur it would work even without it. But at the end, I like it.

So can we skip this?

docker/react_ui_devcontainer/Dockerfile
bedrich-schindler: I must disagree with this comment, I tested it maybe 10 times in all shells in VS Code and it worked every time. But I will look at it and try it 11th time ...

src/docs/contribute/general-guidelines.md
bedrich-schindler: I mean this whole auto-start functionality is not crafted well. I want to discuss it on Monday call, but there was no time for that and I forgot about it at the same time.

So yes. You need to either restart the devcontainers or kill the process from the terminal. There are also issues at the other end of this chain while auto starting it.

I will need to discuss it at least with one of you on call as I need to have opinion on this. We do not autostart npm start in private repositories by default, but it seemed like "great idea" for opensource project, where you do not need to think about watchers and starting the server.

src/docs/contribute/testing-guidelines.md
bedrich-schindler: I was thinking about it, but decided that I will try to avoid mentioning ENV variables as we might simply change something and forgot to update the docs (as I rewrote complete documentation in 3 repos recently, I guarantee you that those things are first we forgot about).

If you decide you want it, I can add it. But it is mentioned everywhere you need to tweak it. It might not be just workers, but other vars. @mbohal?

docker/react_ui_devcontainer/files/usr/local/bin/copilot
bedrich-schindler: It is solved by `COPY` command in `Dockerfile`:

```
COPY --chmod=+x files/usr/local/bin/node /usr/local/bin/node
```

src/docs/contribute/general-guidelines.md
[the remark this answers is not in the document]
bedrich-schindler: I must strongly disagree. Supposedly, I went into direction where I use as much Docker (Compose) as I can and I configure minimum things in `devcontainer.json`. It is as simple as I suggested. 

There are lot of cases when it does not make sense to run IDE just to do something. E.g. I want to rebuild RUI when linking it on host machine. I want to start local docs server. And so on. Those are terminal only things that does not require to have IDE (started or installed).

Devcontainer is just superset of Docker Compose. it is nothing like when we supported local development AND dockerization. Those were totally different approaches. Devcontainer is just superset.

I though that the suggestion does make sense:

> The alternative is to use Docker Compose directly,
which is suitable for cases when IDE integration is not needed (e.g. CI/CD
or terminal-only workflows), but the IDE will run on the host outside the container,
so mentioned features may require additional manual configuration. 

I am mentioning here what might be the case, that IDE runs outside the container (so the difference from Development Containers aka Devcontaier) and that it may require additional manual configuration. 

But you two with @mbohal decide what you think.

docker-compose.base.yml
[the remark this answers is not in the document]
bedrich-schindler: There is different comment somewhere where you ask how do you stop "process" when it is autostarted. I think that this command is not good and I wanted to have a discussion about it on Monday. But due to Easter holidays, there are two more weeks to call. We need to decide what to do with this ...

.env.dist
bedrich-schindler: In both IDEs, I see it correctly:
image
Terminal displays it incorrectly. Weird.

docker-compose.base.yml
```diff
@@ -0,0 +1,91 @@
...
+    # Start dependent services before starting the `devcontainer` service to ensure that the necessary environments
+    # and tools are available when the `devcontainer` starts.
+    depends_on:
+      node:
+        condition: service_started
+      playwright:
+        condition: service_started
+      docs:
+        condition: service_started
+    user: ${COMPOSE_UID}:${COMPOSE_GID}
```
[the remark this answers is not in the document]
bedrich-schindler: @mbohal ping

docker/react_ui_devcontainer/files/usr/local/bin/mkdocs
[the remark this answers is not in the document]
bedrich-schindler: Added comments to all the wrapper scripts

docker/react_ui_devcontainer/files/home/developer/shell-init.sh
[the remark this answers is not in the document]
bedrich-schindler: Comment added shortly describing behavior of two most common IDEs.

src/docs/contribute/general-guidelines.md
bedrich-schindler: I do not see added value.  What more, Development Containers can be run using CLI only (I do not use it, I do not see why I should use it, but it still possible). So it is not just IDE thing.

src/docs/contribute/general-guidelines.md
bedrich-schindler: Finally found this hidden comment. As commit message says:

> This is breaking change commit for development environment, so it is required to remove all the env files and start with fresh one.

I do not think that this should be not part of documentation, rather the release notes. We have never ever informed users about changes in .env or docker-compose in the past, even though new parameters were introduced.

What do @mbohal think?

src/docs/contribute/general-guidelines.md
bedrich-schindler: @mbohal?

src/docs/contribute/testing-guidelines.md
bedrich-schindler: @mbohal?

src/docs/contribute/general-guidelines.md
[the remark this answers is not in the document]
bedrich-schindler: @mbohal?

setup.sh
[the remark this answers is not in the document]
bedrich-schindler: @adamkudrna, any update?

src/docs/contribute/testing-guidelines.md
mbohal: I think there it is better not to mention the individual configuration parameters in docs because:

1. As @bedrich-schindler point out, it makes maintenance harder
2. It is not needed. I'm told in docs to tweak parameters, which makes it obvious, that I should open the `.env` file, read it and make changes as needed.

I can imagine splitting the `.env` file into two sections:

1. You probably want to change these
2. You probably do not want to change these

src/docs/contribute/general-guidelines.md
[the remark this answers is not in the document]
mbohal: I would drop the:

>  but the IDE will run on the host outside the container,
so mentioned features may require additional manual configuration.

part.

Then it comes out is simple:

1. I want to use IDE -> use devcontainers
2. I do not want to / can not use IDE -> use docker compose directly

That makes it simple.

Also my VS code connects to the previously running  containers (with some issues though) so the sentence is actually kind of wrong.

docker-compose.base.yml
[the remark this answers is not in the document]
mbohal: I just quickly tested it on our " private project" and copilot seems to be right.

I would suggest we remove the `user` directive on out private project and if it does work there for a few days of active development, we can remove it in this PR as well.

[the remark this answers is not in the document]
bedrich-schindler: Are you sure? `fixuid` does need user parameter as stated in its docs: https://github.com/boxboat/fixuid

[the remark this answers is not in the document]
bedrich-schindler: Consider adding comment about this.

src/docs/contribute/general-guidelines.md
bedrich-schindler: Martin agrees with me.

docker/react_ui_devcontainer/Dockerfile
bedrich-schindler: Will be fixed.

bedrich-schindler:
@claude make code review.

docker/react_ui_devcontainer/Dockerfile
bedrich-schindler: Fixed and tested.

src/docs/contribute/general-guidelines.md
bedrich-schindler: Closing this comment in favor of another comment where to discuss auto mode.

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,228 @@
...
+| Container    | Purpose                                          |
+|--------------|--------------------------------------------------|
+| `node`       | Runs Node.js commands (`npm`, `node`)            |
+| `playwright` | Runs Playwright and Lighthouse tests             |
+| `docs`       | Serves documentation via MkDocs                  |
+
+All service containers mount the workspace at `/workspace` so that file changes
+are shared.
+
+## Automatic Service Bootstrap
```
bedrich-schindler, line 168: We spoke about this mode several times. I did my best to solve it fast and easy.

I added this section together with designated scripts. I tried to craft it with Claude as much as I can to be simple, covering as most cases as I can. Hope that documentation emphasize  that this is for experienced users only and if it causes some problems, lets turn it off, restart container and continue without it.

But unless new deps are installed, it should work a pretty good and it also make sure on startup that your deps are up to date! Personally, I sometimes forget to update them and for example Playwright might result in incorrect snapshots.

setup.sh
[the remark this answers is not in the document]
bedrich-schindler: @adamkudrna, any update (again)?

src/docs/contribute/general-guidelines.md
bedrich-schindler: cc @adamkudrna @mbohal

src/docs/contribute/general-guidelines.md
bedrich-schindler: Closing this comment in favor of another comment where to discuss auto mode.

docker-compose.base.yml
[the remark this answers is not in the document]
bedrich-schindler: Closing this comment in favor of another comment where to discuss auto mode.

src/docs/contribute/general-guidelines.md
bedrich-schindler: I am not sure I understand this comment. We updated section in overview where we distinguish between those two modes. I would not repeat again things that are obvious. It dos not add any new information imho.

mbohal (approved):
I did not test it extensively,nor did I do very throughout CR. I do not have that much time now. 

However it works on my machine and I saw nothing I would not feel comfortable merging. So here goes the ✔️

setup.sh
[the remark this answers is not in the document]
adamkudrna: I no longer see this issue.

src/docs/contribute/general-guidelines.md
adamkudrna: OK.

src/docs/contribute/general-guidelines.md
```diff
@@ -2,99 +2,232 @@
...
+### Setup
 
-To log into the container, run:
+#### Automatic setup
 
-```bash
-docker compose run --rm node_shell
-```
+> You can skip this section when using [Development Containers]. They run this
+> script automatically the first time the project is opened.
```
adamkudrna, line 34: I think this paragraph belongs to the parent section, so I can skip right to "Accessing the Development Environment" instead of "Manual setup".

src/docs/contribute/general-guidelines.md
adamkudrna: Current structure of the document is much clearer, thank you 👍🏻.

docker/react_ui_devcontainer/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        sudo \
+        ca-certificates \
+        curl \
+        gnupg \
+        vim \
+        nano \
+        openssh-client \
+        git \
```
adamkudrna, line 28: I am testing a fresh install and have issues with Git. I can reach Git in the devcontainer's terminal:

```
$ git
```

But I cannot turn on Jetbrains' integration of Git in IDE Settings:

> Git is not installed
> Failed to start Git process: Couldn't run In "/root" execute `/usr/bin/git -c core.quotepath=false -c log.showSignature=false` version: 13: Permission denied (os error 13)

Image

It already pops out on devcontainer startup:

Image

When I run the same command `/usr/bin/git -c core.quotepath=false -c log.showSignature=false` in the devcontainer's terminal, Git successfully responds.

.devcontainer/devcontainer.json
adamkudrna, line 2: I found out the name is used on the Dev Containers screen in Jetbrains:

Image

Project's terminal is also nice:

Image

But when I open the project in IDE, the project name matches the root folder which is simply `workspace`:

Image

This is not very useful when navigating in recent projects, and I wonder what happens with worktrees… Is there an easy way to make Jetbrains project name inherit from this setting?

docker/react_ui_devcontainer/Dockerfile
adamkudrna, line 1: FYI, Jetbrains IDE asks for permission about five times during setup:

Image

All I need is to click "Allow".

src/docs/contribute/general-guidelines.md
adamkudrna: I quickly tested both modes by editing a component (no re-install) and it seems to work well. 👏🏻

scripts/auto-start-node.sh
```diff
@@ -0,0 +1,33 @@
...
+# File to read the hash of the package-lock.json
+LOCK_HASH_FILE="node_modules/.package-lock-hash"
+
+# Parent directory of the script
+SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
+
+# Change to the parent directory of the script
+cd "$SCRIPT_DIR/.."
+
+# Install dependencies if node_modules is missing or out of date
```
adamkudrna, line 14: 👍🏻 for the check.

adamkudrna (approved):
You have done a great job @bedrich-schindler 👏🏻.

- Docs are clear and understandable.
- I tested the default mode and auto-start mode, both OK.
- I successfully started Claude (but stopped before logging in).
- I did **not** test any customization features.
- Only issue I have observed so far is the missing Git intergration in my Jetbrains.

Can't wait to start using it!

PR #715, opened 2026-06-29
Add `Popover` z-index token and `FormLayoutCustomField` label auto-hide

adamkudrna:
## Summary

- Add `--rui-Popover__z-index` CSS custom property (default: `auto`) so consumers can control stacking context without attribute selectors or class injection; document usage in README
- Allow `visibility` through `cleanPlacementStyle` so callers can hide the Popover before its initial position is computed (flash guard); document in README
- Auto-hide labels of box fields nested in `FormLayoutCustomField` via context so label visibility is handled automatically by the layout

## Test plan

- [ ] Popover appears above sticky/fixed elements when `--rui-Popover__z-index` is set
- [ ] Popover with `visibility: hidden` in `placementStyle` is hidden before position is ready
- [ ] Labels of box fields inside `FormLayoutCustomField` are hidden automatically
- [ ] Visual tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adamkudrna:
🙈

PR #719, opened 2026-07-07
Fix Claude Atlassian plugin installation

bedrich-schindler:


mbohal:
@bedrich-schindler do we want it here? Generally we create (or should be creating) GH issues for work done on RUI.

Why do we want the Atlassian plugin here?

adamkudrna (approved):

adamkudrna:
> Why do we want the Atlassian plugin here?

@mbohal It's not a plugin, just the official plugin repository.

adamkudrna (commented):
I tried on master and **I'm able** to install/enable the Atlassian plugin without this change. I'm all for adding the plugin, but this change neither installs it, nor fixes any real issue (for me at least).

bedrich-schindler:
> @bedrich-schindler do we want it here? Generally we create (or should be creating) GH issues for work done on RUI.
> 
> Why do we want the Atlassian plugin here?

I am not sure if I understand. To automatically install plugins, we need to have this added - you were the one that reported that was not able to install Atllasian plugin in our private project, so we added it everywhere and this is reason why it is added here as well. It is same for Github one, it is one "repo". 

We need this for the same reason we have Atlassian in ours. Or not?

> I tried on master and **I'm able** to install/enable the Atlassian plugin without this change. I'm all for adding the plugin, but this change neither installs it, nor fixes any real issue (for me at least).

As I stated above, @mbohal had issues with Attlasian plugin in private repos and the fix is same for Github. If you manually approved it once or you have it enabled generally for Claude (not just for the project), you never meet this issues. But if you do fresh install with docker compose down -v and you do not have installed and enabled plugins in your Claude configuration (which I do), it should fail. I did not test it here, but it should be same case as for Attlasian which is in same repo.

@adamkudrna and @mbohal, it is up to you two to decide.

PR #726, opened 2026-08-13
Skip writing lockfile hash in `postinstall` when installed as a dependency (#725)

### LINKED ISSUE #725: `postinstall` script in published 0.64.0 package breaks `npm ci` in consumer projects
# Description

`@react-ui-org/react-ui@0.64.0` was published with a `postinstall` hook in
`package.json`:

```json
"postinstall": "sh scripts/write-lockfile-hash.sh"
```

npm runs `postinstall` not only in the react-ui repo checkout, but also every
time the package is installed as a dependency. In a consumer project, the
script fails and aborts the consumer's entire install.

# Steps to reproduce

1. In any project, add `"@react-ui-org/react-ui": "^0.64.0"` as a dependency.
2. Run `npm ci` (or `npm install`).

## Actual behavior

The install fails:

```
npm error code 2
npm error path <project>/node_modules/@react-ui-org/react-ui
npm error command failed
npm error command sh -c sh scripts/write-lockfile-hash.sh
npm error sha256sum: package-lock.json: No such file or directory
npm error scripts/write-lockfile-hash.sh: 15: cannot create node_modules/.package-lock-hash: Directory nonexistent
```

# Expected behavior

Installing the package as a dependency succeeds; the lockfile-hash mechanism
only runs inside the react-ui repo checkout.

# Root cause

`scripts/write-lockfile-hash.sh` assumes it runs in a repo checkout: it hashes
`package-lock.json` into `node_modules/.package-lock-hash` (consumed by
`scripts/docker/autoStartNode.sh` to detect stale installs). Neither
assumption holds for an installed copy of the package:

* npm never includes `package-lock.json` in a published tarball, so
  `sha256sum package-lock.json` fails.
* The dependency's own `node_modules/` directory does not exist, so the
  redirect to `node_modules/.package-lock-hash` fails.

With `set -e`, the script exits non-zero and npm treats it as a fatal install
error.

For reference, the `postinstall` in 0.63.0
(`cp -n .env.dist .env && cp -n .env.playwright.dist .env.playwright || true`)
ended with `|| true`, so it could never fail for consumers. The unguarded
script is the 0.64.0 regression.

# Suggested fix

In `scripts/write-lockfile-hash.sh`, right after the `cd "$SCRIPT_DIR/.."`
line, exit early when not in a repo checkout:

```sh
# Skip when running as an installed dependency: the published package
# contains no package-lock.json.
if [ ! -f package-lock.json ]; then
  exit 0
fi
```

The lockfile is a reliable marker of a repo checkout because npm never packs
it into a published tarball.

Removing the `postinstall` hook instead is not a good fix: the hash must be
rewritten whenever a developer runs `npm ci` manually in the devcontainer,
otherwise the stale-install detection in `scripts/docker/autoStartNode.sh`
silently breaks.

# Verification

1. In the repo, run `npm ci` and confirm `node_modules/.package-lock-hash` is
   still written with the sha256 of `package-lock.json`.
2. Simulate a consumer install: `npm pack`, then in an empty temp directory
   run `npm init -y && npm install <path-to-tarball>` and confirm the install
   succeeds with no postinstall error.

Afterwards, release the fix as 0.64.1 (patch version bump).

# Workarounds until fixed

* Pin `@react-ui-org/react-ui` to `0.63.0`, or
* `npm ci --ignore-scripts` (skips all dependencies' lifecycle scripts).

bedrich-schindler:


bedrich-schindler:
## Why this fix: alternatives considered

**1. `INIT_CWD`/path heuristic** (à la [`skip-local-postinstall`](https://github.com/tobua/skip-local-postinstall))
Compares where `npm install` was invoked with where the script runs. Works for
the common case, but it is package-manager-specific, unset when the script is
run directly, and has symlink/workspace edge cases (it would skip the hash in
a legitimate checkout inside a monorepo). The referenced package actually
solves the inverse problem (skip locally, run for consumers) and would add a
dependency.

**2. Append `|| true` to the `postinstall`** (the 0.63.0 style)
Hides the failure for consumers, but also silently swallows real errors in the
repo checkout.

**3. Exit early when `package-lock.json` is absent** ✅ (chosen)
npm unconditionally excludes the lockfile from published tarballs, so its
absence is a deterministic marker of "installed as a dependency" — not a
heuristic. It is package-manager- and invocation-independent, and its
worst-case failure mode is benign: the hash is not written and the
devcontainer just does one unnecessary reinstall.

adamkudrna (approved):

mbohal (approved):

PR #729, opened 2026-08-15
Add `remoteUser` to `.devcontainer` to fix JetBrains IDE support (#728)

### LINKED ISSUE #728: JetBrains IDE fails to start Development Containers in Clone mode
JetBrains IDE fails to start Development Containers in Clone mode with following error:

```
Setting global Git config:                                                                                                                                                                                                                                 
  user.name = Bedřich Schindler                                                                                                                                                                                                                              
  user.email = bedrich.schindler@gmail.com                                                                                                                                                                                                                   
  Process is exited with non zero code: 1.                                                                                                                                                                                                                   
  stderr: chgrp: invalid group: '1000:1000'  
```

bedrich-schindler:
Opening the project in WebStorm using the Dev Containers "Clone" mode failed during the IDE remote agent bootstrap with:

chgrp: invalid group: '1000:1000'

The `chgrp` call is issued by the JetBrains dev-container helper while it fixes the ownership of its remote agent files. With no user declared in `devcontainer.json`, WebStorm falls back to the Compose `user:` directive (`${COMPOSE_UID}:${COMPOSE_GID}`) and passes the literal `1000:1000` string to `chgrp`, which accepts only a group, never a `uid:gid` pair.

Declaring `"remoteUser": "developer"` makes WebStorm resolve the user via `/etc/passwd` inside the container instead. This is safe on all platforms: `fixuid` rewrites the `developer` entry at container start to match the uid:gid the container runs as, and `devcontainer.json` is not read outside the dev-container workflow (plain `docker compose up`, terminals, CI), so nothing else is affected.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,7 @@
     "../docker-compose.yml"
   ],
   "service": "devcontainer",
+  "remoteUser": "developer",
   "shutdownAction": "stopCompose",
   "workspaceFolder": "/workspace",
   "forwardPorts": [


.devcontainer/devcontainer.json
```diff
@@ -5,6 +5,7 @@
     "../docker-compose.yml"
   ],
   "service": "devcontainer",
+  "remoteUser": "developer",
```
adamkudrna, line 8: I tested it and it works ✅. But I'm seeing a strange thing: no `node_modules` directory in the workspace, but running `npm start` and serving the docs after `npm ci` works. Does it have anything to do with the "Shallow repository" IDE notification?

Image

adamkudrna (approved):

.devcontainer/devcontainer.json
bedrich-schindler: According to JetBrains documentation:

> If you want to perform [shallow clone](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--depthcodeemltdepthgtem) with a limited history, select the Shallow clone with a history truncated to checkbox and specify the number of commits that you want to clone.
You can fetch the rest of the history later by selecting Git | Unshallow Repository in the main menu.

---

Regarding `node_modules`, I need to verify this.

bedrich-schindler: > Regarding node_modules, I need to verify this.

This is happening despite of change in this branch. I can confirm that when connected to `devcontainer`, `node_modules` are not present in this container, but they are present in `node` container. As `node` container serves the client, it works. So it is an issue in Devcontainers, will to dig into it separately.

PR #730, opened 2026-08-25
Bump `docoff` version to v0.7.2

mbohal:


mkdocs.yml
```diff
@@ -69,7 +69,7 @@ extra_javascript:
   - docs/_assets/js/setupDocoff.js
 
   # Then load and init Docoff
-  - https://unpkg.com/@react-ui-org/docoff@0.5.4/public/generated/bundle.js
+  - https://unpkg.com/@react-ui-org/docoff@0.7.1/public/generated/bundle.js
```
bedrich-schindler: When you type into the code preview, it flickers with each key stroke. 

https://github.com/user-attachments/assets/70b7102a-8070-4036-9d70-00ba3057c339

bedrich-schindler (changes requested):

mkdocs.yml
mbohal: fixed in `v0.7.2`

mkdocs.yml
mbohal, line 1: Update `docoff` version in commit message before merge

mkdocs.yml
```diff
@@ -69,7 +69,7 @@ extra_javascript:
   - docs/_assets/js/setupDocoff.js
 
   # Then load and init Docoff
-  - https://unpkg.com/@react-ui-org/docoff@0.5.4/public/generated/bundle.js
+  - https://unpkg.com/@react-ui-org/docoff@0.7.2/public/generated/bundle.js
```
bedrich-schindler, line 72: When preview is empty, cursor is overflowing:

Image

bedrich-schindler, line 72: I am just curious. When invalid output is entered, it sometimes show error, sometimes is does not, sometimes it shows old preview.

Image

Image

Image

From my point of view, it seems unpredictable.

bedrich-schindler (changes requested):

mkdocs.yml
mbohal: When there is an unknown tag it shows nothing. This has nothing to do with docoff.

image

produces:

image

Which browser handles normally, that is ignores it.

Unclosed tag on the other hand is malformed HTML and it raises an error.

Previous version behaved the same way.

mbohal: That has nothing to do with this PR, it is a bug already present in the previous version.

I created https://github.com/react-ui-org/docoff/issues/52

mbohal: I think it is behaving correctly. 

If we would want to change this, I'm not really sure how to detect unknown tag.

There is no error. The code produces HTML and browser renders it. The output is empty, but that is a decision made by the browser, not docoff.

bedrich-schindler: I was not aware of that. Thanks for info.

bedrich-schindler (approved):

bedrich-schindler (approved):

mkdocs.yml
mbohal: fixed

PR #734, opened 2026-08-28
Migrate the library to TypeScript (#394)

### LINKED ISSUE #394: Transform codebase to Typescript

bedrich-schindler:
Convert all components, providers, helpers and the library entry from JSX with `prop-types` to TypeScript (`strict: true`) while keeping the runtime behaviour and the public exports unchanged.

* Every component gets a co-located `<Component>.types.ts` with its `Props` type (extending the matching `*HTMLAttributes`) and the JSDoc descriptions copied from `propTypes`; context values, provider props and helper types live in `.types.ts` files too. Shared unions (`Color`, `Priority`, `Size`, `Layout`, `ValidationState`, `Breakpoint`, `LabelPosition`) live in `src/types`.
* `defaultProps` are replaced by default values in the parameter destructuring (react-docgen reads them for the documentation); `propTypes` are kept for runtime validation. Plain function components are annotated as `React.FunctionComponent` so `propTypes` do not leak into the declarations, and `withGlobalProps` is generic in props and ref element (`never` for components that do not forward a ref).
* Toolchain: TypeScript 7 (`@typescript/native`) type-checks and emits declarations, TypeScript 6 stays installed as `typescript` for tools that need the compiler API (typescript-eslint); Babel (`@babel/preset-typescript`) transpiles for webpack and Jest, which replaces `ts-jest` and the second Jest config. `eslint-plugin-deprecation` and `eslint-plugin-typescript-sort-keys` are replaced by `@typescript-eslint/no-deprecated` and `eslint-plugin-perfectionist` because they do not support TypeScript 6.
* Build emits `dist/types/**` and bundles them with `dts-bundle-generator` into the published `dist/react-ui.d.ts` referenced by `types`; package contents are defined by a `files` allowlist instead of `.npmignore`.
* Stories use the real component props (`StoryProps` helper for props the story defaults), `npm run typecheck` is part of `npm run lint`, and the contributing rules and docs describe the TypeScript setup.

Closes #394


bedrich-schindler:
@mbohal @adamkudrna, Claude was used to migrate whole RUI from JS to TS, but I check every single file before creating pull request (yes, it took a while).

I would like to point out, that with @mbohal we decided to do necessary steps to convert library to TS. Things like removing prop types is part of React update and can be done after updating docoff.

Unfortunately, it was not possible to split it into multiple files as it would be more messy that doing it in one shot.

bedrich-schindler:
One more thing. I did minimum changes to source code. So instead of fixing code, I simply added lint ignore so it is not huge mess.

.claude/rules/frontend.md
```diff
@@ -1,15 +1,16 @@
...
 ---
 
 # Frontend (React component library)
 
 ## Commands
 
 * If dependencies change, run `npm ci`.
-* Run `npm run eslint` after changes (add `-- --fix` to autofix).
+* Run `npm run eslint` after changes (add `-- --fix` to autofix) and
+  `npm run typecheck` (TypeScript 7 `tsc --noEmit`).
```
mbohal: I don't think we should specify the package versions here (TS here and more bellow).

For one we use TS 7 and also it will be hard to keep in sync here. I think we should omit it here. Possible we could add note to look in `package.json`

src/components/_helpers/getRootColorClassName.ts
```diff
@@ -1,4 +1,9 @@
-export const getRootColorClassName = (variant, styles) => {
+import type {
+  Color,
+  CssModuleClasses,
+} from '../../types';
+
+export const getRootColorClassName = (variant: Color | undefined, styles: CssModuleClasses) => {
```
mbohal: This API where optional parameter is first is strange. I would suggest changing the signature to `getRootColorClassName(styles, color)`. Ideally in separate commit/issue.

It likely has no relation to this PR, but it became more obvious.

Same goes for:
* `src/components/_helpers/getRootPriorityClassName.ts`
* `src/components/_helpers/getRootSizeClassName.ts`
*  `src/components/_helpers/getRootValidationStateClassName.ts`

.eslintrc
```diff
@@ -17,7 +17,20 @@
   ],
   "parser": "@babel/eslint-parser",
   "rules": {
+    "import/extensions": ["error", "ignorePackages", {
+      "js": "never",
```
mbohal: I assume we need to keep these here for the untransformed files (e.g. `src/components/Grid/_helpers/__tests__/generateResponsiveCustomProperties.test.js`).

But why do we need to keep the `*.js` files around? Can we not just use TS everywhere?

package.json
```diff
@@ -82,6 +97,7 @@
     "@babel/eslint-parser": "^7.29.7",
     "@babel/preset-env": "^7.29.7",
     "@babel/preset-react": "^7.29.7",
+    "@babel/preset-typescript": "^7.29.7",
```
mbohal, line 97: I wonder if we need to keep babel around.

The transpiled code is used only for:
* documentation website
* prebuild and distributed `react-ui[.development].js` for "easy" use

I think for both uses we could maybe drop babel/browserlist and just use ES2020. It would simplify the deps/stack.

Would @adamkudrna agree?

src/providers/translations/TranslationsProvider.tsx
```diff
@@ -4,27 +4,28 @@ import React, {
...
-  translations,
-}) => {
+  translations = {},
+}: TranslationsProviderProps) => {
   const contextTranslations = useContext(TranslationsContext);
 
   return (
     <TranslationsContext.Provider
-      value={mergeDeep(contextTranslations, translations)}
+      value={mergeDeep(contextTranslations, translations) as Translations}
```
mbohal: Could we maybe replace `mergeDeep` with `mergeDeep<returnType>`?

I think it would be nicer to use as the the `as`  annotations always seem bolted on.

mbohal (changes requested):

package.json
bedrich-schindler: Your proposal is a big change, it is more that about deliverable output than about TypeScript migration. Can you write proposal and raise it on our Monday meeting?

bedrich-schindler:
@mbohal, ready for round 2.

bedrich-schindler:
Update commit message before merge.

package.json
mbohal: I added to our Monday meeting program.

However I do not think we need to wait for the decision here. I'm fine with merging this PR and getting back to babel later on.

jest.config.js
```diff
@@ -1,23 +1,22 @@
 module.exports = {
   moduleFileExtensions: [
     'js',
```
mbohal, line 4: I think we should add a comment that `js` can not be removed as jest wont start without it.

mbohal (approved):

Issue #723, opened 2026-08-03
`changed` state for input fields


mbohal:
Presently input fields support optional prop `validationState` that takes values  `valid`, `invalid`, `warning` that correspond to colors green, red and yellow respectively.

Now we nee a new state `changed` that will be indicated by yet another color.

mbohal:
I think there are several approaches how to handle this.

1. Adding a new `validationState` value
2. Reusing the `valid` state to indicate `changed`.
3. Adding a new prop `changed`
4. Renaming `validationState` to something more generic (e.g. `state`)  and adding the new option `changed`
5. Just use `color` prop with the palette that is used everywhere

My take on these:

1. Is strange as `changed` has nothing to do with validation.
2. is wrong as being changed has nothing to do with being valid. Also the `valid` state could be useful although we do not use it presently.
3. would require us to handle situation where both `changed` and `invalid` / `warning` / `valid`  are used. I think we could assume that `validationState` beats `changed` in priority. However adding another prop makes the API more complex especially with the "hidden" rule of `validationState` taking priority over `changed`.
4.  Is a breaking change and is less flexible then 3).
5. It allows using needlessly  rich color palette that could lead to confusion if not handled properly.

I think we should decide between 3) and 4) and 5). I do not have strong feelings for either one of these approaches.

adamkudrna:
An ADR for this is now open in #735 — it lands on option 4, generalised into a `State` collection (`valid` / `invalid` / `warning` / `changed`), with four open questions there that need deciding before implementation.

Issue #712, opened 2026-05-29
`helpText` color


mbohal:
The `helpText` of input fields should not change color depending on the `validationState` prop.

When the `helpText` is the same color as the `validationText` it makes it hard to distinguish these two.

Relates: #711

Issue #711, opened 2026-05-29
Switch `helpText` and `validationText` order


mbohal:
For our design system based on RUI, we would need to show the `validationText` before the `helpText` when some input element is in invalid state.

The logic in support of this change is:

1. if I see the error the `helpText` was likely not helpful enough and the `validationText` is more useful to the user.
2. In cases where the `helpText` is long it would put the `validationText` too far from the input element.

For our use case the change of rendering order can be either hardcoded or it can be somehow configurable.

PR #735, opened 2026-08-31
Add ADR for the `State` collection (#723)

adamkudrna:
Records the decision to introduce a `State` collection (`valid`, `invalid`, `warning`, `changed`) and to rename the form-field prop `validationState` to `state`, as the first architecture decision record in this project.

This is the record of the decision only — no implementation. Relates to #723, but deliberately does **not** close it: the `changed` state itself is still to be built.

## What's here

- `src/docs/contribute/decisions/0001-state-collection.md` — the ADR, status _Proposed_.
- `mkdocs.yml` — a new _Decisions_ group under _Guides › Contribute_, so the page isn't an orphan in the nav.

## Open questions in the ADR

The document ends with four questions that need a decision before implementation starts:

1. Do #711 and #712 ride along in the same breaking-change window?
2. Is `changed` the right word (vs. `modified`, `unsaved`, `dirty`)?
3. Which hue for `--rui-color-state-changed`? Every blue-ish slot is taken.
4. Does `Card` land in the same change or a follow-up?

## Notes

- No ADR structure existed before, so this PR establishes the location and numbering. No index page yet — worth adding once there's a second ADR.
- Verified: `npm run lint`, `mkdocs build`, `npm run build` and `npm run test:jest` all pass. `npm run test:playwright-ct:all` could not run — the `playwright` container aborts with `EBADDEVENGINES` (Node v24.17.0 vs. the `^22.23.0 || ^24.19.0` pin from dd332ed). This is pre-existing on `master` and unrelated to this change; the container image likely needs rebuilding.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -153,3 +153,5 @@ nav:
           - Composition: 'docs/contribute/composition.md'
           - CSS Guidelines: 'docs/contribute/css.md'
           - Releasing: 'docs/contribute/releasing.md'
+          - Decisions:
+              - '1. State Collection': 'docs/contribute/decisions/0001-state-collection.md'
--- a/src/docs/contribute/decisions/0001-state-collection.md
+++ b/src/docs/contribute/decisions/0001-state-collection.md
@@ -1,0 +1,176 @@
+# 1. A State collection for data-bearing components
+
+**Status:** Proposed — resolves #723, supersedes the approach discussed in #721.
+
+## Context
+
+Form fields support `validationState` with the values `valid`, `invalid` and
+`warning`, rendered in green, red and amber. #723 asks for a new `changed`
+state — "this field's value differs from the saved one" — in a colour of its own.
+
+Feedback from projects building on React UI shows the current API doesn't cover
+what they need, and the workarounds they've adopted are informative:
+
+- **`valid` is not used.** Across the products we have visibility into, no form
+  field ever sets `valid`. The state exists but earns nothing.
+- **`warning` is being abused to mean "changed".** Because there's no state for
+  "this value has been edited but not saved", projects set `warning` and then
+  override the `--rui-FormField--warning--*` tokens to repaint it. One project
+  has an entire app-level theme file that exists for no other purpose, and has to
+  place it outside any cascade layer to win against the library's `theme` layer.
+  The override is global, so it also recolours genuine `warning` uses elsewhere
+  in the same app — the state becomes unusable for its actual meaning.
+- **Projects invent the vocabulary we didn't give them.** More than one has
+  defined its own state type (`invalid | modified | null`) plus adapter functions
+  whose only job is to downgrade that vocabulary at the React UI boundary. One
+  carries *two* such adapters for a single concept: one mapping it to a field's
+  `validationState`, another mapping it to a container's `color`, because nothing
+  relates the two.
+- **Precedence is already settled downstream.** Every implementation resolves its
+  own precedence — invalid beats changed — before reaching us.
+
+That last point matters for the API shape: a separate `changed` prop would make
+the library re-implement a policy consumers already own, and introduce a hidden
+interaction between two props.
+
+The deeper problem is that `validationState` is an outlier in our own design
+system. `foundation/collections.md` defines Collections with the rule *"if an
+option from a collection is used in a component, all other options from the same
+collection must be available for use in that component too."* Card, Alert, Badge,
+Button and Modal implement the Feedback **colour** collection through `color`,
+looping it in their settings. `validationState` is an ad-hoc, undocumented
+three-value list that only *coincidentally* maps to three of the six feedback
+colours, hand-written as three SCSS blocks in each of nine components.
+
+## Decision
+
+Introduce a **State collection** — `valid`, `invalid`, `warning`, `changed` — as
+a first-class collection alongside the Colour collections, and rename the
+form-field prop `validationState` → `state`.
+
+| | Answers the question | Values | Taken by |
+|---|---|---|---|
+| **Feedback colours** | what should this look like? | `success` `warning` `danger` `help` `info` `note` | Alert, Badge, Button, Modal — messaging and decoration |
+| **State** | what condition is this data in? | `valid` `invalid` `warning` `changed` | form fields, and later Card and other components that hold or contain data |
+
+`changed` receives a colour of its own but **does not** join the Feedback palette.
+The state → colour mapping becomes one small, themeable surface:
+
+```scss
+--rui-color-state-valid: var(--rui-color-feedback-success);
+--rui-color-state-invalid: var(--rui-color-feedback-danger);
+--rui-color-state-warning: var(--rui-color-feedback-warning);
+--rui-color-state-changed: /* one new hue */;
+```
+
+Documented precedence, for components that derive a state from their children
+(such as `InputGroup`): `invalid` > `warning` > `changed` > `valid`.
+
+## Rationale
+
+### Aren't `valid`/`success` and `invalid`/`danger` duplicates?
+
+No — they're different kinds of list. **Feedback is a palette; State is a
+vocabulary.** `success` is a colour name meaning "whatever you want to say in
+green". `valid` is a fact about data. `valid`/`success` isn't a duplicated pair,
+it's a *mapping* between the two.
+
+The mismatch at both ends is what proves they're distinct collections: `changed`
+has no feedback colour, and `help`, `info` and `note` have no state. If they were
+one list, that couldn't happen.
+
+The test that separates them is what question the prop answers. `color` answers
+"what should this look like" — a free authoring choice. `state` answers "what
+condition is this data in" — not a choice, a consequence. You'd never write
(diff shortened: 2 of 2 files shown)


src/docs/contribute/decisions/0001-state-collection.md
```diff
@@ -0,0 +1,176 @@
...
+|---|---|
+| 1. New `validationState` value | The name would then lie: `changed` is not a validation result. Superseded by renaming the prop. |
+| 2. Reuse `valid` for `changed` | Rejected in the thread; and `valid` is unused in practice, so this trades one unused state for a misleading one. |
+| 3. Separate `changed` prop | Makes the library own a precedence rule consumers already implement themselves, and adds a hidden interaction between two props. |
+| 4. Rename + new value | **This proposal,** generalised beyond form fields into a documented collection. |
+| 5. Generic `color` prop (#721) | Gives up the semantics entirely; consumers keep hand-mapping meaning → colour, so the duplicate adapters survive. It also makes a field's colour a free authoring choice, which it shouldn't be. |
+
+## Open questions
+
+1. **Do #711 and #712 ride along?** All three rework the same rendering block,
```
mbohal, line 153: I think they should be a separate commit in the same release so consumers only deal with BC change once.

src/docs/contribute/decisions/0001-state-collection.md
```diff
@@ -0,0 +1,176 @@
...
+   the same `surrounding-text-color` token and the same documentation sections.
+   One breaking-change window means dependent projects migrate once instead of
+   twice; separate PRs are cleaner to review. Leaning towards bundling.
+2. **Is `changed` the right word?** Alternatives: `modified`, `unsaved` (names
+   the workflow rather than the element), `dirty` (React jargon, reads poorly in
+   end-user documentation). Hard to change later.
+3. **Which hue?** Every blue-ish slot is taken (`note` and `selected` `#007bff`,
```
mbohal, line 160: The consumer apps generally speaking redefine these, so I would let @adamkudrna pick whatever he likes.

src/docs/contribute/decisions/0001-state-collection.md
```diff
@@ -0,0 +1,176 @@
...
+*state* `warning` ("validation raised a non-blocking concern") and the *colour*
+`warning` (amber) are different things sharing a word. This needs an explicit
+note in the documentation; renaming either is out of scope here.
+
+**`validationText` is untouched.** It and `helpText` are already independent
+props, both rendered, both coloured from the same token — which is precisely
+what #712 is about, and their order is what #711 is about. Folding them together
+would destroy the distinction those issues exist to sharpen.
+
+## Alternatives considered
```
mbohal, line 141: I'm unsure about the relationship of the members of the `state` dictionary. Should they be mutually exclusive?

If we go with the proposal, then the usage would be:

* `invalid`: used for modified and unmodified data that fail validation
* `warning`: used for modified and unmodified data that fail validation soflty
* `valid`: unused
* `changed`: used for changed and valid data (as invalid data use the `invalid` state)

This seems somehow asymmetrical to me. 

@adamkudrna what arguments do we have against using `valid` for modified and valid?

src/docs/contribute/decisions/0001-state-collection.md
```diff
@@ -0,0 +1,176 @@
...
+| 4. Rename + new value | **This proposal,** generalised beyond form fields into a documented collection. |
+| 5. Generic `color` prop (#721) | Gives up the semantics entirely; consumers keep hand-mapping meaning → colour, so the duplicate adapters survive. It also makes a field's colour a free authoring choice, which it shouldn't be. |
+
+## Open questions
+
+1. **Do #711 and #712 ride along?** All three rework the same rendering block,
+   the same `surrounding-text-color` token and the same documentation sections.
+   One breaking-change window means dependent projects migrate once instead of
+   twice; separate PRs are cleaner to review. Leaning towards bundling.
+2. **Is `changed` the right word?** Alternatives: `modified`, `unsaved` (names
```
mbohal, line 157: I would  go with either `changed` or `modified`. I have no clear preference between these two.

mbohal (changes requested):

mkdocs.yml
```diff
@@ -153,3 +153,5 @@ nav:
           - Composition: 'docs/contribute/composition.md'
           - CSS Guidelines: 'docs/contribute/css.md'
           - Releasing: 'docs/contribute/releasing.md'
+          - Decisions:
+              - '1. State Collection': 'docs/contribute/decisions/0001-state-collection.md'
```
mbohal, line 157: I like having the ADRs public

src/docs/contribute/decisions/0001-state-collection.md
mbohal, line 153: Unfortunately the hash syntax does not work in `*.md` files:

Image

We need to use regular link I guess.

mkdocs.yml
bedrich-schindler: Me not. I agree there is added value for having ADR, but to be honest. We have made a lot of decisions in the past from which this is no more important than rest we made. What more, rest of the documentation is quite curated and compact in comparison how this ADR looks like. Nobody will read such a long text, nobody - I just do not want to read obviously generated text by AI. It is longer than most of the documentation markdowns we have.

I see those options:

1. Have it only in file structure, not in the public docs 
2. In public docs only say that if you want to check decisions we made under the hood, check this folder
3. Have it as Adam made it


To be honest, 3 is almost no go for me at the monent. If you want to have it, I think 2 is the way

if adr is humanized and shortened without AI noise, we can discuss it again.

bedrich-schindler:
Technically, suggested option with new pallete is the way to go.

mkdocs.yml
mbohal: > adr is humanized and shortened without AI noise

This should be always the case regardless of what we choose here.

Lets (especially in time of AI) not forget that:

> Code/text is not a feature, it is a liability.


As far as the ADR publication goes, I think having them public might be beneficial, but I have no strong feelings. Option `2)` is OK with me.

bedrich-schindler: My point of view. I check libraries like Material ui, Bootstrap and Adobe's React Spectrum. I have not find any dedicated ADRs/decisions. As an end-user, I do not care about the decisions the team has made, their library, their choices. So if we need ADRs, I would keep it "private" (not part of public docs). 

To be honest, I find having ADRs at this point unnecessary. We made so much decisions on calls or in issues and none of them is documented. Having one ADR at this point is waste of time. I personally do not want to have it in git at all. Period. If you two decide other way, do it, but I am not supporting this decision ... or not unless we are changing something after 1.0 is released, then we can communicated architectonical changes if we change them.

mbohal: On phone call agreed we will leave it in git, but we will not include it in docs.

We will ADR `001` that describes why we want to start using ADR so late in the project.

adamkudrna: I was thinking about it and I agree that starting with a prop ADR is weird. I created an umbrella issue to agree on the most important decisions we made in the project's history which we want to record: #754. This can PR hang here until the ADR list is settled so it gets the right number.

src/docs/contribute/decisions/0001-state-collection.md
adamkudrna: Technically, the states are not mutually exclusive: a field _can_ be e.g. changed and invalid at the same time. But we don't want users to distinguish that, at least the consumer app designs don't mention it and put the invalid state as the one more important.

---

I can imagine displaying both kinds of information at the same time. For example, changed fields could alter not only the input color, but also the label color. Invalid state would then just override the input color:

- changed, valid: blue label, blue input
- changed invalid: blue label, red input

But as I said, nobody asked for that just yet and designs don't show it either.

adamkudrna: And to your arguments question. I understand it as "what's important for the user":

1. `invalid` – needs me to take action to be able to continue, top priority
2. `warning` – I should know it may break something
3. `changed` – does not need action, but I want to see what I changed before saving
4. `valid` – probably hardly ever combined with the `changed` use case on the same page (or in the same form)

Take it as an example, consumer apps are free to decide. And the one we are making it for just does not want to show more than one state at a time.

Maybe we could record this in the ADR or somewhere?

src/docs/contribute/decisions/0001-state-collection.md
adamkudrna: My preference is slightly leaning towards `changed` because its connection to native HTML and its `onchange` attribute. Also, it's easier to understand.

decisions/0010-state-collection.md
```diff
@@ -0,0 +1,88 @@
...
+## Context
+
+Form fields take `validationState` with the values `valid`, `invalid` and
+`warning`. [#723][723] asks for a `changed` state: the value differs from the
+saved one.
+
+Projects using React UI already work around the missing state:
+
+- `valid` is never used.
+- `warning` is repainted to mean "changed", which makes it unusable for real
```
mbohal: `repainted` -> `repurposed`

mbohal (approved):

PR #736, opened 2026-09-07
Keep theming out of component README prose (#733)

### LINKED ISSUE #733: Remove all mentions of theming from README prose to make downstream sync easier
We reuse the component READMEs in our downstream design system. Theming is considered an implementation details in that design system, so keeping the Theming tables makes no sense there.

We want to sync component READMEs automatically, and to be able to do that, we would appreciate if we could have a simple rule like "never include the Theming section" rather then updating the docs prose whenever it mentions a theming feature.

See RADES-139.

adamkudrna:
Component READMEs are reused in downstream design systems where theming is an implementation detail, so the whole `## Theming` section is dropped when syncing. Today that cut is not safe: theming leaks into prose elsewhere in the file, which turns every upstream edit into a manual sync conflict.

This moves every theming reference into the `## Theming` section, so removing that section leaves a coherent document.

## Changes

* Dropped theming links and the clauses that only describe theming from `Card`, `Paper`, `FormLayout`, `TextField`, `TextArea`, and `SelectField`.
* Split `Styling the Required State` in `CheckboxField`, `Radio`, `SelectField`, and `Toggle` into two parts:
    * `Rendering as Required` under _States_ keeps the `renderAsRequired` prop story, with a plain example.
    * `Theming the Required State` under _Theming_ carries the Forms Theming link, the label color edge case, and the example that overrides `--rui-*` custom properties.
* Moved the `Popover` `z-index` section under _Theming_ — it documented nothing but the `--rui-Popover__z-index` override, which is already listed in the theming table.
* Inlined the `[theming-forms]` link in `TextField` so its definition doesn't outlive the section that uses it.
* Repointed the four component links in _Forms Theming_ to the new anchors.
* Recorded the rule in the [contributing guide](src/docs/contribute/general-guidelines.md#documenting) and in [Claude rules](.claude/rules/docs.md).

## Verification

* `npm run lint` and `mkdocs build` pass.
* No `theming` mention or `--rui-*` custom property is left above a `## Theming` heading, apart from `ScrollView`, whose `--rui-local-*` properties are arguments for the `startShadowBackground` and `endShadowBackground` options, not theming options.
* Simulated the downstream cut on all 20 READMEs with a `## Theming` section: no leftovers and no orphaned link reference definitions.

Closes #733

### DIFF
--- a/.claude/rules/docs.md
+++ b/.claude/rules/docs.md
@@ -26,6 +26,11 @@ See [Commands](../../CLAUDE.md#commands).
   each component's `README.md`; guides and foundations live under `src/docs/`.
 * New pages must be wired into the navigation in [mkdocs.yml](../../mkdocs.yml).
 * Use relative links between docs.
+* Keep theming out of README prose: document it only in the `## Theming` section
+  at the end of a component's `README.md`, including examples that override
+  `--rui-*` custom properties. Other sections must not link to
+  `/docs/customize/theming/*` or to `#theming` — downstream design systems drop
+  the whole section when they sync our docs.
 
 ## Reference
 
--- a/src/components/Card/README.md
+++ b/src/components/Card/README.md
@@ -25,10 +25,9 @@ See [API](#api) for all available options.
 
 ## General Guidelines
 
-- Default card is **designed for non-white background.** You may want to either
-  use the [raised variant](#raised-card) or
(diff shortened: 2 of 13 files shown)


bedrich-schindler:
Not sure whether I like this change or not, but approving anyway.

bedrich-schindler (approved):

Issue #738, opened 2026-09-13
Playwright container fails devEngines check, so component tests cannot run locally


bedrich-schindler:
## Problem

Running component tests locally in the `devcontainer` fails before any test starts:

```
$ npm run test:playwright-ct:all
npm error code EBADDEVENGINES
npm error EBADDEVENGINES Invalid devEngines.runtime
npm error EBADDEVENGINES Invalid semver version "^22.23.0 || ^24.19.0" does not match "v24.18.1" for "runtime"
```

## Cause

The `npm` wrapper in the `devcontainer` runs `test:playwright*` scripts in the
`playwright` service and all other scripts in the `node` service. The two
services use different Node.js and npm versions:

| Service      | Image                                  | Node.js  | npm     |
|--------------|----------------------------------------|----------|---------|
| `node`       | `node:24.19`                           | v24.19.0 | 11.17.0 |
| `playwright` | `mcr.microsoft.com/playwright:v1.61.1` | v24.17.0 | 11.13.0 |
| `playwright` | `mcr.microsoft.com/playwright:v1.62.1` | v24.18.1 | 11.16.0 |
| required by `devEngines` | —                          | `^24.19.0` | `^11.17.0` |

Commit dd332ed0 (_Update and pin `node` and `npm` versions_) tightened
`devEngines` and pinned `docker/node/Dockerfile` and the workflows to Node.js
24.19. `docker/playwright/Dockerfile` was not updated.

The Playwright images do not pin Node.js. They install the latest Node.js 24.x
from NodeSource when the image is built (`ARG NODE_VERSION=24`), so the Node.js
version depends on the image build date. Both v1.61.1 (built 2026-06-23) and
v1.62.1 (built 2026-07-30) predate Node.js 24.19.0 (released 2026-08-03). As a
result, the local `playwright` service has not met `devEngines` since dd332ed0,
whichever Playwright version is used.

## Why CI passes

`.github/workflows/playwright.yml` runs `actions/setup-node` with
`node-version: 24.19` inside the Playwright container. That puts Node.js
24.19 (with npm 11.17.0) ahead of the image's Node.js, so the `devEngines`
check passes. The local `playwright` service has no such step, so the problem
shows up only locally.

## Steps to reproduce

1. `docker compose up -d`
2. `docker compose exec -T devcontainer npm run test:playwright-ct:all`

## Expected behavior

Component tests run locally in the `playwright` service, using the same Node.js
and npm versions as the `node` service and CI.

[...]

PR #740, opened 2026-09-14
Downgrade `node` to v24.17 and `npm` to v11.13 (#738)

bedrich-schindler:


mbohal:
@bedrich-schindler  The commit message should mention why are we doing this.

mbohal (approved):

Issue #739, opened 2026-09-14
Config npm to use min-release-age


bedrich-schindler:
Set min-release-age to 7

PR #741, opened 2026-09-14
Set `min-release-age` to 7 days (#739)

bedrich-schindler:


.npmrc
```diff
@@ -0,0 +1 @@
+min-release-age=7
```
mbohal: I thought we said we wanted se it to 8 to be one day behind recommended and thus likely widely used value.

But I have no strong feeling about it - wither way is OK.

mbohal (approved):

.npmrc
bedrich-schindler: Amended with reason. My mistake.

